PR #55, opened 2020-04-01
Feat/msdk 76 update currentidentity

shuffledex:


src/utils/__tests__/index.ts
```diff
@@ -219,20 +221,34 @@ describe('stringToAccountKey and accountKeyToString', () => {
...
     const result = stringToAccountKey(value, context);
 
     expect(result).toBe(fakeResult);
   });
 
   test('accountKeyToString should convert a polkadot AccountKey object to a string', () => {
-    const fakeResult = 'someAccountKey';
+    const fakeResult = '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty';
     const accountKey = polkadotMockUtils.createMockAccountKey(fakeResult);
+    const encodeAddressResult = utilsCrypto.encodeAddress(u8aToString(accountKey));
```
monitz87, line 246:
```suggestion
    const fakeResult = 'someAccountId';
    const accountKey = polkadotMockUtils.createMockAccountKey(fakeResult);
    const encodedValue = ('encodedAccountId' as unknown) as Uint8Array;
```

src/utils/index.ts
```diff
@@ -3,7 +3,7 @@ import { createType } from '@polkadot/types/create/createType';
 import { Balance, EventRecord, Moment } from '@polkadot/types/interfaces';
 import { ISubmittableResult } from '@polkadot/types/types';
 import { stringToU8a, u8aConcat, u8aFixLength, u8aToString } from '@polkadot/util';
-import { blake2AsHex } from '@polkadot/util-crypto';
+import * as utilsCrypto from '@polkadot/util-crypto';
```
monitz87, line 6: You should import each function individually instead of doing this

src/utils/__tests__/index.ts
```diff
@@ -219,20 +221,34 @@ describe('stringToAccountKey and accountKeyToString', () => {
   });
 
   test('stringToAccountKey should convert a string to a polkadot AccountKey object', () => {
-    const value = 'someAccountKey';
+    const value = '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty';
     const fakeResult = ('convertedAccountKey' as unknown) as AccountKey;
     const context = polkadotMockUtils.getContextInstance();
+    const decodeAddressResult = utilsCrypto.decodeAddress(value);
```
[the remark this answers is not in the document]
shuffledex: The cast doesn't work as expected:  `Non-base58 character`

src/utils/__tests__/index.ts
shuffledex: The cast doesn't work as expected:  `Non-base58 character`

src/utils/__tests__/index.ts
```diff
@@ -83,6 +83,8 @@ jest.mock(
 );
 jest.mock('~/context', require('~/testUtils/mocks/polkadot').mockContextModule('~/context'));
 
+const encodedAddressStub = sinon.stub(encodeAddressModule, 'default');
```
monitz87, line 86: Better go back to stubbing in each of the tests. To bypass the sinon error, in both test suites (inside the describe but outside the tests), add an afterAll that calls sinon.restore()

src/utils/__tests__/index.ts
```diff
@@ -1113,14 +1112,17 @@ describe('signerToSignatory and signatoryToSigner', () => {
...
     fakeResult = {
       type: SignerType.AccountKey,
-      value: '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty',
+      value: someAccountKey,
     };
     signatory = polkadotMockUtils.createMockSignatory({
       AccountKey: polkadotMockUtils.createMockAccountKey(fakeResult.value),
     });
 
+    encodedAddressStub.withArgs(someAccountKey).returns(someAccountKey);
```
monitz87, line 1124: This should definitely not return the same value that is passed to it

src/utils/index.ts
```diff
@@ -203,6 +203,7 @@ export function stringToAccountKey(accountKey: string, context: Context): Accoun
  * @hidden
  */
 export function accountKeyToString(accountKey: AccountKey): string {
+  console.log(accountKey, u8aToString(accountKey));
```
monitz87, line 206:
```suggestion
```

src/utils/__tests__/index.ts
```diff
@@ -1121,7 +1124,10 @@ describe('signerToSignatory and signatoryToSigner', () => {
       AccountKey: polkadotMockUtils.createMockAccountKey(fakeResult.value),
     });
 
-    encodedAddressStub.withArgs(someAccountKey).returns(someAccountKey);
+    sinon
```
monitz87, line 1127: Again, this shouldn't return the same value that is passed to it

shuffledex: I fixed it. weird

PR #60, opened 2020-04-06
Accept, reject and cancel authorizations (MSDK-59, MSDK-63)

monitz87:


monitz87:
Both `isAuthorized` tests handle the case where the authorizations have nothing to do with the caller as well as the ones where they have proper permissions

PR #67, opened 2020-04-14
feat: create types for claims

monitz87:


src/types/index.ts
```diff
@@ -149,6 +151,50 @@ export type Authorization =
...
+export type Condition = { target: ConditionTarget; trustedIssuers: string[] } & (
+  | {
+      type: ConditionType.IsPresent | ConditionType.IsAbsent;
+      claim: Claim;
+    }
+  | { type: ConditionType.IsAnyOf | ConditionType.IsNoneOf; claims: Claim[] }
+);
+
+export interface Rule {
+  id: BigNumber;
```
VictorVicente, line 194: Do we need a BigNumber to represent the id?

PR #79, opened 2020-04-27
feat: delete all rules of a token

shuffledex:


src/api/entities/SecurityToken/Compliance/Rules.ts
```diff
@@ -25,4 +25,15 @@ export class Rules extends Namespace<SecurityToken> {
     } = this;
     return setTokenRules.prepare({ ticker, ...args }, context);
   }
+
+  /**
+   * Detele all the actual rules for the Security Token.
```
monitz87, line 30: current instead of actual

PR #81, opened 2020-04-28
feat: pause and unpause rules

shuffledex:


src/api/procedures/__tests__/togglePauseRules.ts
```diff
@@ -43,6 +46,11 @@ describe('togglePauseRules procedure', () => {
     assetRulesMapStub = polkadotMockUtils.createQueryStub('generalTm', 'assetRulesMap', {
       returnValue: [],
     });
+    assetRulesMapStub.withArgs(rawTicker).returns({
```
monitz87, line 49: resolves instead of returns

PR #84, opened 2020-05-07
fix: comment getIdentityBalance method for v1

shuffledex:


src/Polymesh.ts
```diff
@@ -84,6 +84,7 @@ export class Polymesh {
     }
   }
 
+  // TODO: uncomment the method after v1
```
monitz87, line 87: This comment is misplaced

PR #86, opened 2020-05-08
feat: implement canTransfer on issueTokens method

shuffledex:


monitz87:
I just found out that `canTransfer` cannot be used to check minting, so this task has to be put on hold until we find an alternative solution

src/Polymesh.ts
```diff
@@ -11,6 +11,7 @@ import {
   transferPolyX,
   TransferPolyXParams,
 } from '~/api/procedures';
+import { revokeClaims, RevokeClaimsParams } from '~/api/procedures/revokeClaims';
```
monitz87: This should be imported from `~/api/procedures`. If it's not exported from there, please add the export (my bad)

src/api/entities/SecurityToken/Transfers.ts
```diff
@@ -69,10 +69,34 @@ export class Transfers extends Namespace<SecurityToken> {
    * @param args.to - receiver identity
    * @param args.amount - amount of tokens to transfer
    */
-  public async canTransfer(args: {
+  public canTransfer(args: {
     from?: string | Identity;
     to: string | Identity;
     amount: BigNumber;
+  }): Promise<TransferStatus> {
+    const { context } = this;
```
monitz87, line 77: No need to assign context to a variable since you're only using it once

src/api/procedures/__tests__/issueTokens.ts
```diff
@@ -134,7 +134,7 @@ describe('issueTokens procedure', () => {
     );
   });
 
-  test('shoud throw an error if at least one identity id has not a canTransfer success status', () => {
+  test('shoud throw an error if at least one identity id has not a canMint success status', () => {
```
monitz87, line 137: Should throw an error if canMint returns a status different from Success

src/testUtils/mocks/entities.ts
```diff
@@ -519,6 +524,18 @@ export function getSecurityTokenTransfersCanTransferStub(status?: TransferStatus
   return securityTokenTransfersCanTransferStub;
 }
 
+/**
+ * @hidden
+ * Retrieve the stub of the `SecurityToken.Transfers.canMint` method
+ */
+export function getSecurityTokenTransfersCanMintStub(status: TransferStatus): SinonStub {
```
monitz87, line 531: Status should be optional

PR #87, opened 2020-05-08
feat: update to latest polymesh specification

monitz87:


monitz87:
Will force merge because this is blocking everything else

PR #97, opened 2020-05-22
feat: fetch all the claims I made

shuffledex:


src/testUtils/mocks/polkadot.ts
```diff
@@ -622,6 +627,26 @@ export function createTxStub<
...
+ * @hidden
+ * Create and return an apollo query stub
+ *
+ * @param query - apollo document node
+ * @param returnValue
+ */
+export function createApolloQueryStub(query: DocumentNode, returnData: any): MockApolloClient {
+  const instance = mockInstanceContainer.apolloInstance;
+
+  instance.query = sinon
```
monitz87, line 640: What is the point of using 'mock-apollo-client' if you're just going to replace the query function with a sinon stub?

src/testUtils/mocks/apollo.ts
```diff
@@ -0,0 +1,32 @@
+import { InMemoryCache } from 'apollo-cache-inmemory';
```
monitz87, line 1: Logic related to the mock apollo client can't be split between this file and the polkadot one. I suggest renaming testUtils/mocks/polkadot.ts to testUtils/mocks/dataSources.ts and moving this logic over there. For the constructor, use the same logic that is being used in that file (see keyringConstructorStub)

monitz87: (While you're at it, rename all references to polkadotMockUtils to dsMockUtils)

src/testUtils/mocks/polkadot.ts
```diff
@@ -750,6 +775,15 @@ export function updateTxStatus<
   txMockData.statusCallback(statusToReceipt(status, failReason));
 }
 
+/**
+ * @hidden
+ * Make calls to `Harvester.query` throw an error
+ */
+export function throwOnHarvesterQuery(): void {
+  const instance = mockInstanceContainer.apolloInstance;
+  instance.query = createErrStub;
```
monitz87, line 784: rename createErrStub to errorStub

src/harvester/__tests__/queries.ts
```diff
@@ -1,38 +1,9 @@
-import { didsWithClaims, gqlArray, gqlNumber, gqlString } from '../queries';
+import { didsWithClaims } from '../queries';
 
 describe('didsWithClaims', () => {
-  test('should retrieve a graphql ast object', () => {
+  test('should return a GraphqlQuery object', () => {
```
monitz87, line 4: This test should verify that the variables are the same that were passed to the query

src/harvester/queries.ts
```diff
@@ -1,67 +1,53 @@
 import { DocumentNode } from 'graphql';
 import gql from 'graphql-tag';
 
-import { Maybe, QueryDidsWithClaimsArgs } from '~/harvester/types';
+import { QueryDidsWithClaimsArgs } from '~/harvester/types';
 
 interface GraphqlQuery {
```
monitz87, line 6: This interface should be generic and take the type of the variables as a parameter

src/testUtils/mocks/dataSources.ts
```diff
@@ -111,6 +111,17 @@ const MockKeyringClass = class {
   }
 };
 
+let apolloConstructurStub: SinonStub;
```
monitz87, line 114: typo

src/testUtils/mocks/dataSources.ts
```diff
@@ -657,7 +683,10 @@ export function createTxStub<
...
  * @param returnValue
  */
-export function createApolloQueryStub(query: DocumentNode, returnData: any): MockApolloClient {
+export function createApolloQueryStub(
+  query: DocumentNode,
+  returnData: any
+): ApolloClient<NormalizedCacheObject> {
   const instance = mockInstanceContainer.apolloInstance;
 
   instance.query = sinon
```
monitz87, line 692: This is not doing what you think it does. Sinon has obscure behavior when chaining calls like this (more info: https://nikas.praninskas.com/javascript/2015/07/28/quickie-sinon-withargs-not-working/)

The proper way of doing it is to assign sinon.stub() to a separate variable, call .withArgs and .resolves on that variable, and then assign the variable to instance.query. This will break your tests because you're expecting the .query stub to receive a DocumentNode object, which is not what is actually being passed to it in the code.

Also, createApolloQueryStub should return the stub, not the instance

Finally, `cleanup` in this file should also reset the apollo client instance

src/testUtils/mocks/dataSources.ts
```diff
@@ -848,6 +877,14 @@ export function getApiInstance(): ApiPromise & EventEmitter {
   return mockInstanceContainer.apiInstance as ApiPromise & EventEmitter;
 }
 
+/**
+ * @hidden
+ * Retrieve an instance of the mocked Apollo Client
+ */
+export function getHarvesterClient(): ApolloClient<NormalizedCacheObject> {
+  return (apolloConstructurStub as unknown) as ApolloClient<NormalizedCacheObject>;
```
monitz87, line 885: This should return the mocked instance, not the constructor stub

src/utils/index.ts
```diff
@@ -695,35 +695,29 @@ export function claimToMeshClaim(claim: Claim, context: Context): MeshClaim {
...
-    return ClaimType.Jurisdiction;
-  }
-  if (claim === 'Whitelisted') {
-    return ClaimType.Whitelisted;
-  }
-  if (claim === 'Blacklisted') {
-    return ClaimType.Blacklisted;
-  }
-  return ClaimType.NoData;
+export function createClaim(
```
monitz87, line 698:
```
export function createClaim(
  claimType: string,
  jurisdiction?: string | null,
  scope?: string | null
): Claim {
  const type = claimType as ClaimType;
  if (type === ClaimType.Jurisdiction) {
    return {
      type,
      name: jurisdiction as string,
      scope: scope as string,
    };
  } else if (type !== ClaimType.NoData && type !== ClaimType.CustomerDueDiligence) {
    return {
      type,
      scope: scope as string,
    };
  }

  return { type };
}
```

Replace it with this

For this to work though, the `Claim` type in types/index has to be replaced with 
```
export type Claim =
  | { type: ClaimType.Jurisdiction; name: string; scope: string }
  | { type: Exclude<ClaimType, ClaimType.NoData | ClaimType.Jurisdiction>; scope: string }
  | { type: ClaimType.NoData | ClaimType.CustomerDueDiligence };
```

src/harvester/queries.ts
```diff
@@ -3,17 +3,19 @@ import gql from 'graphql-tag';
 
 import { QueryDidsWithClaimsArgs } from '~/harvester/types';
 
-interface GraphqlQuery {
+export interface GraphqlQuery<VariablesType> {
```
monitz87, line 6: Should just be `Variables`

monitz87: Also, this interface should be defined in `types/internal.ts`

src/testUtils/mocks/dataSources.ts
```diff
@@ -82,11 +82,23 @@ function createApi(): Mutable<ApiPromise> & EventEmitter {
...
 }
 
+const apolloConstructorStub = sinon.stub();
+
+const MockApolloClientClass = class {
+  /**
+   * @hidden
+   */
+  public constructor() {
+    // apolloConstructorStub = sinon.stub();
```
monitz87, line 97: Delete this comment

PR #103, opened 2020-05-28
Feat/msdk 135 graphql api key

shuffledex:


src/Polymesh.ts
```diff
@@ -76,37 +104,52 @@ export class Polymesh {
       });
 
       harvesterClient = new ApolloClient({
```
monitz87, line 106: I think it makes more sense to have `harvesterClient` here defined as `ApolloClient<NormalizedCacheObject> | null` and initialized as null. You then only instance the apollo client and re-assign it to that variable IF the harvester config was passed, and instead of having the `isApolloConfigured` variable, the `Context.create` function should just deduce that from the harvesterClient being null or not

monitz87: Behind the scenes in the Context class, you save the harvester client in a private _harvesterClient variable, and define

```
public get harvesterClient()
```

Which throws an error if the variable isn't defined and returns it if it is.

To the user it's totally transparent and we don't even need to use the Proxy pattern

src/Polymesh.ts
```diff
@@ -78,6 +79,21 @@ export class Polymesh {
 
   static async connect(params: ConnectParamsBase & { accountUri: string }): Promise<Polymesh>;
 
+  static async connect(
```
monitz87, line 82: These types are slightly off. The SDK should allow connecting without any seed, keyring or URI but still connect to the middleware. These signatures don't allow that. You should put the harvester config as an optional property of `ConnectParamsBase` and remove it from these overloads

src/context/index.ts
```diff
@@ -51,14 +51,18 @@ export class Context {
 
   private currentIdentity?: Identity;
 
-  public harvesterClient: ApolloClient<NormalizedCacheObject>;
+  private _harvesterClient?: ApolloClient<NormalizedCacheObject>;
```
monitz87, line 54: If you define this as `private _harvesterClient: ApolloClient<NormalizedCacheObject> | null;` you don't have to wrap the assignment in line 63 in an if clause

src/context/__tests__/index.ts
```diff
@@ -36,6 +37,68 @@ describe('Context class', () => {
     );
   });
 
+  test('should throw an error if accessing the harvester query submodule without an active connection', async () => {
```
monitz87, line 40: ... if accessing the harvester client without an active connection

src/context/__tests__/index.ts
```diff
@@ -36,6 +37,68 @@ describe('Context class', () => {
...
+      ),
+    });
+
+    const context = await Context.create({
+      polymeshApi: dsMockUtils.getApiInstance(),
+      harvesterClient: null,
+      seed: 'Alice'.padEnd(32, ' '),
+    });
+
+    expect(() => context.harvesterClient.query).toThrow(
```
monitz87, line 65: The .query isn't necessary here

src/context/__tests__/index.ts
```diff
@@ -36,6 +37,68 @@ describe('Context class', () => {
...
+      'Cannot perform this action without an active harvester connection'
+    );
+  });
+
+  test('should return an ApolloClient instance if harvester credentials are set', async () => {
```
monitz87, line 70: This test should just check if context.harvesterClient is equal to the instance passed to the constructor (no need to stub a query and call it

PR #105, opened 2020-05-29
feat: get creation token event id

shuffledex:


src/Polymesh.ts
```diff
@@ -459,6 +460,40 @@ export class Polymesh {
     });
   }
 
+  /**
+   * Retrieve the event id of the token creation transaction
+   *
+   * @param args.ticker - Security Token ticker
+   */
+  public async getCreationTokenEventId(args: { ticker: string }): Promise<number | null> {
```
monitz87, line 468: This shouldn't be exposed in the client root. My bad for not specifying the interface in the task.

This should be part of a `createdAt` function in the SecurityToken entity root. The function should return an object that conforms to

```
// types/index.ts
interface EventIdentifier {
  blockNumber: number;
  eventIndex: number;
}
```

(eventIndex is event_idx in the query results)

src/Polymesh.ts
```diff
@@ -459,6 +460,40 @@ export class Polymesh {
...
+      );
+    } catch (e) {
+      throw new PolymeshError({
+        code: ErrorCode.FatalError,
+        message: `Error in harvester query: ${e.message}`,
+      });
+    }
+
+    if (result.data.eventByIndexedArgs) {
+      return result.data.eventByIndexedArgs.block_id || null;
```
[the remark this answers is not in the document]
monitz87: In the meantime, use a non-null assertion (!)

src/__tests__/Polymesh.ts
```diff
@@ -504,6 +505,88 @@ describe('Polymesh Class', () => {
...
   });
 
+  describe('method: getCreationTokenEventId', () => {
+    test('should return the event id of the token creation', async () => {
+      const ticker = 'SOMETICKER';
+      const blockId = 1234;
+      const variables = {
+        moduleId: 'asset',
+        eventId: 'AssetCreated',
+        eventArg1: ticker + '\u0000'.repeat(MAX_TICKER_LENGTH - ticker.length),
```
monitz87, line 515: Should use padTicker here

src/Polymesh.ts
```diff
@@ -459,6 +460,40 @@ export class Polymesh {
...
+      result = await harvesterClient.query<Query>(
+        eventByIndexedArgs({
+          moduleId: 'asset',
+          eventId: 'AssetCreated',
+          eventArg1: ticker + '\u0000'.repeat(MAX_TICKER_LENGTH - ticker.length),
+        })
+      );
+    } catch (e) {
+      throw new PolymeshError({
+        code: ErrorCode.FatalError,
```
monitz87, line 485: I think we should have a specific error code for grapqhl errors. Something like `ErrorCode.MiddlewareError`

src/harvester/__tests__/queries.ts
```diff
@@ -17,3 +17,18 @@ describe('didsWithClaims', () => {
     expect(result.variables).toEqual(variables);
   });
 });
+
+describe('eventByIndexedArgs', () => {
+  test('should verify that the variables are the same that were passed to the query', () => {
```
monitz87, line 22: "should pass the variables to the grapqhl query"

(change the other test to say the same)

src/api/entities/SecurityToken/index.ts
```diff
@@ -185,4 +189,42 @@ export class SecurityToken extends Entity<UniqueIdentifiers> {
 
     return tokenIdentifiers;
   }
+
+  /**
+   * Retrieve the event identifier of the token creation transaction
```
monitz87, line 194:
```
Retrieve the identifier data (block number and event index) of the event that was emitted when the token was created

@note This data is harvested from the chain and stored in a database, so there is a possibility that the data is not ready by the time it is requested. In that case, `null` is returned
```

src/api/entities/SecurityToken/index.ts
```diff
@@ -185,4 +189,42 @@ export class SecurityToken extends Entity<UniqueIdentifiers> {
...
+        })
+      );
+    } catch (e) {
+      throw new PolymeshError({
+        code: ErrorCode.MiddlewareError,
+        message: `Error in harvester query: ${e.message}`,
+      });
+    }
+
+    // TODO remove null check once types fixed
```
monitz87, line 218: This TODO should be inside the if clause

PR #121, opened 2020-06-22
feat: get network name and runtime version

shuffledex:


src/context/index.ts
```diff
@@ -326,6 +327,21 @@ export class Context {
     return invalidDids;
   }
 
+  /**
+   * Retrieve information for the current network
+   */
+  public getNetworkProperties(): { name: string; version: BigNumber } {
```
monitz87, line 333: Return type should be defined in types as `NetworkProperties`. Version should be a number

There are multiple names apparently. You have `specName`, `implName` and the chain name you get by calling `api.rpc.system.chain`. They are not the same. For example, in pme, you have

```
specName = polymesh
implName = polymath-polymesh
chain = Development (in aldebaran this is Polymesh Aldebaran Testnet)
```

IMO the `chain` value is more significant, so we should return that

src/testUtils/mocks/dataSources.ts
```diff
@@ -950,7 +952,7 @@ const createMockCodec = (codec: object, isEmpty: boolean): Codec => {
 /**
  * @hidden
  */
-const createMockStringCodec = (value?: string): Codec =>
+export const createMockStringCodec = (value?: string): Codec =>
```
monitz87, line 955: This shouldn't be exported. You should create a `createMockTextCodec` that uses this and export that

src/context/__tests__/index.ts
```diff
@@ -1,3 +1,4 @@
+import BigNumber from 'bignumber.js';
```
monitz87, line 1: Unnecessary import

src/testUtils/mocks/dataSources.ts
```diff
@@ -1578,3 +1578,6 @@ export const createMockCanTransferResult = (
 export const setRuntimeVersion = (args: unknown): void => {
   mockInstanceContainer.apiInstance.runtimeVersion = args as RuntimeVersion;
 };
+
+export const createMockText = (value: string): Text =>
+  (createMockStringCodec(value) as unknown) as Text;
```
monitz87, line 1583: I think the `as unknown` isn't necessary

shuffledex: If I remove the cast: `Conversion of type 'Codec' to type 'Text' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

PR #122, opened 2020-06-26
feat: add canMint to tokenholders data

shuffledex:


src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -11,10 +14,22 @@ import { SecurityToken } from './';
  * Handles all Security Token Holders related functionality
  */
 export class TokenHolders extends Namespace<SecurityToken> {
+  public async get(opts: { pagination: PaginationOptions }): Promise<ResultSet<IdentityBalance>>;
+
+  public async get(opts: { mintStatus: boolean }): Promise<ResultSet<IdentityBalance>>;
```
monitz87, line 19: These overloads are incorrect.

First, you should be able to pass pagination options regardless of whether you pass other options. Also, the typings should reflect the fact that if you don't pass `mintStatus`, you get an object without the `canMint` property.

So, something like

```
// SecurityToken/types.ts

interface TokenHolderOptions {
  canBeIssuedTo: boolean;
}

interface TokenHolderProperties {
  canBeIssuedTo: boolean;
}

// This file

public async get(opts: Pick<TokenHolderOptions, 'canBeIssuedTo'>, paginationOpts?: PaginationOptions): Promise<ResultSet<IdentityBalance & Pick<TokenHolderProperties, 'canBeIssuedTo'>>>;

public async get(paginationOpts?: PaginationOptions): Promise<ResultSet<IdentityBalance>>;
```

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
-      balance: balanceToBigNumber(balance),
-    }));
+    let data: IdentityBalance[] = [];
+
+    const securityToken = new SecurityToken({ ticker }, context);
+
+    const areFrozen = opts?.mintStatus ? await securityToken.transfers.areFrozen() : undefined;
+
+    if (opts?.mintStatus && !areFrozen) {
+      const entriesChunks = chunk(entries, 10);
```
monitz87, line 55: There is a similar statement with a hardcoded 10 in `procedures/issueTokens.ts`. Please refactor both numbers into a `MAX_CONCURRENT_REQUESTS` variable in `constants.ts` and change it to a bigger number (something like 200 will do for now)

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
+
+      await Promise.all(
```
monitz87, line 57: This one (and the one in issueTokens as well) should use Bluebird's `.each` here instead of Promise.all (my bad). The idea is to process each chunk serially, otherwise there is no point in separating them into chunks in the first place. The one in the inside is fine, as we want all elements in each chunk to be processed simultaneously

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
+        entriesChunks.map(async entriesChunk => {
+          const canMint = await Promise.all(
+            entriesChunk.map(([storageKey]) => {
+              return securityToken.transfers.canMint({
```
monitz87, line 61: You can remove the `return` here

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
+    const securityToken = new SecurityToken({ ticker }, context);
+
+    const areFrozen = opts?.mintStatus ? await securityToken.transfers.areFrozen() : undefined;
+
+    if (opts?.mintStatus && !areFrozen) {
+      const entriesChunks = chunk(entries, 10);
+
+      await Promise.all(
+        entriesChunks.map(async entriesChunk => {
+          const canMint = await Promise.all(
```
monitz87, line 59: transferStatuses

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
+            })
+          );
+
+          entriesChunk.forEach(([storageKey, balance], index) => {
+            data.push({
+              identity: new Identity(
+                { did: identityIdToString(storageKey.args[1] as IdentityId) },
+                context
+              ),
+              balance: balanceToBigNumber(balance),
```
monitz87, line 74: All this should be only done once outside the if/else and then you can chunk `data` and process that instead

src/api/entities/SecurityToken/__tests__/TokenHolders.ts
```diff
@@ -38,33 +73,102 @@ describe('TokenHolders class', () => {
...
+
+      const token = entityMockUtils.getSecurityTokenInstance();
+      const tokenHolders = new TokenHolders(token, context);
+
+      const result = await tokenHolders.get();
+
+      expect(result).toEqual({ data: expectedHolders, next: null });
+    });
 
+    test('should retrieve all the token holders with balance and mint status in false', async () => {
```
monitz87, line 112: should have canBeIssuedTo set to false for all holders if transfers are frozen

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -27,16 +42,56 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
-        { did: identityIdToString(storageKey.args[1] as IdentityId) },
-        context
-      ),
-      balance: balanceToBigNumber(balance),
-    }));
+    let data: IdentityBalance[] = [];
+
+    const securityToken = new SecurityToken({ ticker }, context);
+
+    const areFrozen = opts?.mintStatus ? await securityToken.transfers.areFrozen() : undefined;
```
monitz87, line 52: - You should just default opts to `{ mintStatus: false }` so you can destructure it at the beginning and avoid having to use null conditional operators
- You should use a nested if inside the one below. The case where you requested the mint status and transfers are frozen is different from the case where you didn't request the mint status

src/api/entities/SecurityToken/__tests__/TokenHolders.ts
```diff
@@ -38,33 +73,102 @@ describe('TokenHolders class', () => {
...
+
+      const token = entityMockUtils.getSecurityTokenInstance();
+      const tokenHolders = new TokenHolders(token, context);
+
+      const result = await tokenHolders.get({ mintStatus: true });
+
+      expect(result).toEqual({ data: expectedHolders, next: null });
+    });
+
+    test('should retrieve all the token holders with balance and canMint attribute setting in true', async () => {
```
monitz87, line 156: should return whether each token holder can receive primary issuance to if canBeIssuedTo is passed as true

src/api/entities/SecurityToken/types.ts
```diff
@@ -16,4 +16,5 @@ export interface SecurityTokenDetails {
 export interface IdentityBalance {
   identity: Identity;
   balance: BigNumber;
+  canMint?: boolean;
```
monitz87, line 19: Should not be a property of IdentityBalance (see first comment)

src/api/entities/SecurityToken/TokenHolders.ts
```diff
@@ -41,31 +41,44 @@ export class TokenHolders extends Namespace<SecurityToken> {
...
+        ),
+        balance: balanceToBigNumber(balance),
+      };
+      return entry;
+    });
 
     if (canBeIssuedTo) {
-      if (!areFrozen) {
-        const entriesChunks = chunk(entries, MAX_CONCURRENT_REQUESTS);
+      const entriesChunks = chunk(entries, MAX_CONCURRENT_REQUESTS);
```
monitz87, line 75: You should chunk data, not entries

PR #123, opened 2020-06-30
[DO NOT MERGE] Feat/msdk 150 filter authorizations

shuffledex:


monitz87:
You can't have pagination and filtering at the same time, since you'd get less (or possibly none) elements per page than requested. I've already asked @adamdossa if it's possible to add pagination to the upcoming RPC. In the meantime, we should leave this task in standby

PR #126, opened 2020-07-06
feat: add fee getters to queue and transactions

monitz87:


src/base/__tests__/PolymeshTransaction.ts
```diff
@@ -308,4 +355,195 @@ describe('Polymesh Transaction class', () => {
...
+    //   ): Promise<string> {
+    //     this.addTransaction(tx1, {}, args.ticker);
+    //     this.addTransaction(tx2, {}, args.signingItems);
+    //     return returnValue;
+    //   };
+    //   const proc = new Procedure(func);
+    //   await expect(proc.prepare(procArgs, context)).rejects.toThrow(
+    //     'Did not set batch size for batch transaction. Please report this error to the Polymath team'
+    //   );
+    // });
```
VictorVicente, line 547: Any reason to keep this as commented?

src/base/TransactionQueue.ts
```diff
@@ -143,6 +158,24 @@ export class TransactionQueue<
     return res;
   }
 
+  /**
+   * Retrieves a lower bound of the fees required to execute this transaction queue
```
VictorVicente, line 162: Could you add why fees could be greater?

PR #148, opened 2020-07-27
feat: check if an identity has voted the proposal

shuffledex:


src/api/entities/Proposal/index.ts
```diff
@@ -47,6 +48,50 @@ export class Proposal extends Entity<UniqueIdentifiers> {
     this.pipId = pipId;
   }
 
+  /**
+   * Check if an identity has voted the proposal
```
monitz87, line 52: ... voted on the proposal

src/api/entities/Proposal/__tests__/index.ts
```diff
@@ -43,6 +44,54 @@ describe('Proposal class', () => {
...
+          block_id: 'someBlockId',
+        },
+        /* eslint-enable @typescript-eslint/camelcase */
+      });
+
+      const result = await proposal.identityHasVoted();
+      expect(result).toEqual(fakeResult);
+    });
+
+    test('should return false if the identity has not voted the proposal', async () => {
```
monitz87, line 80: voted on the proposal

PR #151, opened 2020-07-31
Feat/add count to GQL queries

VictorVicente:
Paginated GQL queries return a ResultSet including the total amount of records for that query, the
next key if there are more pages and the data from the query.

BREAKING CHANGE: 

- getIssuedClaims and getIdentitiesWithClaims return ResultSet<IdentityWithClaims> instead of IdentityWithClaims[]

### DIFF
--- a/src/utils/__tests__/index.ts
+++ b/src/utils/__tests__/index.ts
@@ -2123,7 +2123,7 @@ describe('batchArguments', () => {
 });
 
 describe('calculateNextKey', () => {
-  test('should return NextKey null as there are less element than default page size', () => {
+  test('should return NextKey null as there are less elements than the default page size', () => {
     const totalCount = 20;
     const nextKey = calculateNextKey(totalCount);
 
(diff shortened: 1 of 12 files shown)


src/utils/__tests__/index.ts
```diff
@@ -2120,3 +2121,30 @@ describe('batchArguments', () => {
     );
   });
 });
+
+describe('calculateNextKey', () => {
+  test('should return NextKey null as there are less element than default page size', () => {
```
monitz87, line 2126:
```suggestion
  test('should return NextKey null as there are less elements than the default page size', () => {
```

src/utils/__tests__/index.ts
```diff
@@ -2120,3 +2121,30 @@ describe('batchArguments', () => {
...
+    const totalCount = 20;
+    const nextKey = calculateNextKey(totalCount);
+
+    expect(nextKey).toBeNull();
+  });
+
+  test('should return NextKey null as it is latest page', () => {
```
monitz87, line 2133:
```suggestion
  test('should return NextKey null as it is the last page', () => {
```

PR #153, opened 2020-08-04
feat: add `Governance.getTransactionArguments`

monitz87:
This function receives a transaction tag and returns a list of the transaction's arguments and their
types

### DIFF
--- a/src/context/index.ts
+++ b/src/context/index.ts
@@ -1,19 +1,28 @@
 import { ApiPromise, Keyring } from '@polkadot/api';
+import { getTypeDef } from '@polkadot/types';
 import { AccountInfo } from '@polkadot/types/interfaces';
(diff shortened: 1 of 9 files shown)


src/context/index.ts
```diff
@@ -355,6 +365,212 @@ export class Context {
...
+
+    const balanceTypes = ['Amount', 'AssetOf', 'Balance', 'BalanceOf'];
+
+    const numberTypes = ['u8', 'u16', 'u32', 'u64', 'u128', 'u256', 'U256', 'BlockNumber'];
+
+    const textTypes = ['String', 'Text', 'Ticker'];
+
+    const booleanTypes = ['bool'];
+
+    const dateTypes = ['Moment'];
```
VictorVicente, line 402: Should we move these to `constants.ts`?

PR #158, opened 2020-08-10
feat: implement identity balance methods

monitz87:


src/Polymesh.ts
```diff
@@ -214,16 +214,12 @@ export class Polymesh {
...
-  /*
-   * Get the POLYX balance of the current account
-   * NOTE: We don't expose this method for Testnet v1
+  /**
+   * Get the POLYX balance of an identity
    */
-  /*
-  public getIdentityBalance(): Promise<BigNumber> {
-    return this.context.getCurrentIdentity().getPolyXBalance();
+  public getIdentityBalance(args?: { did: string }): Promise<BigNumber> {
```
VictorVicente, line 220: I know that if you have the Identity, you could call directly to getPolyXBalance or getTokenBalance. But should `did` support `Identity` entity too? To keep consistence.

monitz87: Agreed. Will fix

PR #160, opened 2020-08-11
feat: add `getTrustingTokens` to Identity Entity

monitz87:


src/api/entities/Identity/index.ts
```diff
@@ -255,4 +259,23 @@ export class Identity extends Entity<UniqueIdentifiers> {
...
+
+    const { order } = args;
+
+    const {
+      data: { tokensByTrustedClaimIssuer: tickers },
+    } = await context.queryMiddleware<Ensured<Query, 'tokensByTrustedClaimIssuer'>>(
+      tokensByTrustedClaimIssuer({ claimIssuerDid: did, order })
+    );
+
+    return tickers.map(ticker => new SecurityToken({ ticker }, context));
```
VictorVicente, line 279: Tickers from harvester comes with `\u0000` to complete 12 characters. You should trim them off to return tickers.

PR #161, opened 2020-08-12
feat: getCddClaims

shuffledex:


src/context/index.ts
```diff
@@ -546,6 +554,78 @@ export class Context {
...
+          count: size,
+          skip: start,
+        })
+      );
+    } catch (e) {
+      throw new PolymeshError({
+        code: ErrorCode.FatalError,
+        message: `Error in middleware query: ${e.message}`,
+      });
+    }
```
monitz87, line 597: Use the new way of querying (context.queryMiddleware)

PR #162, opened 2020-08-13
Feat/msdk 219 get transaction history

shuffledex:


src/types/index.ts
```diff
@@ -222,6 +223,18 @@ export interface IdentityWithClaims {
   claims: ClaimData[];
 }
 
+export interface ExtrinsicData {
```
monitz87, line 226: Why are these values optional?

src/types/index.ts
```diff
@@ -3,6 +3,7 @@ import { IKeyringPair, TypeDef } from '@polkadot/types/types';
 import BigNumber from 'bignumber.js';
 
 import { Identity } from '~/api/entities';
+import { TxTag } from '~/polkadot';
```
monitz87, line 6:
```suggestion
import { TxTag } from 'polymesh-types/types';
```

src/utils/__tests__/index.ts
```diff
@@ -2099,6 +2102,28 @@ describe('txTagToProtocolOp', () => {
   });
 });
 
+describe('txTagToExtrinsicIdentifier and extrinsicIdentifierToTxTag', () => {
```
monitz87, line 2105: Use examples where both the module and the call names have more than one word in them

src/Polymesh.ts
```diff
@@ -665,15 +665,16 @@ export class Polymesh {
...
    *
    * @param opts.address - account that signed the transaction
+   * @param opts.tag - tag associated with the transaction
    * @param opts.success - whether the transaction was successful or not
    * @param opts.size - page size
    * @param opts.start - page offset
    */
   public async getTransactionHistory(
-    opts: {
+    filters: {
```
monitz87, line 674: Change this back to opts or change the comments to reflect this new name

src/Polymesh.ts
```diff
@@ -682,20 +683,23 @@ export class Polymesh {
...
 
-    let extrinsicIdentifier;
-    if (txTag) {
-      extrinsicIdentifier = txTagToExtrinsicIdentifier(txTag);
+    let moduleId;
+    let callId;
+    if (tag) {
+      const extrinsicIdentifier = txTagToExtrinsicIdentifier(tag);
+      moduleId = extrinsicIdentifier.moduleId;
+      callId = extrinsicIdentifier.callId;
```
monitz87, line 693: You can just destructure here

src/types/index.ts
```diff
@@ -224,15 +224,15 @@ export interface IdentityWithClaims {
 }
 
 export interface ExtrinsicData {
-  blockId?: number;
-  extrinsicIdx?: number;
+  blockId: number;
+  extrinsicIdx: number;
   address?: string | null;
```
monitz87, line 229: Why is this value still optional (or null)?

shuffledex: address can be null if the user searches for system transactions. @VictorVicente can add more details about it.

monitz87: Then have it just be `string | null`. Doesn't make sense for it to be undefined as well

PR #164, opened 2020-08-14
feat: ticker validation

shuffledex:


src/Polymesh.ts
```diff
@@ -279,6 +279,15 @@ export class Polymesh {
    * @param args.ticker - ticker symbol to reserve
    */
   public reserveTicker(args: ReserveTickerParams): Promise<TransactionQueue<TickerReservation>> {
+    const { ticker } = args;
+
+    if (ticker.length < 1 || ticker.length > 12 || ticker !== ticker.toUpperCase()) {
```
monitz87, line 284: These validations should go in the procedure

src/Polymesh.ts
```diff
@@ -279,6 +279,15 @@ export class Polymesh {
...
+      throw new PolymeshError({
+        code: ErrorCode.ValidationError,
+        message: 'The ticker symbol does not conform to the standards',
```
monitz87, line 287: The ticker must be between 1 and 12 characters and cannot contain lower case letters

PR #167, opened 2020-08-17
feat: change inputs for `addClaims`, `editClaims` and `revokeClaims`

monitz87:
These functions now receive an array of `ClaimTarget` (an object with a target DID, a claim and an
optional expiry)

BREAKING CHANGE: The `ClaimTargets` interface has been replaced with `ClaimTarget`. Instead of
representing a set of identities with one claim, it now represents a single identity/claim pair

### DIFF
--- a/src/base/__tests__/PolymeshError.ts
+++ b/src/base/__tests__/PolymeshError.ts
@@ -1,6 +1,7 @@
+import { Polymesh } from '~/Polymesh';
 import { ErrorCode } from '~/types';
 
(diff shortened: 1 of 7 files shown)


src/base/__tests__/PolymeshError.ts
```diff
@@ -1,6 +1,7 @@
+import { Polymesh } from '~/Polymesh';
```
VictorVicente, line 1: This can be removed

PR #172, opened 2020-08-20
Feat/msdk 224 claims grouped

shuffledex:


src/middleware/queries.ts
```diff
@@ -273,3 +274,35 @@ export function scopesByIdentity(
...
   };
 }
+
+/**
+ * @hidden
+ *
+ * Get issuer dids with at least one claim for given target
+ */
+export function issuerDidsWithClaimsByTarget(
+  variables?: QueryIssuerDidsWithClaimsByTargetArgs
```
monitz87, line 284: These can't be undefined, since target has to be present always

src/middleware/queries.ts
```diff
@@ -281,7 +281,7 @@ export function scopesByIdentity(
  * Get issuer dids with at least one claim for given target
  */
 export function issuerDidsWithClaimsByTarget(
-  variables?: QueryIssuerDidsWithClaimsByTargetArgs
+  variables: QueryIssuerDidsWithClaimsByTargetArgs
 ): GraphqlQuery<QueryIssuerDidsWithClaimsByTargetArgs | undefined> {
```
monitz87, line 285:
```suggestion
): GraphqlQuery<QueryIssuerDidsWithClaimsByTargetArgs> {
```

PR #176, opened 2020-08-24
Feat/msdk 172 get votes threshold

shuffledex:


src/api/entities/Proposal/index.ts
```diff
@@ -207,4 +209,37 @@ export class Proposal extends Entity<UniqueIdentifiers> {
...
+      },
+      pipId,
+    } = this;
+
+    const [stage, metadata] = await Promise.all([this.getStage(), pips.proposalMetadata(pipId)]);
+
+    const { end: endBlock } = metadata.unwrap();
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const opts: any = {
```
monitz87, line 232: The type here should be `{ args: []; blockHash?: string | BlockHash }`. Please modify the signature of `requestAtBlock` so that blockHash is typed as `blockHash?: string | BlockHash`

src/api/entities/Proposal/index.ts
```diff
@@ -207,4 +209,37 @@ export class Proposal extends Entity<UniqueIdentifiers> {
...
+      args: [],
+    };
+
+    if (stage !== ProposalStage.Open) {
+      const blockHash = await chain.getBlockHash(u32ToBigNumber(endBlock).toString());
```
monitz87, line 237:
```suggestion
      const blockHash = await chain.getBlockHash(endBlock);
```

PR #179, opened 2020-08-25
Docs/mm 218/documentation changes

monitz87:


src/api/entities/Proposal/types.ts
```diff
@@ -28,12 +28,6 @@ export interface Metadata {
   totalVotes: number;
 }
 
-export interface Proposal {
-  call: Call;
-  metadata: Metadata;
-  state: ProposalState;
-}
-
```
VictorVicente, line 36: Why is this deleted?

shuffledex: I added it at the beginning of the Proposal entity but we never used it

PR #183, opened 2020-08-28
feat: add includeExpired opts

shuffledex:


src/Polymesh.ts
```diff
@@ -558,6 +558,7 @@ export class Polymesh {
    * @param opts.trustedClaimIssuers - identity IDs of claim issuers. Defaults to all claim issuers
    * @param opts.scope - scope of the claims to fetch. Defaults to any scope
    * @param opts.claimTypes - types of the claims to fetch. Defaults to any type
+   * @param opts.includeExpired - default to true
```
monitz87, line 561: whether to include expired claims. Defaults to true

src/api/entities/Identity/index.ts
```diff
@@ -372,19 +372,22 @@ export class Identity extends Entity<UniqueIdentifiers> {
   /**
    * Retrieve all claims issued about this identity, grouped by claim issuer
    *
+   * @param opts.includeExpired - default to true
```
monitz87, line 375: whether to include expired claims. Defaults to true

PR #188, opened 2020-09-02
Feat/msdk 255 middleware up to date event

monitz87:


src/base/TransactionQueue.ts
```diff
@@ -213,10 +224,34 @@ export class TransactionQueue<
...
+   * @param listener - callback function that will be called whenever the middleware is updated with the latest data.
+   *   If there is an error (timeout or middleware offline) it will be passed to this callback
+   *
+   * @returns unsubscribe function
+   * @throws if the middleware wasn't enabled when instantiating the SDK client
+   */
+  public onProcessedByMiddleware(listener: (err?: PolymeshError) => void): () => void {
+    if (!this.context.isMiddlewareEnabled()) {
+      throw new PolymeshError({
+        code: ErrorCode.FatalError,
```
VictorVicente, line 246: I'm not sure if `FatalError` is the best option here. Why not `MiddlewareError`?

src/utils/index.ts
```diff
@@ -102,6 +102,8 @@ import {
   SS58_FORMAT,
 } from '~/utils/constants';
 
+export { cryptoWaitReady } from '@polkadot/util-crypto';
```
VictorVicente, line 105: Why do we need to export this one?

VictorVicente (commented):
Left some comments

src/utils/index.ts
monitz87: They are using it in the dApp when loading keyrings

src/base/TransactionQueue.ts
```diff
@@ -159,6 +167,7 @@ export class TransactionQueue<
       throw err;
     } finally {
       this.hasRun = true;
+      this.emitWhenMiddlewareIsSynced();
```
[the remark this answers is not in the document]
monitz87: As discussed during standup, in a queue with more than one transaction, even if the queue fails, there might be changes in the chain, and so you might want to know when they are reflected in the middleware. I'll add a comment in the subscribe function's docs explaining this

src/base/TransactionQueue.ts
monitz87: It's not actually the middleware that is failing here. It's just a user error (akin to passing a string to a function that requires a number) so I figured fatal error is a better fit

PR #189, opened 2020-09-03
Feat/msdk 249 is archive node

shuffledex:


src/context/index.ts
```diff
@@ -170,6 +171,28 @@ export class Context {
     return new Context({ polymeshApi, middlewareApi, keyring });
   }
 
+  /**
+   * Get if the current node is archive or not
```
monitz87, line 175:
```suggestion
   * Retrieve whether the current node is an archive node (contains a full history from genesis onward) or not
```

src/context/index.ts
```diff
@@ -170,6 +171,28 @@ export class Context {
...
+        query: { balances, system },
+      },
+    } = this;
+
+    try {
+      const blockHash = await system.blockHash(0);
+      await requestAtBlock(balances.totalIssuance, {
+        args: [],
+        blockHash,
+      });
```
monitz87, line 189: Can you double check with the core team whether you actually need to call a second function? It seems like the call to `blockHash` itself should be enough to know if the node is an archive node

src/context/index.ts
```diff
@@ -170,6 +171,28 @@ export class Context {
...
+   */
+  public async isCurrentNodeArchive(): Promise<boolean> {
+    const {
+      polymeshApi: {
+        query: { balances, system },
+      },
+    } = this;
+
+    try {
+      const blockHash = await system.blockHash(0);
```
monitz87, line 185: use `numberToU32` here

Also, in `Proposal.minimumBondedAmount` we're using rpc to get the block hash from block number. We should use `system.blockHash` there too for consistency

monitz87 (changes requested):

src/context/index.ts
shuffledex: Mudit told me to fetch something known and check that exact value is returned because it's not guaranteed that the node will return an error if it doesn't have the data. It might just return garbage instead an error

shuffledex: I'm going to rewrite some lines here

monitz87 (approved):

PR #195, opened 2020-09-07
Feat/msdk 260 modify treasury did

shuffledex:


monitz87:
docs shouldn't be in this branch

src/api/entities/SecurityToken/index.ts
```diff
@@ -283,4 +285,14 @@ export class SecurityToken extends Entity<UniqueIdentifiers> {
...
+
+  /**
+   * Sets a treasury DID for this token
+   *
+   * @param args.target - identity to set as treasury. Empty to reset it
+   */
+  public setTreasury(args: SetTreasuryDidParams): Promise<TransactionQueue<void>> {
+    const { ticker } = this;
+    return setTreasuryDid.prepare({ ticker, ...args }, this.context);
+  }
```
monitz87, line 297: Instead of having a separate procedure and endpoint, this should be part of modifyToken and the modify endpoint

monitz87 (changes requested):

PR #198, opened 2020-09-09
feat: fetch pending authorizations filtered by their type

shuffledex:


src/api/entities/Identity/Authorizations.ts
```diff
@@ -25,35 +27,51 @@ export class Authorizations extends Namespace<Identity> {
...
    * Fetch all pending authorization requests for which this identity is the target
    *
-   * @note supports pagination
+   * @param opts.includeExpired - whether to include expired authorizations. Defaults to true
    */
-  public async getReceived(
-    paginationOpts?: PaginationOptions
-  ): Promise<ResultSet<AuthorizationRequest>> {
+  public async getReceived(opts: {
+    filterByType: AuthorizationType;
```
monitz87, line 33: Should just be called `type`, and the docs should say `@param opts.type - fetch only authorizations of this type. Fetches all types if not passed`

src/api/entities/Identity/Authorizations.ts
```diff
@@ -25,35 +27,51 @@ export class Authorizations extends Namespace<Identity> {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const result: Authorization[] = await (rpc as any).identity.getFilteredAuthorizations(
+      signatory,
+      rawBoolean,
+      rawAuthorizationType
     );
 
     const data = this.createAuthorizationRequests(
-      entries.map(([, auth]) => ({ auth, target: did }))
+      // eslint-disable-next-line @typescript-eslint/camelcase
```
monitz87, line 70: This shouldn't be necessary

monitz87 (changes requested):

monitz87 (approved):

PR #200, opened 2020-09-10
feat: old blocks must throw an error if node is not archive

shuffledex:


src/context/index.ts
```diff
@@ -161,15 +164,19 @@ export class Context {
...
         middlewareApi,
         keyring,
         pair: currentPair,
       });
+    } else {
+      context = new Context({ polymeshApi, middlewareApi, keyring });
     }
 
-    return new Context({ polymeshApi, middlewareApi, keyring });
+    context.isArchiveNode = await context.isCurrentNodeArchive();
```
monitz87, line 177: Having `context.isCurrentNodeArchive()` as a public function along `context.isArchiveNode` is confusing. You should just make the relevant calls here and delete the function

monitz87: The comments from the `isCurrentNodeArchive` function should be adapted and used for the `isArchiveNode` variable

src/utils/index.ts
```diff
@@ -1229,11 +1229,18 @@ export async function requestAtBlock<F extends AnyFunction>(
...
+  },
+  context: Context
 ): Promise<ObsInnerType<ReturnType<F>>> {
   const { blockHash, args } = opts;
 
   if (blockHash) {
+    if (!context.isArchiveNode) {
+      throw new PolymeshError({
+        code: ErrorCode.QueryAborted,
+        message: 'The node must be archive to execute the query',
```
monitz87, line 1241: `Cannot query previous blocks in a non-archive node`

src/context/index.ts
```diff
@@ -184,10 +191,14 @@ export class Context {
...
-        args: [],
-        blockHash,
-      });
+      const balance = await requestAtBlock(
+        balances.totalIssuance,
+        {
+          args: [],
+          blockHash,
+        },
+        this
```
monitz87, line 200: This will throw an error because `isArchiveNode` is false when this is called for the first time

src/types/index.ts
```diff
@@ -304,6 +304,7 @@ export enum ErrorCode {
   NotAuthorized = 'NotAuthorized',
   MiddlewareError = 'MiddlewareError',
   IdentityNotPresent = 'IdentityNotPresent',
+  QueryAborted = 'QueryAborted',
```
monitz87, line 307: I think `DataUnavailable` would be a better fit for this case

monitz87 (changes requested):

src/context/__tests__/index.ts
```diff
@@ -105,6 +103,15 @@ describe('Context class', () => {
   });
 
   describe('method: create', () => {
+    beforeEach(() => {
+      dsMockUtils.createQueryStub('balances', 'totalIssuance', {
+        returnValue: dsMockUtils.createMockBalance(100),
+      });
+      dsMockUtils.createQueryStub('system', 'blockHash', {
+        returnValue: dsMockUtils.createMockHash(),
```
monitz87, line 111: Pass a value here

src/context/index.ts
```diff
@@ -77,6 +76,9 @@ export class Context {
 
   public currentPair?: KeyringPair;
 
+  /**
+   * Retrieve whether the current node is an archive node (contains a full history from genesis onward) or not
```
monitz87, line 80:
```suggestion
   * Whether the current node is an archive node (contains a full history from genesis onward) or not
```

monitz87 (changes requested):

monitz87 (approved):

PR #204, opened 2020-09-16
Feat/msdk 257 account entity

monitz87:


src/api/entities/Account.ts
```diff
@@ -0,0 +1,206 @@
...
+        stringToAccountId(address, context)
+      );
+      const did = identityIdToString(identityIdWrapper.unwrap().asUnique);
+
+      return new Identity({ did }, context);
+    } catch (err) {
+      throw new PolymeshError({
+        code: ErrorCode.IdentityNotPresent,
+        message: 'The current account does not have an associated identity',
+      });
```
VictorVicente, line 113: Should it throw? Maybe it should return null. Will it impact on how we are catching this problem where Identity is required?

monitz87:
Closed in favor of #205

PR #216, opened 2020-09-30
fix: add validations into numberToBalance utils method

shuffledex:


src/utils/index.ts
```diff
@@ -477,10 +479,47 @@ export function authorizationDataToAuthorization(auth: AuthorizationData): Autho
 /**
  * @hidden
  */
-export function numberToBalance(value: number | BigNumber, context: Context): Balance {
+export function numberToBalance(
+  value: number | BigNumber,
+  context: Context,
+  divisible?: boolean
```
monitz87, line 485: I know I said the default should be false, but I was mistaken. It should be true. Most amounts (such as POLYX amounts) have decimals, so divisibility is the sensible default

src/utils/__tests__/index.ts
```diff
@@ -399,11 +404,63 @@ describe('numberToBalance and balanceToBigNumber', () => {
       .withArgs('Balance', value.multipliedBy(Math.pow(10, 6)).toString())
       .returns(fakeResult);
 
-    const result = numberToBalance(value, context);
+    let result = numberToBalance(value, context);
```
monitz87, line 407: This test should explicity pass false

monitz87 (changes requested):

monitz87 (changes requested):
Also, `canSettle` should fetch the `isDivisible` value from the parent and pass it to the `numberToBalance`call

src/api/entities/SecurityToken/Settlements.ts
```diff
@@ -39,6 +39,8 @@ export class Settlements extends Namespace<SecurityToken> {
 
     const { from = await this.context.getCurrentIdentity(), to, amount } = args;
 
+    const { isDivisible } = await details();
```
monitz87, line 42: You should group both these awaits into a `Promise.all`

monitz87 (changes requested):

src/api/entities/SecurityToken/Settlements.ts
shuffledex: but we are going to make two request every time although `from` is set

monitz87:
```
const { to, amount } = args;
let { from } = args;
let isDivisible;

if (!from) {
  ([{ isDivisible }, from] = await Promise.all([details(), this.context.getCurrentIdentity()]);
} else {
  ({ isDivisible } = await details());
}
```

src/api/entities/SecurityToken/Settlements.ts
```diff
@@ -37,9 +37,15 @@ export class Settlements extends Namespace<SecurityToken> {
...
     } = this;
 
-    const { from = await this.context.getCurrentIdentity(), to, amount } = args;
+    const { to, amount } = args;
+    let { from } = args;
+    let isDivisible;
 
-    const { isDivisible } = await details();
+    if (!from) {
+      [{ isDivisible }, from] = await Promise.all([details(), this.context.getCurrentIdentity()]);
```
monitz87, line 45:
```suggestion
      [{ isDivisible }, from] = await Promise.all([details(), context.getCurrentIdentity()]);
```

monitz87 (changes requested):

monitz87 (approved):

PR #222, opened 2020-10-06
feat: getstatus method

shuffledex:


monitz87:
Status is already returned by `details`

PR #230, opened 2020-10-13
Feat/msdk 274 refactor instructions procedures

shuffledex:


src/api/procedures/modifyInstructionAuthorization.ts
```diff
@@ -0,0 +1,98 @@
...
+export interface ModifyInstructionAuthorizationParams {
+  operation: InstructionAuthorizationOperation;
+}
+
+/**
+ * @hidden
+ */
+export type Params = ModifyInstructionAuthorizationParams & {
+  id: BigNumber;
+};
```
monitz87, line 22: No need for two separate types here, they can be merged into a single interface

src/api/entities/Instruction/types.ts
```diff
@@ -47,3 +47,9 @@ export interface InstructionAuthorization {
   identity: Identity;
   authorizationStatus: AuthorizationStatus;
 }
+
+export enum InstructionAuthorizationOperation {
```
monitz87, line 51: This should be an internal type

src/api/procedures/utils.ts
```diff
@@ -29,3 +33,51 @@
 //     });
 //   }
 // }
+
+/**
+ * @hidden
+ */
+export async function checkInstructionValidation(
```
monitz87, line 40: assertInstructionValid

src/api/procedures/modifyInstructionAuthorization.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+
+      this.addTransaction(tx.settlement.authorizeInstruction, {}, rawInstructionId, [
+        rawPortfolioId,
+      ]);
+
+      return instruction;
+    }
+    case InstructionAuthorizationOperation.Unauthorize: {
+      if (authorizationStatus === AuthorizationStatus.Pending) {
```
monitz87, line 69: This error should also be thrown on `AuthorizationStatus.Rejected`

src/api/procedures/modifyInstructionAuthorization.ts
```diff
@@ -0,0 +1,98 @@
...
+        throw new PolymeshError({
+          code: ErrorCode.ValidationError,
+          message: 'The Instruction is unauthorize',
```
monitz87, line 72: The instruction is not authorized

src/api/procedures/modifyInstructionAuthorization.ts
```diff
@@ -0,0 +1,98 @@
...
+ * @hidden
+ */
+export type Params = ModifyInstructionAuthorizationParams & {
+  id: BigNumber;
+};
+
+/**
+ * @hidden
+ */
+export async function prepareModifyInstructionAuthorization(
```
monitz87, line 27: Why are you only returning the instruction in 2/3 cases?

monitz87 (changes requested):

src/api/procedures/modifyInstructionAuthorization.ts
shuffledex: I thought if operation is to reject the Instruction, return it seems some weird

monitz87: Makes sense, but rejecting is not deleting. The instruction can still be re-authorized in some cases and the pertinent data still exists

src/api/procedures/modifyInstructionAuthorization.ts
```diff
@@ -66,10 +61,13 @@ export async function prepareModifyInstructionAuthorization(
       return instruction;
     }
     case InstructionAuthorizationOperation.Unauthorize: {
-      if (authorizationStatus === AuthorizationStatus.Pending) {
+      if (
+        authorizationStatus === AuthorizationStatus.Pending ||
+        authorizationStatus === AuthorizationStatus.Rejected
```
monitz87, line 66:
```suggestion
        [AuthorizationStatus.Pending, AuthorizationStatus.Rejected].includes(authorizationStatus)
```

monitz87 (changes requested):
Return the instruction at the end of the procedure instead of on only 2 of 3 cases

monitz87 (approved):

PR #237, opened 2020-10-19
feat: add `getPendingInstructions` to `CurrentIdentity`

monitz87:
Also add `addInstruction` to `Venue`

### DIFF
--- a/src/api/entities/types.ts
+++ b/src/api/entities/types.ts
@@ -35,3 +35,4 @@ export * from './SecurityToken/types';
 export * from './Venue/types';
 export * from './Instruction/types';
 export * from './Portfolio/types';
+export * from './Proposal/types';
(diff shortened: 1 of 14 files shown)


src/api/entities/types.ts
```diff
@@ -35,3 +35,4 @@ export * from './SecurityToken/types';
 export * from './Venue/types';
 export * from './Instruction/types';
 export * from './Portfolio/types';
+export * from './Proposal/types';
```
VictorVicente, line 38: Everything related to Governance is commented. Should we comment it too?

VictorVicente (approved):

src/api/entities/types.ts
monitz87: Sure, I'll comment it and any imports that depend on it

PR #240, opened 2020-10-20
feat: middleware upgrade to work with chain breaking changes

VictorVicente:
Claim scopes have changed to an object in the form of {type; value}

BREAKING CHANGE: Scope filtering in `getIdentitiesWithClaims` and `getTargetingClaims` take an
`Scope` object instead of a string.


src/Claims.ts
```diff
@@ -124,7 +132,7 @@ export class Claims {
     const result = await context.queryMiddleware<Ensured<Query, 'didsWithClaims'>>(
       didsWithClaims({
         dids: targets?.map(target => signerToString(target)),
-        scope,
+        scope: scope ? { type: ClaimScopeTypeEnum[scope.type], value: scope.value } : scope,
```
monitz87, line 135:
```suggestion
        scope: scope ? { type: ClaimScopeTypeEnum[scope.type], value: scope.value } : undefined,
```

Just to be more explicit

src/Claims.ts
```diff
@@ -251,7 +259,7 @@ export class Claims {
     const result = await context.queryMiddleware<Ensured<Query, 'issuerDidsWithClaimsByTarget'>>(
       issuerDidsWithClaimsByTarget({
         target: did,
-        scope,
+        scope: scope ? { type: ClaimScopeTypeEnum[scope.type], value: scope.value } : scope,
```
monitz87, line 262:
```suggestion
        scope: scope ? { type: ClaimScopeTypeEnum[scope.type], value: scope.value } : undefined,
```

src/utils/__tests__/index.ts
```diff
@@ -1962,7 +1962,7 @@ describe('createClaim', () => {
   test('', () => {
```
monitz87, line 1962: Can you add a description for this test?

monitz87 (changes requested):

src/utils/__tests__/index.ts
VictorVicente: Done

monitz87 (approved):

PR #243, opened 2020-10-21
fix: fix how jurisdiction are converted from middleware

VictorVicente:


src/utils/index.ts
```diff
@@ -1117,7 +1117,7 @@ export function createClaim(
     return {
       type,
       // this assertion is necessary because CountryCode is not in the middleware types
-      code: jurisdiction as CountryCode,
+      code: stringUpperFirst(jurisdiction?.toLowerCase()) as CountryCode,
```
monitz87, line 1120:
```suggestion
      code: stringUpperFirst(jurisdiction!.toLowerCase()) as CountryCode,
```

To reflect that we don't ever expect `jurisdiction` to be falsy in this scenario

monitz87 (changes requested):

monitz87 (approved):

PR #249, opened 2020-10-25
chore(types): preserve historic transaction tags on chain upgrade

monitz87:
Also upgraded polkadot to latest version and upgraded types to latest mesh metadata


package.json
```diff
@@ -84,9 +84,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@polkadot/api": "1.29.1",
-    "@polkadot/util": "3.2.1",
-    "@polkadot/util-crypto": "3.2.1",
+    "@polkadot/api": "2.3.1",
+    "@polkadot/util": "3.6.1",
+    "@polkadot/util-crypto": "3.6.1",
```
VictorVicente, line 89: Is there any breaking change from polkadot on these versions?

src/Polymesh.ts
```diff
@@ -143,13 +144,20 @@ export class Polymesh {
 
     try {
       const { types, rpc } = polymesh;
+      console.log('HEY');
```
VictorVicente, line 147: This should be removed.

src/Polymesh.ts
```diff
@@ -143,13 +144,20 @@ export class Polymesh {
...
-        types,
+        // https://github.com/polkadot-js/api/releases/tag/v2.0.1 TODO @monitz87: remove once Polymesh is updated to substrate 2.0
+        types: {
+          ...types,
+          RefCount: 'RefCountTo259',
+        },
         rpc,
       });
 
+      console.log('HEY2');
```
VictorVicente, line 159: This one too.

VictorVicente (changes requested):

package.json
monitz87: Already addressed

VictorVicente (approved):

PR #254, opened 2020-11-04
feat: get the portfolio custodian

shuffledex:


src/api/entities/Portfolio/index.ts
```diff
@@ -130,4 +135,30 @@ export class Portfolio extends Entity<UniqueIdentifiers> {
 
     return values(assetBalances);
   }
+
+  /**
+   * Retrieve the custodian of this portfolio
```
monitz87, line 140:
```suggestion
   * Retrieve the custodian Identity of this Portfolio. Returns null if no custodian is set
```

src/api/entities/Portfolio/index.ts
```diff
@@ -130,4 +135,30 @@ export class Portfolio extends Entity<UniqueIdentifiers> {
...
+   */
+  public async getCustodian(): Promise<Identity> {
```
monitz87, line 142: Should return null if there is no custodian. We will implement `getAuthorizer` in the future, which will do what you were trying to do here

src/api/entities/Portfolio/__tests__/index.ts
```diff
@@ -166,4 +166,44 @@ describe('Portfolio class', () => {
...
+      let portfolio = new Portfolio({ did, id }, context);
+      let result = await portfolio.getCustodian();
+
+      expect(result.did).toEqual(custodianDid);
+
+      dsMockUtils.createQueryStub('portfolio', 'portfolioCustodian').returns({});
+
+      identityIdToStringStub.returns(did);
+
+      portfolio = new Portfolio({ did, id }, context);
```
monitz87, line 203: No need to redefine portfolio

monitz87 (changes requested):

src/api/entities/Portfolio/index.ts
```diff
@@ -130,4 +135,30 @@ export class Portfolio extends Entity<UniqueIdentifiers> {
...
+    } = this;
+
+    const rawPortfolioId = portfolioIdToMeshPortfolioId({ did, number: _id }, context);
+    const portfolioCustodian = await portfolio.portfolioCustodian(rawPortfolioId);
+
+    try {
+      const rawIdentityId = portfolioCustodian.unwrap();
+      return new Identity({ did: identityIdToString(rawIdentityId) }, context);
+    } catch (_) {
+      return new Identity({ did }, context);
```
monitz87, line 161: You can just return the owner here, no need to create a new instance

monitz87 (changes requested):

monitz87 (approved):

PR #268, opened 2020-11-19
Feat/msdk 312 refactor batch

monitz87:
Major changes:

- Base classes, entities and procedures are now being exported/imported from a single entry point (`~/internal`) to control load order
- Created a new `PolymeshTransactionBatch` class that works similar to the `PolymeshTransaction` class but uses `utility.batch` to call the same transaction several times with different arguments
- Added `addBatchTransaction` to the `Procedure` class. This function uses `batchArguments` behind the scenes to separate the arguments into multiple batches if necessary, and adds `PolymeshTransactionBatch` objects to the queue


src/types/index.ts
```diff
@@ -3,6 +3,8 @@ import { IKeyringPair, TypeDef } from '@polkadot/types/types';
 import BigNumber from 'bignumber.js';
 import { TxTag } from 'polymesh-types/types';
 
+// import { ProposalDetails } from '~/api/entities/Proposal/types';
+import { CountryCode } from '~/generated/types';
 // NOTE uncomment in Governance v2 upgrade
```
VictorVicente, line 8: Move the comment too

VictorVicente (approved):

PR #279, opened 2020-12-02
feat(trusted claim issuers): require trusted claim types on claimissuers

monitz87:
Methods that modify default trusted claim issuers for a token now require passing Claim Types for
which that issuer is trusted

BREAKING CHANGE: - `TrustedClaimIssuers.set` and `TrustedClaimIssuers.add` now require an array of
`TrustedClaimIssuer` objects (each with an identity and an array of ClaimTypes)

### DIFF
--- a/src/api/entities/DefaultTrustedClaimIssuer.ts
+++ b/src/api/entities/DefaultTrustedClaimIssuer.ts
@@ -13,7 +13,7 @@ export interface UniqueIdentifiers {
 }
 
 export interface Params {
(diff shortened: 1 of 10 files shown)


VictorVicente (changes requested):
I'd prefer to leave it as `null` by default. This way there are no breaking changes, and if you don't want to use specific claim types by trusted claim issuer, it's easier.

VictorVicente (approved):

PR #281, opened 2020-12-02
feat: revoke all permissions for secondary keys

shuffledex:


src/api/procedures/revokePermissions.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @hidden
+ */
+export async function isAuthorized(this: Procedure<RevokePermissionsParams>): Promise<boolean> {
+  const { context } = this;
+
+  const identity = await context.getCurrentIdentity();
+  const primaryKey = await identity.getPrimaryKey();
+
+  return primaryKey === context.getCurrentPair().address;
+}
```
monitz87, line 67: This is incorrect. You can set permissions even if you're not the primary key. Nothing should be checked here until we start checking permissions

monitz87 (approved):

src/api/procedures/__tests__/utils.ts
```diff
@@ -238,3 +241,59 @@ describe('assertInstructionValid', () => {
...
 });
+
+describe('assertSecondaryKeys', () => {
+  let signerToSignerValueStub: sinon.SinonStub<[Signer], SignerValue>;
+
+  beforeAll(() => {
+    signerToSignerValueStub = sinon.stub(utilsConversionModule, 'signerToSignerValue');
+  });
+
+  test('should not throw an error', async () => {
```
monitz87, line 252:
```suggestion
  test('should not throw an error if all signers are secondary keys', async () => {
```

monitz87 (changes requested):
Please remove `isAuthorized` from the `removeSecondaryKeys` procedure as well (not only the primary key can remove secondary keys)

monitz87 (approved):

PR #294, opened 2020-12-13
Feat/msdk 378 account permissions

monitz87:


src/api/procedures/consumeJoinIdentityAuthorization.ts
```diff
@@ -67,26 +69,33 @@ export async function prepareConsumeJoinIdentityAuthorization(
...
   const fetchDid = async (): Promise<string> => getDid(did, context);
 
+  let transactions: TxTag[] = [];
+
+  let paidByThirdParty = false;
+
   if (target instanceof Account) {
     const { address } = context.getCurrentAccount();
     condition = address === target.address;
+    paidByThirdParty = condition;
```
VictorVicente, line 90: I think `transactions = [TxTags.identity.JoinIdentityAsKey];` is missing here

src/api/procedures/togglePauseRequirements.ts
```diff
@@ -51,14 +52,24 @@ export async function prepareTogglePauseRequirements(
...
-export function getRequiredRoles({ ticker }: Params): Role[] {
-  return [{ type: RoleType.TokenOwner, ticker }];
+export function getAuthorization(
+  this: Procedure<Params, SecurityToken>,
+  { ticker }: Params
+): ProcedureAuthorization {
+  return {
+    identityRoles: [{ type: RoleType.TokenOwner, ticker }],
+    signerPermissions: {
+      transactions: [TxTags.complianceManager.PauseAssetCompliance],
```
VictorVicente, line 62: It could be also `ResumeAssetCompliance`

src/base/Procedure.ts
```diff
@@ -107,14 +112,31 @@ export class Procedure<Args extends unknown = void, ReturnValue extends unknown
...
+
+      let signerAllowed: boolean;
+      if (typeof signerPermissions !== 'boolean') {
+        const account = context.getCurrentAccount();
+        signerAllowed = await account.hasPermissions(signerPermissions);
+      } else {
+        signerAllowed = signerPermissions;
+      }
+
+      allowed = identityAllowed && signerAllowed;
```
VictorVicente, line 137: We should throw different messages for Identity and Account errors.

VictorVicente (changes requested):
`modifySignerPermissions.ts` has no `getAuthorization` method.
TxTag is wrong in `setTokenDocuments.ts` line 89. It should be `AddDocuments`

src/api/procedures/consumeJoinIdentityAuthorization.ts
monitz87: No, that would always fail since the account can't have any permissions if it's not affiliated to any Identity

VictorVicente (approved):

PR #299, opened 2020-12-15
feat: get cdd claims from storage

shuffledex:


src/base/Context.ts
```diff
@@ -607,58 +611,97 @@ export class Context {
...
+          claim_type: claimTypeToMeshClaimType(ClaimType.CustomerDueDiligence, this),
+        },
+        paginationOpts: {
+          size: size ?? DEFAULT_GQL_PAGE_SIZE,
+        },
+      });
+
+      const data: ClaimData[] = [];
+      entries.forEach(([key, identityClaim]) => {
+        const claim1stKey = key.args[0] as Claim1stKey;
```
monitz87, line 688: You should destructure this

src/base/Context.ts
```diff
@@ -607,58 +611,97 @@ export class Context {
...
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          claim_type: claimTypeToMeshClaimType(ClaimType.CustomerDueDiligence, this),
+        },
+        paginationOpts: {
+          size: size ?? DEFAULT_GQL_PAGE_SIZE,
+        },
+      });
+
+      const data: ClaimData[] = [];
+      entries.forEach(([key, identityClaim]) => {
```
monitz87, line 687: You should destructure identityClaim

monitz87 (changes requested):

src/base/Context.ts
```diff
@@ -606,59 +653,75 @@ export class Context {
...
-    });
+      const next = calculateNextKey(count, size, start);
 
-    const next = calculateNextKey(count, size, start);
+      return {
+        data,
+        next,
+        count,
+      };
+    } else {
```
monitz87, line 713: This else is not necessary, there is a return at the end of the previous block

src/base/Context.ts
```diff
@@ -606,59 +653,75 @@ export class Context {
...
+      }
 
-    return {
-      data,
-      next,
-      count,
-    };
+      const identityClaimsEntries = await this.identityClaimsEntries({ targets, claimTypes });
+
+      return identityClaimsEntries;
```
monitz87, line 723:
```suggestion
      return this.identityClaimsEntries({ targets, claimTypes });
```

src/base/__tests__/Context.ts
```diff
@@ -1174,13 +1182,13 @@ describe('Context class', () => {
...
 
-      let result = await context.issuedClaims({
+      let result = (await context.issuedClaims({
         targets: [targetDid],
         trustedClaimIssuers: [targetDid],
         claimTypes: [ClaimType.Accredited],
         includeExpired: true,
         size: 1,
-      });
+      })) as ResultSet<ClaimData>;
```
monitz87, line 1191: These assertions shouldn't be necessary once the function is returning a single type

src/base/__tests__/Context.ts
```diff
@@ -1200,12 +1208,108 @@ describe('Context class', () => {
         }
       );
 
-      result = await context.issuedClaims();
+      result = (await context.issuedClaims()) as ResultSet<ClaimData>;
```
monitz87, line 1211: These assertions shouldn't be necessary once the function is returning a single type

src/base/__tests__/Context.ts
```diff
@@ -1200,12 +1208,108 @@ describe('Context class', () => {
...
 
       expect(result.data).toEqual(fakeClaims);
       expect(result.count).toEqual(25);
       expect(result.next).toBeNull();
     });
+
+    test('should return a list of claims', async () => {
```
monitz87, line 1218: Merge these two tests

src/base/__tests__/Context.ts
```diff
@@ -1200,12 +1208,108 @@ describe('Context class', () => {
...
+
+      const result = (await context.issuedClaims({
+        targets: [targetDid],
+        claimTypes: [ClaimType.CustomerDueDiligence],
+      })) as ClaimData[];
+
+      expect(result).toEqual(fakeClaims);
+    });
+
+    test('should throw if the middleware query fails and targets or claimTypes are not seted', async () => {
```
monitz87, line 1296:
```suggestion
    test('should throw if the middleware is not available and targets or claimTypes are not set', async () => {
```

src/base/__tests__/Context.ts
```diff
@@ -1266,7 +1370,7 @@ describe('Context class', () => {
       });
     });
 
-    test('should return a list of claims', async () => {
+    test('should return a result set of claims', async () => {
```
monitz87, line 1373: Isn't this test exactly the same as the first one?

src/base/Context.ts
```diff
@@ -585,6 +590,48 @@ export class Context {
     return assembleResult(didRecords);
   }
 
+  /**
+   * @hidden
+   */
+  private async identityClaimsEntries(args: {
```
monitz87, line 596: getIdentityClaimsFromChain

src/base/Context.ts
```diff
@@ -585,6 +590,48 @@ export class Context {
...
+      polymeshApi: {
+        query: { identity },
+      },
+    } = this;
+
+    const { targets, claimTypes } = args;
+    const data: ClaimData[] = [];
+
+    await P.each(targets, async rawTarget => {
+      await P.each(claimTypes, async rawClaimType => {
```
monitz87, line 610: Just `claimType`

monitz87, line 610: Double each is too slow. There should be a way of doing this using P.map

monitz87 (changes requested):

src/base/Context.ts
```diff
@@ -585,6 +591,60 @@ export class Context {
...
+              target: new Identity({ did: identityIdToString(target) }, this),
+              issuer: new Identity({ did: identityIdToString(claimissuer) }, this),
+              issuedAt: momentToDate(issuanceDate),
+              expiry,
+              claim: meshClaimToClaim(claim),
+            });
+          }
+        }
+      );
+    });
```
monitz87, line 643: This approach may lead to inconsistent ordering. You should have the map function return the claim data if `(!includeExpired && (expiry === null || expiry > new Date())) || includeExpired` is true, and undefined if not, and then filter out the undefined values afterwards with lodash `compact`

src/base/Context.ts
```diff
@@ -609,55 +669,79 @@ export class Context {
...
+    }
+
+    if (!targets || !claimTypes) {
+      throw new PolymeshError({
+        code: ErrorCode.FatalError,
+        message: 'Cannot perform this action without an active middleware connection',
+      });
+    }
+
+    const identityClaimsFromChain = await this.getIdentityClaimsFromChain({
```
monitz87, line 735: You're not filtering the results by claim issuers

monitz87 (changes requested):

src/base/Context.ts
monitz87: The `getIdentityClaimsFromChain` function should do the filtering

src/base/Context.ts
```diff
@@ -618,31 +623,40 @@ export class Context {
...
-        }
+        )
       );
+
+      return compactData;
     });
 
-    return data;
+    return flattenDeep(claimData).filter(({ issuer }) =>
+      trustedClaimIssuers ? trustedClaimIssuers.includes(issuer.did) : true
```
monitz87, line 658: `trustedClaimIssuers` can also contain Identities

monitz87 (changes requested):

src/base/Context.ts
shuffledex: line 626:

```
    const trustedClaimIssuers = rawTrustedClaimIssuers?.map(trustedClaimIssuer =>
      signerToString(trustedClaimIssuer)
    );
```

monitz87: Missed that. Call the original variable `trustedClaimIssuers` and this one `claimIssuerDids`

monitz87 (changes requested):

src/base/Context.ts
```diff
@@ -629,32 +629,31 @@ export class Context {
...
+            });
+          } else {
+            data.push(undefined);
           }
-        )
+        }
       );
-
-      return compactData;
+      return compact(data);
```
monitz87: Since you're using `forEach`, you no longer need the `undefined` values nor `compact`

monitz87 (changes requested):

monitz87 (approved):

PR #318, opened 2021-01-13
Feat/msdk 407 ticker validation

shuffledex:


src/utils/__tests__/internal.ts
```diff
@@ -424,15 +425,25 @@ describe('calculateNextKey', () => {
...
+    expect(stringIsClean('Clean String')).toBe(true);
+  });
+});
+
+describe('isPrintableASCII', () => {
+  test('should return true if the string contains printable ASCII', () => {
+    expect(isPrintableASCII('TICKET')).toBe(true);
+  });
+
+  test("should return false if the string doesn't contain printable ASCII", () => {
```
monitz87, line 445:
```suggestion
  test("should return false if the string doesn't contain only printable ASCII characters", () => {
```

monitz87 (changes requested):
`isPrintableAscii` should replace `stringIsClean`, and it doesn't need to be called elsewhere (since it will be called in `stringToTicker`, which is called every time a ticker is being sent to chain)

src/api/procedures/reserveTicker.ts
```diff
@@ -47,6 +47,13 @@ export async function prepareReserveTicker(
   } = this;
   const { ticker, extendPeriod = false } = args;
 
+  if (!isPrintableAscii(ticker)) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'Only printable ASCII is alowed as ticker name',
+    });
+  }
```
monitz87, line 55: Both this check and the next one are unnecessary, since those validations are done in `stringToTicker`

Also, please make `stringToTicker` also validate that the string is not empty (change the error message to `Ticker length must be between 1 and ${MAX_TICKER_LENGTH} characters`, and validate that the string is all upper case (separate error)

src/utils/__tests__/internal.ts
```diff
@@ -424,15 +424,15 @@ describe('calculateNextKey', () => {
...
   });
+});
 
-  describe('stringIsClean', () => {
-    test('should return false if the string contains charcode 65533', () => {
-      expect(stringIsClean(String.fromCharCode(65533))).toBe(false);
-    });
+describe('isPrintableASCII', () => {
+  test('should return true if the string only contains printable ASCII characters', () => {
+    expect(isPrintableAscii('TICKET')).toBe(true);
```
monitz87, line 431:
```suggestion
    expect(isPrintableAscii('TICKER')).toBe(true);
```

src/utils/__tests__/internal.ts
```diff
@@ -424,15 +424,15 @@ describe('calculateNextKey', () => {
 
     expect(nextKey).toEqual(30);
   });
+});
 
-  describe('stringIsClean', () => {
-    test('should return false if the string contains charcode 65533', () => {
-      expect(stringIsClean(String.fromCharCode(65533))).toBe(false);
-    });
+describe('isPrintableASCII', () => {
```
monitz87, line 429:
```suggestion
describe('isPrintableAscii', () => {
```

monitz87 (changes requested):

src/utils/conversion.ts
```diff
@@ -201,6 +201,13 @@ export function bytesToString(bytes: Bytes): string {
  * @hidden
  */
 export function stringToTicker(ticker: string, context: Context): Ticker {
+  if (!ticker.length) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: `Ticker length must be between 1 and ${MAX_TICKER_LENGTH} character`,
+    });
+  }
```
monitz87, line 209: You should only do one check that engulfs this error and the one below

monitz87 (changes requested):

VictorVicente (approved):

PR #330, opened 2021-01-20
Feat/msdk 420 validate cddid

shuffledex:


src/api/procedures/modifyClaims.ts
```diff
@@ -169,6 +169,47 @@ export async function prepareModifyClaims(
...
+
+      const issuedCddClaims = await context.issuedClaims({
+        targets: claimTargets.map(({ target }) => target),
+        claimTypes: [ClaimType.CustomerDueDiligence],
+        includeExpired: false,
+      });
+
+      claimTargets.forEach(({ target, claim }) => {
+        const did = signerToString(target);
+        const filterCdds = issuedCddClaims.data.filter(
```
monitz87, line 187: issuedClaimsForTarget

src/api/procedures/modifyClaims.ts
```diff
@@ -169,6 +169,47 @@ export async function prepareModifyClaims(
...
     );
   } else {
+    if (operation === ClaimOperation.Add) {
+      const existingCddId: string[] = [];
+
+      const claimTargets = claims.filter(
+        ({ claim: { type } }) => type === ClaimType.CustomerDueDiligence
+      );
+
+      const issuedCddClaims = await context.issuedClaims({
```
monitz87, line 179: All of this should only be done only if `newCddClaims.length` is a positive number. Otherwise you're performing an unnecessary request

src/api/procedures/modifyClaims.ts
```diff
@@ -169,6 +169,47 @@ export async function prepareModifyClaims(
...
+
+      if (existingCddId.length) {
+        throw new PolymeshError({
+          code: ErrorCode.ValidationError,
+          message: 'This Identity already has CDD claims with a different ID',
+          data: {
+            existingCddId: uniq(existingCddId),
+          },
+        });
+      }
```
monitz87, line 210: You need to take into account the fact that there might be multiple identities that cause the error. You should be storing occurrences in a `{ target: Identity; currentCddId: string; newCddId: string }[]` variable called `invalidCddClaims` instead of just the CDD id without its corresponding Identity.

Also, the message should be `A target Identity cannot have CDD claims with different IDs`

monitz87 (changes requested):

monitz87 (changes requested):

monitz87 (changes requested):

monitz87 (approved):

PR #341, opened 2021-01-26
Feat/msdk 358 pagination support

shuffledex:


src/Polymesh.ts
```diff
@@ -387,11 +394,13 @@ export class Polymesh {
    *
    * @param args.owner - identity representation or Identity ID as stored in the blockchain
    *
-   * * @note reservations with unreadable characters in their tickers will be left out
+   * @note reservations with unreadable characters in their tickers will be left out
+   * @note supports pagination
    */
   public async getTickerReservations(args?: {
```
monitz87, line 400: Can't add pagination to this one because results are filtered on the client side

src/Polymesh.ts
```diff
@@ -547,8 +560,12 @@ export class Polymesh {
    * @param args.owner - identity representation or Identity ID as stored in the blockchain
    *
    * @note tokens with unreadable characters in their tickers will be left out
+   * @note supports pagination
    */
-  public async getSecurityTokens(args?: { owner: string | Identity }): Promise<SecurityToken[]> {
+  public async getSecurityTokens(args?: {
```
monitz87, line 565: Can't add pagination to this one because results are filtered on the client side

src/api/entities/Instruction/index.ts
```diff
@@ -169,22 +172,33 @@ export class Instruction extends Entity<UniqueIdentifiers> {
...
+    // const legs = await settlement.instructionLegs.entries(numberToU64(id, context));
+    const { entries, lastKey: next } = await requestPaginated(settlement.instructionLegs, {
+      arg: numberToU64(id, context),
+      paginationOpts,
+    });
+
+    const legs: Leg[] = [];
 
-    return legs.map(([, leg]) => {
+    entries.forEach(([, leg]) => {
```
monitz87, line 183: This can be done with map

monitz87 (changes requested):
You missed `portfolios.getCustodiedPortfolios` and `Instruction.getAffirmations` (please replace  the forEach in that one with a map)

PR #344, opened 2021-02-01
feat: update the ST identifiers

shuffledex:


src/api/procedures/modifyToken.ts
```diff
@@ -88,6 +108,22 @@ export async function prepareModifyToken(
     );
   }
 
+  if (newIdentifiers) {
+    if (isEqual(identifiers, newIdentifiers)) {
```
monitz87, line 112: This is not enough to compare arrays. If the arrays have the same elements but in a different order, this check would be false when it should be true.

To check if an array `a` has the same elements as an array `b`, you have to check `if (!differenceWith(a, b, isEqual).length && a.length === b.length)` (`differenceWith` and `isEqual` from `lodash`)

src/api/procedures/modifyToken.ts
```diff
@@ -88,6 +108,22 @@ export async function prepareModifyToken(
...
+      throw new PolymeshError({
+        code: ErrorCode.ValidationError,
+        message: 'New identifiers are the same as current token identifiers',
```
monitz87, line 115: ...as current identifiers

src/testUtils/mocks/entities.ts
```diff
@@ -1483,7 +1489,19 @@ export function getSecurityTokenIsFrozenStub(frozen?: boolean): SinonStub {
 
 /**
  * @hidden
- * Retrieve the stub of the `SecurityToken.transfers.canTransfer` method
+ * Retrieve the stub of the `SecurityToken.getIdentifiers` method
+ */
+export function getSecurityTokenGetIdentifiersStub(ientifiers?: TokenIdentifier): SinonStub {
```
monitz87, line 1494: typo in the parameter name

monitz87 (changes requested):

monitz87 (approved):

PR #348, opened 2021-02-03
feat: freeze and unfreeze sto

shuffledex:


src/api/procedures/toggleFreezeSto.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @hidden
+ */
+export type Params = ToggleFreezeStoParams & {
+  ticker: string;
+};
+
+/**
+ * @hidden
+ */
+export async function prepareToggleFreezeSto(
```
monitz87, line 23: Should throw an error if the STO has reached its end date

"The STO has already ended"

monitz87 (changes requested):

monitz87 (changes requested):

monitz87 (approved):

PR #352, opened 2021-02-05
feat: hardcode permittted TxTags in txTagToProtocolOp

shuffledex:


src/types/internal.ts
```diff
@@ -201,3 +201,23 @@ export interface ProcedureMethod<MethodArgs, ReturnValue> {
   (args: MethodArgs): Promise<TransactionQueue<ReturnValue>>;
   checkAuthorization: (args: MethodArgs) => Promise<ProcedureAuthorizationStatus>;
 }
+
+export enum HardcodedProtocolOp {
```
monitz87, line 205: This enum is not necessary. The values already exist as TxTags, no need to redefine them.

Create an `protocolOpTags` array inside the conversion function and just see if the passed TxTag is in the array

src/utils/conversion.ts
```diff
@@ -1830,6 +1831,13 @@ export function txTagToProtocolOp(tag: TxTag, context: Context): ProtocolOp {
     extrinsicName.replace(new RegExp('Documents$'), 'Document') // `asset.addDocuments` and `asset.removeDocuments`
   )}`;
 
+  if (!(value in HardcodedProtocolOp)) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: `${value} does not match any ProtocolOp value`,
```
monitz87, line 1837:
```suggestion
      message: `${value} does not match any ProtocolOp`,
```

monitz87 (changes requested):

src/utils/conversion.ts
```diff
@@ -1831,10 +1830,30 @@ export function txTagToProtocolOp(tag: TxTag, context: Context): ProtocolOp {
...
+    'IdentityCddRegisterDid',
+    'IdentityAddClaim',
+    'IdentitySetPrimaryKey',
+    'IdentityAddSecondaryKeysWithAuthorization',
+    'PipsPropose',
+    'VotingAddBallot',
+    'ContractsPutCode',
+    'BallotAttachBallot',
+    'DistributionDistribute',
+  ];
```
monitz87, line 1851: Use TxTags here and compare against `tag`

monitz87 (changes requested):

monitz87 (approved):

PR #354, opened 2021-02-08
feat: validate if number is integer and positive

shuffledex:


src/utils/__tests__/internal.ts
```diff
@@ -475,3 +478,41 @@ describe('createProcedureMethod', () => {
...
+
+  afterEach(() => {
+    dsMockUtils.reset();
+  });
+
+  afterAll(() => {
+    dsMockUtils.cleanup();
+  });
+
+  test('assertIsInteger should throw an error if the count is not an integer', async () => {
```
monitz87, line 515: Wrong description

monitz87: Also you should test that the functions don't throw when the condition is met

src/utils/__tests__/internal.ts
```diff
@@ -475,3 +478,41 @@ describe('createProcedureMethod', () => {
...
+    expect(() => assertIsInteger(('noInteger' as unknown) as BigNumber)).toThrow(
+      'The number must be an integer'
+    );
+  });
```
monitz87, line 499: You should be testing that the function throws when a float is passed. As we discussed, it's the compiler's job to validate that inputs are of the correct type

monitz87 (changes requested):

src/utils/__tests__/internal.ts
```diff
@@ -492,10 +492,20 @@ describe('assertIsInteger', () => {
     dsMockUtils.cleanup();
   });
 
+  test('should not throw', async () => {
```
monitz87, line 495:
```suggestion
  test('should not throw if the argument is an integer', async () => {
```

src/utils/__tests__/internal.ts
```diff
@@ -512,7 +522,15 @@ describe('assertIsPositive', () => {
     dsMockUtils.cleanup();
   });
 
-  test('assertIsInteger should throw an error if the count is not an integer', async () => {
+  test('should not throw', async () => {
```
monitz87, line 525:
```suggestion
  test('should not throw if the argument is positive', async () => {
```

src/utils/__tests__/internal.ts
```diff
@@ -512,7 +522,15 @@ describe('assertIsPositive', () => {
...
+    try {
+      assertIsPositive(new BigNumber(1));
+    } catch (_) {
+      expect(true).toBe(false);
+    }
+  });
+
+  test('assertIsPositive should throw an error if the count is not an integer', async () => {
```
monitz87, line 533:
```suggestion
  test('assertIsPositive should throw an error if the argument is negative', async () => {
```

src/utils/__tests__/internal.ts
```diff
@@ -492,10 +492,20 @@ describe('assertIsInteger', () => {
...
 
+  test('should not throw', async () => {
+    try {
+      assertIsInteger(new BigNumber(1));
+    } catch (_) {
+      expect(true).toBe(false);
+    }
+  });
+
   test('assertIsInteger should throw an error if the count is not an integer', async () => {
```
monitz87, line 503:
```suggestion
  test('assertIsInteger should throw an error if the argument is not an integer', async () => {
```

monitz87 (changes requested):

monitz87 (approved):

PR #363, opened 2021-02-18
refactor: remove code smells

monitz87:


src/api/procedures/modifyToken.ts
```diff
@@ -74,11 +74,6 @@ export async function prepareModifyToken(
     }
 
     this.addTransaction(tx.asset.makeDivisible, {}, rawTicker);
-  } else if (makeDivisible === false) {
-    throw new PolymeshError({
-      code: ErrorCode.ValidationError,
-      message: 'You cannot make the token indivisible',
-    });
```
VictorVicente, line 81: Why is this not required any more?

src/base/Context.ts
```diff
@@ -115,68 +116,46 @@ export class Context {
...
     polymeshApi: ApiPromise;
     middlewareApi: ApolloClient<NormalizedCacheObject> | null;
-    seed?: string;
-    keyring?: CommonKeyring;
-    uri?: string;
-    mnemonic?: string;
+    accountSeed?: string;
+    keyring?: CommonKeyring | UiKeyring;
+    accountUri?: string;
+    accountMnemonic?: string;
```
VictorVicente, line 128: Are these marked as breaking changes?

src/api/procedures/modifyToken.ts
monitz87: Because it was never possible. `makeDivisible` is typed as `true | undefined`

src/base/Context.ts
monitz87: They're not really breaking changes because the `Context` class isn't accessible externally

VictorVicente: Oh great, I thought it was the `connect` method

VictorVicente (approved):

PR #369, opened 2021-02-24
feat(checkpoints): add `Checkpoints.get` and create Checkpoint entity

monitz87:
Added `totalSupply` and `timestamp` to Checkpoint


src/api/entities/SecurityToken/Checkpoints.ts
```diff
@@ -24,4 +26,38 @@ export class Checkpoints extends Namespace<SecurityToken> {
...
+      { paginationOpts, arg: rawTicker }
+    );
+
+    const now = new Date();
+    const data = entries
+      .map(([{ args: [, id] }, timestamp]) => ({
+        checkpoint: new Checkpoint({ id: u64ToBigNumber(id), ticker }, context),
+        createdAt: momentToDate(timestamp),
+      }))
+      .filter(({ createdAt }) => createdAt <= now);
```
VictorVicente, line 56: Why do you need these filter?

monitz87: The query also returns the next scheduled checkpoint's time

VictorVicente (approved):

PR #370, opened 2021-02-25
feat(sto): improve offering details

monitz87:
Added total offering amount, total remaining tokens and refined status

BREAKING CHANGE: Sto status is now divided into `timing`, `sale` and `balance`

### DIFF
--- a/src/api/entities/SecurityToken/Offerings.ts
+++ b/src/api/entities/SecurityToken/Offerings.ts
@@ -46,9 +46,9 @@ export class Offerings extends Namespace<SecurityToken> {
   /**
    * Retrieve all of the Token's Offerings. Can be filtered using parameters
    *
-   * @param opts.status - status of the offerings to fetch
+   * @param opts.status - status of the offerings to fetch. As long as the STO has one of the passed statuses, it will be returned
    */
-  public async get(opts: { status?: StoStatus } = {}): Promise<StoWithDetails[]> {
+  public async get(opts: { status?: Partial<StoStatus> } = {}): Promise<StoWithDetails[]> {
     const {
       parent: { ticker },
       context: {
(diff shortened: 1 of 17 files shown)


src/api/entities/SecurityToken/Offerings.ts
```diff
@@ -46,9 +46,9 @@ export class Offerings extends Namespace<SecurityToken> {
   /**
    * Retrieve all of the Token's Offerings. Can be filtered using parameters
    *
-   * @param opts.status - status of the offerings to fetch
+   * @param opts.status - status of the offerings to fetch. As long as the STO has one of the passed statuses, it will be returned
```
VictorVicente, line 49: This comment is not accurate. STO must have all the passed statuses to be returned.

src/api/procedures/investInSto.ts
```diff
@@ -110,25 +118,21 @@ export async function prepareInvestInSto(
...
-    throw new PolymeshError({
-      code: ErrorCode.ValidationError,
-      message: 'The STO is not live',
-    });
-  }
-
-  const now = new Date();
-
-  if (end && now > end) {
+  if (sale !== StoSaleStatus.Live || timing !== StoTimingStatus.Started) {
```
VictorVicente, line 132: Should `balanceStatus` be `Available` too?

src/api/entities/SecurityToken/Offerings.ts
monitz87: I had this in place before we spoke, I forgot to change it 😅

src/api/procedures/investInSto.ts
monitz87: Agreed

monitz87: Wait, that is already considered further down. The reason it's calculated apart and not derived from the status is the `maxPrice` parameter

VictorVicente (approved):

PR #378, opened 2021-03-04
feat: set ss58_format when instantiating the sdk

shuffledex:


src/api/entities/__tests__/Account.ts
```diff
@@ -23,7 +23,7 @@ describe('Account class', () => {
     address = 'someAddress';
     key = 'someKey';
 
-    sinon.stub(utilsConversionModule, 'addressToKey').withArgs(address).returns(key);
+    sinon.stub(utilsConversionModule, 'addressToKey').returns(key);
```
VictorVicente, line 26: Shouldn't it be `.withArgs(account, context)`?

shuffledex: no need to check attrs in this one, the return value will be always the same

monitz87 (changes requested):

monitz87 (approved):

PR #388, opened 2021-03-15
feat: schedule exists

shuffledex:


src/api/entities/__tests__/CheckpointSchedule.ts
```diff
@@ -122,4 +125,24 @@ describe('CheckpointSchedule class', () => {
       expect(result).toBeNull();
     });
   });
+
+  describe('method: exists', () => {
+    test('should return whether the schedule exists', async () => {
```
monitz87, line 130: You should include a false result in this test

monitz87 (changes requested):
You should check if the schedule exists in all relevant methods:

- The removeSchedu

monitz87 (approved):

PR #392, opened 2021-03-16
fix: enforce chain ss58 format to keyring

VictorVicente:


VictorVicente:
We won't be enforcing the format in the keyring. Instead SDK should throw an error if keypairs are not set correctly.

PR #394, opened 2021-03-16
Feat/dividend entities

monitz87:


src/api/entities/DividendDistribution/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface Placeholder {
+  foo: number;
+}
```
VictorVicente, line 3: Do we need this placeholder?

VictorVicente (approved):

PR #395, opened 2021-03-17
feat: checkpointSchedule exists

shuffledex:


src/api/entities/CheckpointSchedule/index.ts
```diff
@@ -23,6 +23,8 @@ export interface Params {
   nextCheckpointDate: Date;
 }
 
+const notExistsMessage = 'Schedule no longer exists. This means it was already finished';
```
monitz87, line 26:
```suggestion
const notExistsMessage = 'Schedule no longer exists. It was either removed or it expired';
```

src/api/entities/CheckpointSchedule/index.ts
```diff
@@ -112,6 +114,17 @@ export class CheckpointSchedule extends Entity<UniqueIdentifiers> {
 
     const rawSchedules = await checkpoint.schedules(stringToTicker(ticker, context));
 
+    const scheduleIds = rawSchedules.map(({ id: scheduleId }) =>
```
monitz87, line 117: You can just do `find` here and save yourself having to go through the entire array converting everything

monitz87: Same for the `exists` method itself

monitz87 (changes requested):
You forgot to add the validation to the `removeCheckpointSchedule` procedure

src/api/procedures/removeCheckpointSchedule.ts
```diff
@@ -31,12 +31,20 @@ export async function prepareRemoveCheckpointSchedule(
...
 
+  const id = schedule instanceof BigNumber ? schedule : schedule.id;
   const rawTicker = stringToTicker(ticker, context);
 
-  const rawScheduleId = numberToU64(
-    schedule instanceof BigNumber ? schedule : schedule.id,
-    context
-  );
+  const rawSchedules = await query.checkpoint.schedules(rawTicker);
+  const exists = rawSchedules.find(({ id: scheduleId }) => u64ToBigNumber(scheduleId).eq(id));
```
monitz87, line 38: Use the `exists` method

monitz87 (changes requested):

src/api/procedures/removeCheckpointSchedule.ts
monitz87: Never mind, this was already discussed

monitz87 (approved):

PR #397, opened 2021-03-19
feat: check ss58 format

shuffledex:


src/utils/internal.ts
```diff
@@ -447,3 +449,20 @@ function isUiKeyring(keyring: any): keyring is UiKeyring {
 export function getCommonKeyring(keyring: CommonKeyring | UiKeyring): CommonKeyring {
   return isUiKeyring(keyring) ? keyring.keyring : keyring;
 }
+
+/**
+ * @hidden
+ */
+export function isPrefixValid(address: string, ss58Format: number): void {
+  const encodedAddress = new Keyring({ type: 'sr25519', ss58Format }).encodeAddress(
```
monitz87, line 457: You should use `encodeAdress` from `@polkadot/util-crypto`

src/utils/internal.ts
```diff
@@ -447,3 +449,20 @@ function isUiKeyring(keyring: any): keyring is UiKeyring {
 export function getCommonKeyring(keyring: CommonKeyring | UiKeyring): CommonKeyring {
   return isUiKeyring(keyring) ? keyring.keyring : keyring;
 }
+
+/**
+ * @hidden
+ */
+export function isPrefixValid(address: string, ss58Format: number): void {
```
monitz87, line 456: `assertFormatValid`

src/utils/internal.ts
```diff
@@ -447,3 +449,20 @@ function isUiKeyring(keyring: any): keyring is UiKeyring {
...
+  const encodedAddress = new Keyring({ type: 'sr25519', ss58Format }).encodeAddress(
+    decodeAddress(address),
+    ss58Format
+  );
+
+  if (address !== encodedAddress) {
+    throw new PolymeshError({
+      code: ErrorCode.FatalError,
+      message: 'The supplied address is not encoding with the current SS58 prefix',
```
monitz87, line 465:
```suggestion
      message: 'The supplied address is not encoded with the chain's SS58 format',
```

monitz87: Also add the format as data

src/base/Context.ts
```diff
@@ -226,6 +228,8 @@ export class Context {
 
     try {
       newCurrentPair = keyring.getPair(address);
+
+      isPrefixValid(address, this.ss58Format);
```
monitz87, line 232: Better put this before the assignment (I know it's mostly the same functionally, it's just a matter of semantics, the idea being that you don't assign the value if it's an incorrect one)

src/Polymesh.ts
```diff
@@ -437,6 +437,10 @@ export class Polymesh {
     const { context } = this;
 
     if (args) {
+      const { ss58Format } = context;
+      const { address } = args;
+
+      isPrefixValid(address, ss58Format);
```
monitz87, line 443: This shouldn't be here, it should be in the Account constructor

monitz87 (changes requested):

src/api/entities/__tests__/Account.ts
```diff
@@ -8,6 +8,7 @@ import { dsMockUtils, entityMockUtils } from '~/testUtils/mocks';
 import { Mocked } from '~/testUtils/types';
```
monitz87, line 8: You should test that an error is thrown when instancing if the format is not valid

src/api/procedures/__tests__/consumeAuthorizationRequests.ts
```diff
@@ -40,6 +41,7 @@ describe('consumeAuthorizationRequests procedure', () => {
     numberToU64Stub = sinon.stub(utilsConversionModule, 'numberToU64');
     booleanToBoolStub = sinon.stub(utilsConversionModule, 'booleanToBool');
     sinon.stub(utilsConversionModule, 'addressToKey');
+    sinon.stub(utilsInternalModule, 'assertFormatValid');
```
monitz87, line 44: Instead of doing this, you should change all instantiations of Account in these tests to `entityMockUtils.getAccountInstance`

src/api/procedures/__tests__/consumeJoinIdentityAuthorization.ts
```diff
@@ -40,6 +41,7 @@ describe('consumeJoinIdentityAuthorization procedure', () => {
     rawFalse = dsMockUtils.createMockBool(false);
 
     sinon.stub(utilsConversionModule, 'addressToKey');
+    sinon.stub(utilsInternalModule, 'assertFormatValid');
```
monitz87, line 44: Instead of doing this, you should change all instantiations of Account in these tests to `entityMockUtils.getAccountInstance`

src/utils/__tests__/internal.ts
```diff
@@ -517,18 +517,18 @@ describe('getCommonKeyring', () => {
...
+    ).toThrow("The supplied address is not encoded with the chain's SS58 format");
   });
 
   test('should not throw if the address is prefixed with valid ss58', async () => {
     try {
-      isPrefixValid('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY', ss58Format);
+      assertFormatValid('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY', ss58Format);
     } catch (_) {
       expect(true).toBe(false);
     }
```
monitz87: This is not the ideal way of testing this

```
expect(() => assertFormatValid('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY', ss58Format)).not.toThrow()
```

monitz87 (changes requested):

PR #401, opened 2021-03-23
feat(claims): add support to addInvestorUniquenessClaimsV2

VictorVicente:


src/api/procedures/addInvestorUniquenessClaim.ts
```diff
@@ -40,14 +42,25 @@ export async function prepareAddInvestorUniquenessClaim(
...
+  } else {
+    this.addTransaction(
+      tx.identity.addInvestorUniquenessClaimV2,
+      {},
+      stringToIdentityId(did, context),
+      claimToMeshClaim({ type: ClaimType.InvestorUniqueness, scope, cddId, scopeId }, context),
+      scopeClaimProofToMeshScopeClaimProof(proof, scopeId, context),
+      expiry ? dateToMoment(expiry, context) : null
+    );
+  }
```
monitz87, line 63: Refactor the repeated conversions into variables

src/api/procedures/addInvestorUniquenessClaim.ts
```diff
@@ -57,6 +70,9 @@ export const addInvestorUniquenessClaim = new Procedure(prepareAddInvestorUnique
   signerPermissions: {
     tokens: [],
     portfolios: [],
-    transactions: [TxTags.identity.AddInvestorUniquenessClaim],
+    transactions: [
+      TxTags.identity.AddInvestorUniquenessClaim,
+      TxTags.identity.AddInvestorUniquenessClaimV2,
+    ],
```
monitz87: This should be refactored into a `getAuthorization` function so that you can use the type of `proof` to determine which transaction you need permissions for

src/utils/conversion.ts
```diff
@@ -2643,3 +2647,62 @@ export function storedScheduleToScheduleParams(storedSchedule: StoredSchedule):
...
+    proofScopeIdWellformed,
+    proofScopeIdCddIdMatch: {
+      firstChallengeResponse,
+      secondChallengeResponse,
+      subtractExpressionsRes,
+      blindedScopeDidHash,
+    },
+  } = proof;
+
+  const zkProofData = context.polymeshApi.createType('ZkProofData', {
```
monitz87, line 2690: Destructure `polymeshApi` so as not to repeat `context.polymeshApi`

monitz87 (changes requested):

src/testUtils/mocks/dataSources.ts
```diff
@@ -2501,3 +2506,52 @@ export const createMockScheduleSpec = (scheduleSpec?: {
     !scheduleSpec
   ) as ScheduleSpec;
 };
+
+/**
```
monitz87, line 2510: Missing comments in the rest of the functions

src/testUtils/mocks/dataSources.ts
```diff
@@ -2501,3 +2506,52 @@ export const createMockScheduleSpec = (scheduleSpec?: {
...
+    {
+      challenge_responses: challengeResponses.map(cr => createMockScalar(cr)),
+      subtract_expressions_res: createMockRistrettoPoint(subtractExpressionsRes),
+      blinded_scope_did_hash: createMockRistrettoPoint(blindedScopeDidHash),
+    },
+    false
+  ) as ZkProofData;
+};
+
+export const createMockScopeClaimProof = (
```
monitz87, line 2538: See how the other struct-creating functions are implemented (`createMockDocument` for example). You need to allow passing nothing to allow simulating empty values

monitz87 (changes requested):

src/testUtils/mocks/dataSources.ts
VictorVicente: I saw them but in this case, is there any use case for empty values? I couldn't think any of them

monitz87: At least the interface should be similar, receiving an object instead of a list of values

monitz87 (approved):

PR #404, opened 2021-03-27
feat(dividends): implement dividend creation

monitz87:


src/api/procedures/configureDividendDistribution.ts
```diff
@@ -0,0 +1,239 @@
...
+  } = this;
+
+  return {
+    identityRoles: [
+      { type: RoleType.TokenCaa, ticker },
+      { type: RoleType.PortfolioCustodian, portfolioId: portfolioToPortfolioId(portfolio) },
+    ],
+    signerPermissions: {
+      transactions: [TxTags.capitalDistribution.Distribute],
+      tokens: [],
```
VictorVicente, line 210: I think you need Token permission. Please check.

src/api/procedures/__tests__/initiateCorporateAction.ts
```diff
@@ -0,0 +1,337 @@
...
+        defaultTaxWithholding,
+        taxWithholdings,
+      });
+    } catch (error) {
+      err = error;
+    }
+
+    expect(err.message).toBe('Too many target Identities');
+    expect(err.data).toEqual({
+      maxTargets: 1000,
```
VictorVicente, line 187: This will fail when you start fetching this value from the chain

VictorVicente (commented):
Excellent job!
I really like the `isCodec` typeguard.

VictorVicente (changes requested):
There are some code smells to fix too.

src/api/procedures/__tests__/initiateCorporateAction.ts
monitz87: We'll have to change the test anyway to mock the value being returned

src/api/procedures/configureDividendDistribution.ts
monitz87: For this you do, but not for creating the CA

VictorVicente (approved):

PR #407, opened 2021-03-30
feat: link documents to corporate action

shuffledex:


src/utils/internal.ts
```diff
@@ -461,6 +462,19 @@ export function getCommonKeyring(keyring: CommonKeyring | UiKeyring): CommonKeyr
   return isUiKeyring(keyring) ? keyring.keyring : keyring;
 }
 
+/**
+ * @hidden
+ */
+export function documentComparator(a: TokenDocument, b: TokenDocument): boolean {
```
monitz87, line 468: I just realized this is not necessary, it can be replaced by lodash `isEquals`

src/types/index.ts
```diff
@@ -172,14 +173,27 @@ export function isPortfolioCustodianRole(role: Role): role is PortfolioCustodian
...
+  type: RoleType.CorporateActionsAgent;
+  ticker: string;
+}
+
+/**
+ * @hidden
+ */
+export function isCorporateActionsAgent(role: Role): role is CorporateActionsAgentRole {
+  return role.type === RoleType.CorporateActionsAgent;
+}
```
monitz87, line 186: Unnecessary, already exists

src/api/procedures/linkCaDocs.ts
```diff
@@ -0,0 +1,87 @@
...
+  rawAssetDocuments.forEach(([key, doc]) => {
+    const [, id] = key.args;
+    currentAssetDocIds.push(id);
+    currentAssetDocs.push(documentToTokenDocument(doc));
+  });
+
+  if (differenceWith(documents, currentAssetDocs, documentComparator).length) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'Some of the provided documents are not from the same asset',
```
monitz87, line 57: Some of the provided documents are not associated with the Security Token

src/api/procedures/linkCaDocs.ts
```diff
@@ -0,0 +1,87 @@
...
+  const rawAssetDocuments = await assetDocuments.entries(stringToTicker(ticker, context));
+
+  const currentAssetDocIds: DocumentId[] = [];
+  const currentAssetDocs: TokenDocument[] = [];
+
+  rawAssetDocuments.forEach(([key, doc]) => {
+    const [, id] = key.args;
+    currentAssetDocIds.push(id);
+    currentAssetDocs.push(documentToTokenDocument(doc));
+  });
```
monitz87, line 52: This approach results in every document associated to the asset being linked to the CA because you're not filtering anything. What you need to do is:

- rename `currentAssetDocIds` to `docIdsToLink`
- use lodash `remove` (together with `isEqual`) to search if the doc (converted to `TokenDocument`) is present in the input array (this will also remove the element from the input array so that subsequent searches are more performant). Be mindful that the result of `remove` is an array, so check for length
- if found, push the id to the `docIdsToLink` array, otherwise do nothing
- if there are any elements left in `documents` after the loop, it means the array had documents not in the asset, so you can throw the error (also, add `documents` to the error data and test for it)

src/api/procedures/__tests__/linkCaDocs.ts
```diff
@@ -0,0 +1,157 @@
...
+    dsMockUtils.reset();
+  });
+
+  afterAll(() => {
+    entityMockUtils.cleanup();
+    procedureMockUtils.cleanup();
+    dsMockUtils.cleanup();
+  });
+
+  test('should throw an error if some of the provided documents are not from the same asset', () => {
```
monitz87, line 110:
```suggestion
  test('should throw an error if some of the provided documents are not associated to the Security Token', () => {
```

monitz87 (changes requested):

monitz87 (approved):

PR #409, opened 2021-04-01
Feat/msdk 553 dividend details

monitz87:


src/api/entities/CorporateAction/index.ts
```diff
@@ -99,4 +107,85 @@ export class CorporateAction extends Entity<UniqueIdentifiers> {
...
+    } = this;
+
+    const rawTicker = stringToTicker(ticker, context);
+
+    const corporateAction = await query.corporateAction.corporateActions(
+      rawTicker,
+      numberToU32(id, context)
+    );
+
+    if (corporateAction.isNone) {
```
VictorVicente, line 147: Should this be calling `exists` instead?

src/api/entities/DividendDistribution/index.ts
```diff
@@ -76,4 +89,53 @@ export class DividendDistribution extends CorporateAction {
...
+   * Retrieve details associated with this Dividend Distribution
+   */
+  public async details(): Promise<DividendDistributionDetails> {
+    const { ticker, id, context } = this;
+
+    const distribution = await context.polymeshApi.query.capitalDistribution.distributions(
+      corporateActionIdentifierToCaId({ ticker, localId: id }, context)
+    );
+
+    if (distribution.isNone) {
```
VictorVicente, line 127: Same here

VictorVicente (changes requested):
There are a couple of comments and code smells.

src/api/entities/CorporateAction/index.ts
monitz87: Exists calls `corporateAction.corporateActions`, so we would be calling the same endpoint twice. I can refactor that call to avoid a bit of repetition

src/api/entities/DividendDistribution/index.ts
monitz87: Same answer as above

VictorVicente (approved):

PR #412, opened 2021-04-06
Feat/msdk 341 add batch instruction

shuffledex:


src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
+      from: PortfolioId;
+      to: PortfolioId;
+      asset: Ticker;
+      amount: Balance;
+    }[]
+  ][] = [];
+
+  const noLegs: number[] = [];
+  const noFutureBlock: number[] = [];
+  const noFutureDate: number[] = [];
```
monitz87, line 149: legErrIndexes
endBlockErrIndexes
datesErrIndexes

Also, they should count from 0 up, no need to mess up with the indexes

src/api/procedures/addInstruction.ts
```diff
@@ -228,40 +318,43 @@ export async function getAuthorization(
...
 
-    if (result.length) {
-      return result;
-    }
+      if (result.length) {
+        return result;
+      }
 
-    return undefined;
+      return [];
```
monitz87, line 351: Now that you always return an array you can just replace all this with `return result;`

src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
-        to: rawToPortfolio,
-        asset: stringToTicker(typeof token === 'string' ? token : token.ticker, context),
-        amount: numberToBalance(amount, context),
-      });
-    })
-  );
-
-  let newInstruction;
+  let affirmedInstructions: PostTransactionValue<Instruction[]> | undefined;
+  let addedInstructions: PostTransactionValue<Instruction[]>;
```
monitz87, line 256: You could just do `let instructions: PostTransactionValue<Instruction[]>;` and use it to store the result of both `adddBatchTransaction` calls

src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
+  const portfolios: (DefaultPortfolio | NumberedPortfolio)[] = [];
+  const transactions: SettlementTx[] = [];
+
+  portfoliosToAffirm.forEach(portfoliosList => {
+    transactions.push(
+      portfoliosList.length
+        ? TxTags.settlement.AddAndAffirmInstruction
+        : TxTags.settlement.AddInstruction
+    );
+    portfolios.push(...portfoliosList);
```
monitz87, line 304: You can use lodash `union` for transactions and `unionWith` together with `isEqual` for portfolios so that you don't have to use `uniq` later on a big array

src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
+      asset: Ticker;
+      amount: Balance;
+    }[]
+  ][] = [];
+
+  const noLegs: number[] = [];
+  const noFutureBlock: number[] = [];
+  const noFutureDate: number[] = [];
+
+  await P.each(instructions, async ({ legs, endBlock, tradeDate, valueDate }, i) => {
```
monitz87, line 151: COMPLEXITY: Let's try first refactoring this whole loop into a `getTxArgsAndErrors` function defined outside the prepare method that returns the necessary arrays as properties of an object

src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
+
+    if (tradeDate && valueDate && tradeDate > valueDate) {
+      noFutureDate.push(i + 1);
+    }
+
+    if (!noLegs.length && !noFutureBlock.length && !noFutureDate.length) {
+      const rawVenueId = numberToU64(venueId, context);
+      const rawSettlementType = endConditionToSettlementType(endCondition, context);
+      const rawTradeDate = tradeDate ? dateToMoment(tradeDate, context) : null;
+      const rawValueDate = valueDate ? dateToMoment(valueDate, context) : null;
```
monitz87, line 178: COMPLEXITY: if you default both tradeDate and valueDate to null, you can transform these into

`const rawTradeDate = tradeDate && dateToMoment(tradeDate, context)`

src/api/procedures/addInstruction.ts
```diff
@@ -98,128 +118,198 @@ export async function prepareAddInstruction(
...
+            assertPortfolioExists(toId, context),
+          ]);
+
+          const rawFromPortfolio = portfolioIdToMeshPortfolioId(fromId, context);
+          const rawToPortfolio = portfolioIdToMeshPortfolioId(toId, context);
+
+          rawLegs.push({
+            from: rawFromPortfolio,
+            to: rawToPortfolio,
+            asset: stringToTicker(typeof token === 'string' ? token : token.ticker, context),
```
monitz87, line 202: COMPLEXITY: You should refactor this into a `getTicker` function in internal utils (also please replace other occurrences of `typeof token === 'string' ? token : token.ticker` in the code with the function)

src/utils/internal.ts
```diff
@@ -196,28 +196,31 @@ type EventData<Event> = Event extends AugmentedEvent<'promise', infer Data> ? Da
 // TODO @monitz87: use event enum instead of string when it exists
 /**
  * @hidden
- * Find a specific event inside a receipt
+ * Find an specific event inside a receipt
```
monitz87, line 199:
```suggestion
 * Find every occurrence of a specific event inside a receipt
```

monitz87 (changes requested):

src/api/procedures/addInstruction.ts
```diff
@@ -93,32 +92,24 @@ export const createAddInstructionResolver = (
...
+): Promise<{
+  errIndexes: {
+    legErrIndexes: number[];
+    endBlockErrIndexes: number[];
+    datesErrIndexes: number[];
+  };
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  addAndAffirmInstructionParams: any;
+  addInstructionParams: any;
+  /* eslint-enabled @typescript-eslint/no-non-null-assertion */
```
monitz87, line 110: What is this for?

src/api/procedures/addInstruction.ts
```diff
@@ -93,32 +92,24 @@ export const createAddInstructionResolver = (
...
+  latestBlock: BigNumber,
+  venueId: BigNumber,
+  context: Context
+): Promise<{
+  errIndexes: {
+    legErrIndexes: number[];
+    endBlockErrIndexes: number[];
+    datesErrIndexes: number[];
+  };
+  /* eslint-disable @typescript-eslint/no-explicit-any */
```
monitz87, line 107: Why not use the proper type instead of any?

src/api/procedures/addInstruction.ts
```diff
@@ -301,15 +339,15 @@ export async function getAuthorization(
...
   });
 
   return {
     identityRoles: [{ type: RoleType.VenueOwner, venueId }],
     signerPermissions: {
       tokens: [],
-      portfolios: uniq(portfolios),
-      transactions: uniq(transactions),
+      portfolios,
+      transactions: union(transactions),
```
monitz87, line 350: The idea was to use union instead of pushing the transaction tags into the array. Using it here doesn't make sense.

```
transactions = union(transactions, [portfoliosList.length ? TxTags.settlement.AddAndAffirmInstruction : TxTags.settlement.AddInstruction]);
```

src/utils/internal.ts
```diff
@@ -534,3 +535,17 @@ export function assertFormatValid(address: string, ss58Format: number): void {
...
+export function getTicker(token: string | SecurityToken): string {
+  return typeof token === 'string' ? token : token.ticker;
+}
+
+/**
+ * @hidden
+ */
+export function calculateMoment(date: Date | undefined, context: Context): Moment | null {
+  return (date && dateToMoment(date, context)) ?? null;
+}
```
monitz87, line 551: We can generalize this to be able to use it with different conversion functions and types

```
/**
 * @hidden
 * Transform a conversion util into a version that returns null if the input is falsy
 */ 
export function optionize<InputType, OutputType>(converter: (input: InputType, context: Context) => OutputType): (val: InputType | null | undefined, context: Context) => OutputType | null {
  return (value: InputType | null = null, context: Context): OutputType | null => value && converter(value, context);
}
```

It receives a conversion function and returns a version of the function that returns null if the argument is falsy

Example:

```
const optionizedFunc = optionize(dateToMoment);
const maybeNull = optionizedFunc(someDate, context); // Moment | null

// or even
const result = optionize(dateToMoment)(someDate, context); // Moment | null
```

monitz87 (changes requested):

src/api/entities/Venue/__tests__/index.ts
```diff
@@ -303,10 +303,9 @@ describe('Venue class', () => {
 
 describe('addInstructionTransformer', () => {
   test('should return a single Instruction', () => {
-    const fakeContext = {} as Context;
     const id = new BigNumber(1);
 
-    const result = addInstructionTransformer([new Instruction({ id }, fakeContext)]);
+    const result = addInstructionTransformer([entityMockUtils.getInstructionInstance()]);
```
monitz87, line 308:
```suggestion
    const result = addInstructionTransformer([entityMockUtils.getInstructionInstance({ id })]);
```

src/api/procedures/addInstruction.ts
```diff
@@ -104,10 +105,31 @@ async function getTxArgsAndErrors(
...
+    SettlementType,
+    Moment | null,
+    Moment | null,
+    {
+      from: PortfolioId;
+      to: PortfolioId;
+      asset: Ticker;
+      amount: Balance;
+    }[]
+  ][];
```
monitz87, line 132: Name these types outside the function to avoid so much repetition (`AddAndAffirmInstructionParams` and `AddInstructionParams`)

src/utils/internal.ts
```diff
@@ -548,6 +548,66 @@ export function getTicker(token: string | SecurityToken): string {
...
+ * Transform a conversion util into a version that returns null if the input is falsy
+ */
+export function optionize<InputType, OutputType>(
+  converter: (input: InputType, context: Context) => OutputType
+): (val: InputType | null | undefined, context: Context) => OutputType | null {
+  return (value: InputType | null | undefined, context: Context): OutputType | null => {
+    value = value ?? null;
+    return value && converter(value, context);
+  };
+}
```
monitz87: Put this one at the end so the other two are together (better readability).

Also, this function isn't being tested. Test it with a simple function that converts a number to a string or something, and then see that it returns null when you pass null or undefined, and that it returns the proper value if not

src/utils/internal.ts
```diff
@@ -548,6 +548,66 @@ export function getTicker(token: string | SecurityToken): string {
...
+
+/**
+ * @hidden
+ * Transform a conversion util into a version that returns null if the input is falsy
+ */
+export function optionize<InputType, OutputType>(
+  converter: (input: InputType, context: Context) => OutputType
+): (val: InputType | null | undefined, context: Context) => OutputType | null {
+  return (value: InputType | null | undefined, context: Context): OutputType | null => {
+    value = value ?? null;
```
monitz87, line 593: Avoid assigning values to function parameters. Use a different variable for this

monitz87 (changes requested):

src/api/procedures/addInstruction.ts
```diff
@@ -70,6 +70,39 @@ export interface Storage {
...
+    asset: Ticker;
+    amount: Balance;
+  }[],
+  PortfolioId[]
+][];
+
+/**
+ * @hidden
+ */
+export type InternalAddInstructionParams = [
```
monitz87, line 93:
```suggestion
type AddAndAffirmInstructionParams = [
  u64,
  SettlementType,
  Moment | null,
  Moment | null,
  {
    from: PortfolioId;
    to: PortfolioId;
    asset: Ticker;
    amount: Balance;
  }[],
  PortfolioId[]
][];

/**
 * @hidden
 */
type InternalAddInstructionParams = [
```

monitz87 (changes requested):

monitz87 (approved):

PR #414, opened 2021-04-06
fix: fetch identity balance for a given checkpoint from RPC

VictorVicente:


VictorVicente:
RPC is going to be removed so it is not possible to fetch balance from there.

PR #417, opened 2021-04-07
feat(types): upgrade types according to Polymesh 3.0.1

VictorVicente:


src/utils/conversion.ts
```diff
@@ -1504,6 +1504,10 @@ export function claimToMeshClaim(claim: Claim, context: Context): MeshClaim {
       );
       break;
     }
+    case ClaimType.InvestorUniquenessV2: {
+      value = tuple(stringToCddId(claim.cddId, context));
```
monitz87, line 1508: Why a tuple?

monitz87 (changes requested):

src/utils/conversion.ts
VictorVicente: It shouldn't be a tuple

monitz87 (approved):

PR #425, opened 2021-04-14
refactor(procedures): make all procedures into arrow functions

monitz87:
This encapsulates procedure data so that the same procedure can be run multiple times in parallel


VictorVicente (changes requested):
There are two code smells to fix

VictorVicente (approved):

PR #429, opened 2021-04-20
feat(corporate actions): add `setDefaults` to the namespace

monitz87:
Allows to set default targets, global tax withholding % and per-identity tax withholding %


VictorVicente (changes requested):
You should add checks for max amount of targets and withholding entries.

VictorVicente (approved):

PR #432, opened 2021-04-23
feat: getWithheldTax method

shuffledex:


monitz87 (changes requested):

src/api/entities/DividendDistribution/index.ts
```diff
@@ -277,27 +285,59 @@ export class DividendDistribution extends CorporateAction {
   }
 
   /**
-   * Retrieve the current amount of withheld tax for a given distribution
+   * Retrieve the amount of taxes that have been withheld up to this point in this Distribution. Optionally, `from` and `to` dates can be passed to retrieve the amount of taxes withheld during a specific period
```
monitz87, line 288:
```suggestion
   * Retrieve the amount of taxes that have been withheld up to this point in this Distribution
```

src/api/entities/DividendDistribution/index.ts
```diff
@@ -277,27 +285,59 @@ export class DividendDistribution extends CorporateAction {
...
-        toDate: to ? to.toISOString().split('T')[0] : null,
+        fromDate: null,
+        toDate: null,
       })
     );
 
-    if (result.data.getWithholdingTaxesOfCA) {
-      return new BigNumber(result.data.getWithholdingTaxesOfCA.taxes);
-    }
+    const withholdingTaxesOfCA = result.data.getWithholdingTaxesOfCA;
```
monitz87, line 303: `const { taxes } = result.data.getWithholdingTaxesOfCA!;`

We already confirmed with the harvester team that it can't be null

monitz87 (changes requested):

monitz87 (approved):

PR #435, opened 2021-04-27
feat: get checkpoints with pagination

shuffledex:


src/api/entities/SecurityToken/Checkpoints/__tests__/index.ts
```diff
@@ -87,30 +81,66 @@ describe('Checkpoints class', () => {
...
+          ticker: 'OTHERTICKER',
+          checkpointId: new BigNumber(2),
+          balance: new BigNumber(1000),
+          moment: new Date('11/11/2020'),
+        },
+      ];
+
+      totalSupply.forEach(({ ticker: t }) =>
+        stringToTickerStub.withArgs(t, context).returns(dsMockUtils.createMockTicker(t))
+      );
```
monitz87, line 108: The ticker would always be the same since you're getting the checkpoints for a specific token. I see no need in having it defined for each entry

src/api/entities/SecurityToken/Checkpoints/index.ts
```diff
@@ -37,27 +45,48 @@ export class Checkpoints extends Namespace<SecurityToken> {
...
-          checkpoint: new Checkpoint({ id: u64ToBigNumber(id), ticker }, context),
-          createdAt: momentToDate(timestamp),
-        }))
-        // the query also returns the next scheduled checkpoint for every schedule (which haven't been created yet)
-        .filter(({ createdAt }) => createdAt <= now)
+    const checkpoints = entries.map(([{ args: [, id] }, balance]) => ({
+      tuple: tuple(id, rawTicker),
+      id,
+      balance,
+    }));
```
monitz87, line 72: You can avoid having to iterate once again over this to create the args for the multi if you use a forEach and two separate arrays

monitz87: Also, there is no reason not to convert id into a Checkpoint and the balance to a BigNumber at this point

monitz87 (changes requested):

src/api/entities/SecurityToken/Checkpoints/__tests__/index.ts
```diff
@@ -88,15 +88,17 @@ describe('Checkpoints class', () => {
...
         {
-          ticker: 'FAKETICKER',
+          ticker: fakeTicker,
           checkpointId: new BigNumber(1),
           balance: new BigNumber(100),
           moment: new Date('10/10/2020'),
         },
         {
-          ticker: 'OTHERTICKER',
+          ticker: fakeTicker,
```
monitz87: Why is it still a property in each entry? You can use the variable directly later on instead of getting it from the entries in the array

monitz87: Same for rawTicker, you can just define it once and then use it where needed

monitz87 (changes requested):

src/api/entities/SecurityToken/Checkpoints/__tests__/index.ts
```diff
@@ -88,36 +89,29 @@ describe('Checkpoints class', () => {
...
 
-      totalSupply.forEach(({ ticker: t }) =>
-        stringToTickerStub.withArgs(t, context).returns(dsMockUtils.createMockTicker(t))
-      );
+      stringToTickerStub.withArgs(ticker, context).returns(rawTicker);
 
-      const rawTotalSupply = totalSupply.map(({ ticker: t, checkpointId, balance }) => ({
-        ticker: dsMockUtils.createMockTicker(t),
+      const rawTotalSupply = totalSupply.map(({ checkpointId, balance }) => ({
+        ticker: rawTicker,
```
monitz87, line 108: This shouldn't be in the array either

monitz87 (changes requested):

monitz87 (approved):

PR #442, opened 2021-04-30
🐛 hasPermissions needs to honor check on TxGroups

hlminh2000:
This is causing a failure in https://github.com/PolymathNetwork/Polymesh-Issuer/pull/641/checks?check_run_id=2467642844.

### DIFF
--- a/src/api/entities/CurrentAccount.ts
+++ b/src/api/entities/CurrentAccount.ts
@@ -38,7 +38,7 @@ export class CurrentAccount extends Account {
       return {
         tokens: null,
         transactions: null,
-        transactionGroups: [],
+        transactionGroups: null,
         portfolios: null,
       };
     }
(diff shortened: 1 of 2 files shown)


src/api/entities/CurrentAccount.ts
```diff
@@ -73,6 +74,15 @@ export class CurrentAccount extends Account {
       hasTokens = tokens.length === 0 || !differenceBy(tokens, currentTokens, 'ticker').length;
     }
 
+    let hasTransactionGroups: boolean;
+    if(currentTransactionGroups === null) {
+      hasTransactionGroups = true;
+    } else if (transactionGroups === null ) {
+      hasTransactionGroups = false;
+    } else {
+      hasTransactionGroups = transactionGroups.length === 0 || !difference(transactionGroups, currentTransactionGroups).length
```
hlminh2000, line 83: @monitz87 I implemented it this so it's consistent with the rest, but the value returned by `this.getPermissions` is `[]` which implies no permission granted for primary key, is this right?

https://github.com/PolymathNetwork/polymesh-sdk/blob/a0872cf42ccbf170d779a6d14bcb69b943311e82/src/api/entities/CurrentAccount.ts#L41

VictorVicente:
This wasn't the issue. 
TxGroups are converted to TxTags to be evaluated. 
But PrimaryKeys has full account permissions always and they can't be restricted.

Thank you anyways Minh for raise it up

PR #454, opened 2021-05-10
chore: if keys are frozen, getAuthorization in procedures should fail

shuffledex:


src/api/entities/__tests__/Account.ts
```diff
@@ -229,4 +229,39 @@ describe('Account class', () => {
       expect(result.next).toBeNull();
     });
   });
+
+  describe('method: isFrozen', () => {
+    test('should return if the Account is frozen or not', async () => {
```
monitz87, line 234: You should also add the case where it is true

src/types/index.ts
```diff
@@ -852,6 +852,7 @@ export interface PortfolioMovement {
 export interface ProcedureAuthorizationStatus {
   permissions: boolean;
   roles: boolean;
+  isFrozenKey: boolean;
```
monitz87, line 855: I think accountFrozen is a better name

src/base/Procedure.ts
```diff
@@ -204,7 +207,14 @@ export class Procedure<
 
       await this.setup(procArgs, context);
 
-      const { roles, permissions } = await this._checkAuthorization(procArgs, context);
+      const { roles, permissions, isFrozenKey } = await this._checkAuthorization(procArgs, context);
+
+      if (isFrozenKey) {
+        throw new PolymeshError({
+          code: ErrorCode.NotAuthorized,
+          message: "Current Account can't execute this procedure with frozen keys",
```
monitz87, line 215: Current Account can't execute this procedure because it is frozen

monitz87 (changes requested):

monitz87 (approved):

PR #455, opened 2021-05-11
feat(identity): add `getPendingDistributions`

monitz87:


src/api/entities/Identity/index.ts
```diff
@@ -506,4 +510,75 @@ export class Identity extends Entity<UniqueIdentifiers> {
...
+     *   - This Identity has already been paid
+     */
+    return P.filter(
+      distributions,
+      async ({ distribution, details: { remainingFunds } }): Promise<boolean> => {
+        const { expiryDate, perShare, ticker, id: localId, paymentDate } = distribution;
+
+        const isExpired = expiryDate && expiryDate < now;
+        const hasNotStarted = paymentDate > now;
+        const noFunds = remainingFunds.eq(0);
```
VictorVicente, line 551: We should remove this filter from here. Pending distributions should be there even if they have not enough remaining funds to pay. 
That's the only way the investor could know the problem.
We should add this check for paying or claiming dividends.

VictorVicente (changes requested):

VictorVicente (approved):

PR #461, opened 2021-05-17
Chore/msdk 584 yarn warnings

monitz87:


.github/workflows/docs.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: '10.x'
+          node-version: '12.x'
```
VictorVicente, line 17: Should we move directly to 14 instead? It is the current active LTS release. Would it affect us?

typedoc.json
```diff
@@ -1,7 +1,6 @@
 {
-  "mode": "file",
   "out": "docs",
-  "theme": "gitbook",
+  "theme": "markdown",
```
VictorVicente, line 3: I don't know if it would work for our docs. Have you tried to generate them with this theme? If this typedoc version does not support gitbook integration, we should check how to solve it.

VictorVicente (changes requested):

.github/workflows/docs.yml
VictorVicente: From @monitz87: "Node 14 resulted in warnings when using commitizen, so I decided to go with 12 just to be safe"

VictorVicente (approved):

PR #475, opened 2021-06-09
feat: sort tickers

shuffledex:


VictorVicente (changes requested):
There should be a test checking we're sorting them

monitz87 (commented):
LGTM

VictorVicente (approved):

PR #479, opened 2021-06-14
feat: adapt permissions to the new format

monitz87:


src/utils/conversion.ts
```diff
@@ -668,16 +674,46 @@ export function txGroupToTxTags(group: TxGroup): TxTag[] {
  * @note tags that don't belong to any group will be ignored.
  *   The same goes for tags that belong to a group that wasn't completed
  */
-export function txTagsToTxGroups(tags: TxTag[]): TxGroup[] {
+export function txTagsToTxGroups(permissions: TransactionPermissions): TxGroup[] {
```
VictorVicente, line 677: Should we change the name to  `transactionPermissionsToTxGroups`?

VictorVicente (changes requested):

src/utils/conversion.ts
```diff
@@ -687,42 +723,153 @@ export function permissionsToMeshPermissions(
...
+          } else if (pallet === undefined) {
+            pallet = extrinsicDict[palletName] = { tx: [] };
+          }
 
-        const pallet = (extrinsicDict[palletName] = extrinsicDict[palletName] || []);
+          pallet.tx.push(dispatchableName);
+        } else {
+          const pallet = extrinsicDict[tag];
+
+          if (pallet) {
```
shuffledex, line 758: `pallet` never will defined at this point because the only way to accomplish that is removing the sort()

src/utils/conversion.ts
```diff
@@ -822,9 +822,8 @@ export function permissionsToMeshPermissions(
   let asset: PermissionsEnum<Ticker> = 'Whole';
   if (tokens) {
     const { values: tokenValues, type } = tokens;
-    const tickers = tokenValues
-      .sort(({ ticker: tickerA }, { ticker: tickerB }) => tickerA.localeCompare(tickerB))
-      .map(({ ticker }) => stringToTicker(ticker, context));
+    tokenValues.sort(({ ticker: tickerA }, { ticker: tickerB }) => tickerA.localeCompare(tickerB));
+    const tickers = tokenValues.map(({ ticker }) => stringToTicker(ticker, context));
```
monitz87, line 836: Why was this change necessary?

VictorVicente (approved):

PR #481, opened 2021-06-17
chore: fix conversions utils tests

shuffledex:


src/utils/__tests__/conversion.ts
```diff
@@ -1193,35 +1231,47 @@ describe('permissionsToMeshPermissions and meshPermissionsToPermissions', () =>
...
-      ],
-      transactionGroups: [],
-      portfolios: [entityMockUtils.getDefaultPortfolioInstance({ did })],
+      tokens: {
+        values: [entityMockUtils.getSecurityTokenInstance({ ticker })],
+        type: PermissionType.Include,
+      },
+      transactions: {
+        type: PermissionType.Include,
+        values: [ModuleName.Identity, ModuleName.Confidential],
```
monitz87, line 1240: This would suggest permissions over the entire Identity module, which is not what the `permissions` variable suggeests.

Also, having the Identity module as a whole and then also having `ClaimsManagement` as a group is redundant

monitz87 (changes requested):

src/testUtils/mocks/dataSources.ts
```diff
@@ -1941,6 +1946,33 @@ export const createMockPermissions = (permissions?: {
...
+
+/**
+ * @hidden
+ */
+export const createMockPortfolioPermissions = (
+  assetPermissions?: 'Whole' | { These: PortfolioId[] } | { Except: PortfolioId[] }
+): PortfolioPermissions => {
+  return createMockEnum(assetPermissions) as PortfolioPermissions;
+};
+
```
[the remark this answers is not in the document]
monitz87: Ignore this, I just saw that it is being addressed in another PR

src/utils/__tests__/conversion.ts
monitz87: This too is being fixed in another PR, so ignore

monitz87 (approved):

PR #482, opened 2021-06-17
chore: fix tests

shuffledex:


src/base/__tests__/Context.ts
```diff
@@ -826,17 +832,17 @@ describe('Context class', () => {
...
           }),
           dsMockUtils.createMockSecondaryKey({
             signer: signerAccountId,
             permissions: dsMockUtils.createMockPermissions({
-              asset: null,
-              portfolio: null,
-              extrinsic: null,
+              asset: dsMockUtils.createMockAssetPermissions(),
+              extrinsic: dsMockUtils.createMockExtrinsicPermissions(),
+              portfolio: dsMockUtils.createMockPortfolioPermissions(),
```
monitz87, line 845: You should use `Whole` for these. Empty codecs are not a possibility in this case

monitz87 (changes requested):

monitz87 (approved):

PR #487, opened 2021-06-25
feat(settlements): add endpoint to fetch a venue by its id

VictorVicente:


src/__tests__/Settlements.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  afterAll(() => {
+    dsMockUtils.cleanup();
+    entityMockUtils.cleanup();
+  });
+
+  describe('method: getVenue', () => {
+    test('should return a Venue by its id', async () => {
+      const venueId = 1;
+      const matchingVenue = new Venue({ id: new BigNumber(venueId) }, context);
```
monitz87, line 40: The convention is to use entityMockUtils.getVenueInstance({ id }). We try to avoid instancing entities in tests because it has caused problems in the past

src/Settlements.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(context: Context) {
+    this.context = context;
+  }
+
+  /**
+   * Retrieve a Venue by its id
+   *
+   * @param id - Identifier number of the venue
+   */
+  public async getVenue(id: number | BigNumber): Promise<Venue> {
```
monitz87, line 25: The parameters should be `{ id: BigNumber }`

monitz87 (changes requested):

monitz87 (approved):

PR #488, opened 2021-06-25
feat(settlements): add getInstruction endpoint

VictorVicente:
Previous behaviour of `Instruction.exists` is replaced by `Instrucion.isPending`.
`Instruction.exists` now checks if instruction has existed or not in the chain

BREAKING CHANGE: Previous behaviour of `Instruction.exists` is replaced by `Instrucion.isPending`


monitz87 (changes requested):

src/api/entities/Instruction/index.ts
```diff
@@ -130,17 +130,15 @@ export class Instruction extends Entity<UniqueIdentifiers> {
       context,
     } = this;
 
-    const { instruction_id: instructionId, status } = await settlement.instructionDetails(
-      numberToU64(id, context)
-    );
-
-    if (instructionId.isEmpty) {
+    if (!(await this.exists())) {
```
monitz87, line 133: We try to avoid inline awaits because of readability concerns

So
```
const exists = await this.exists();

if (!exists) {
  // ...
}
```

src/api/entities/Instruction/index.ts
```diff
@@ -157,14 +155,11 @@ export class Instruction extends Entity<UniqueIdentifiers> {
...
-      context,
     } = this;
 
-    const { instruction_id: instructionId } = await settlement.instructionDetails(
-      numberToU64(id, context)
-    );
+    const instructionCounter = await settlement.instructionCounter();
 
-    return !instructionId.isEmpty;
+    return id.isLessThanOrEqualTo(u64ToBigNumber(instructionCounter));
```
monitz87, line 162: We're using `lte` instead of `isLessThanOrEqualTo` everywhere else

src/api/entities/Instruction/index.ts
```diff
@@ -181,6 +176,13 @@ export class Instruction extends Entity<UniqueIdentifiers> {
       context,
     } = this;
 
+    if (!(await this.exists())) {
```
monitz87, line 179: Same as above

monitz87 (changes requested):

monitz87 (approved):

PR #490, opened 2021-06-26
feat(instructions): add endpoint to fetch current instruction status

VictorVicente:


src/api/entities/Instruction/index.ts
```diff
@@ -307,6 +321,42 @@ export class Instruction extends Entity<UniqueIdentifiers> {
...
+    if (eventIdentifier) {
+      return {
+        status: InstructionStatus.Executed,
+        eventIdentifier,
+      };
+    }
+
+    throw new PolymeshError({
+      code: ErrorCode.FatalError,
+      message: "It wasn't possible to determine current status for this instruction",
```
monitz87, line 356:
```suggestion
      message: "It isn't possible to determine the current status of this Instruction",
```

src/api/entities/Instruction/types.ts
```diff
@@ -52,3 +54,8 @@ export interface InstructionAffirmation {
   identity: Identity;
   status: AffirmationStatus;
 }
+
+export interface InstructionStatusResult {
+  status: InstructionStatus;
+  eventIdentifier?: EventIdentifier;
+}
```
monitz87, line 61: For more type safety:

```
export type InstructionStatusResult = {
  status: InstructionStatus.Pending;
} | {
  staus: Exclude<InstructionStatus, InstructionStatus.Pending>;
  eventIdentifier: EventIdentifier;
}
```

src/api/entities/Instruction/index.ts
```diff
@@ -181,15 +195,15 @@ export class Instruction extends Entity<UniqueIdentifiers> {
...
+    if (status === InternalInstructionStatus.Unknown) {
       throw new PolymeshError({
         code: ErrorCode.FatalError,
         message: notPendingMessage,
       });
     }
 
     const details = {
-      status,
+      status: InstructionStatus.Pending,
```
monitz87, line 206: Why is the explicit assignment necessary? Isn't it enough to use `status`?

src/api/entities/Instruction/index.ts
```diff
@@ -307,6 +321,42 @@ export class Instruction extends Entity<UniqueIdentifiers> {
     };
   }
 
+  /**
+   * Retrieve current status of this Instruction
```
monitz87, line 325: The middleware note is missing

monitz87 (changes requested):

src/api/entities/Instruction/index.ts
VictorVicente: `status` variable's type is `InternalInstructionStatus` which represents `Pending` or `Unknown`. 
`status` in details is `InstructionStatus` which represents `Pending`, `Executed` or `Failed`. 
I had to choose to force `Pending` typing or this explicit assignment which sounds like the best option for me.

monitz87 (approved):

PR #494, opened 2021-06-30
feat: add opcional memo field to move portfolio funds

shuffledex:


src/types/index.ts
```diff
@@ -917,6 +917,7 @@ export type PermissionsLike = {
 export interface PortfolioMovement {
   token: string | SecurityToken;
   amount: BigNumber;
+  memo?: string;
```
monitz87, line 920:
```suggestion
  /**
   * identifier string to help differentiate transfers
   */
  memo?: string;
```

monitz87 (changes requested):
Aside from that detail it looks good

monitz87 (approved):

PR #495, opened 2021-07-01
fix: fix getSecondaryKeys to return them when no account is set

VictorVicente:


src/testUtils/mocks/entities.ts
```diff
@@ -592,22 +589,11 @@ const defaultIdentityOptions: IdentityOptions = {
...
-  authorizations: {
-    getReceived: [],
-  },
-  getVenues: [],
-  getScopeId: 'someScopeId',
   areScondaryKeysFrozen: false,
 };
-let currentIdentityOptions: CurrentIdentityOptions = defaultCurrentIdentityOptions;
+let identityOptions: IdentityOptions = defaultIdentityOptions;
+let currentIdentityOptions: IdentityOptions = defaultIdentityOptions;
```
monitz87, line 596:
```suggestion
let currentIdentityOptions: IdentityOptions = { ...defaultIdentityOptions };
```

src/api/procedures/__tests__/inviteAccount.ts
```diff
@@ -109,16 +109,20 @@ describe('inviteAccount procedure', () => {
...
+          {
+            signer,
+            permissions: {
+              tokens: [],
+              portfolios: [],
+              transactions: [],
+              transactionGroups: [],
+            },
+          },
+        ],
```
monitz87, line 122: Is this still necessary? Shouldn't this option be removed from `ContextOptions` entirely?

monitz87 (changes requested):

src/api/procedures/__tests__/inviteAccount.ts
monitz87: Ignore this

monitz87 (approved):

PR #496, opened 2021-07-05
feat: refactor createST procedure to make totalSupply as option

shuffledex:


src/api/procedures/createSecurityToken.ts
```diff
@@ -118,16 +121,26 @@ export async function prepareCreateSecurityToken(
...
+      rawFundingRound
+    );
+  } else {
+    this.addBatchTransaction(tx.asset.createAsset, { fee }, [
+      tuple(rawName, rawTicker, rawIsDivisible, rawType, rawIdentifiers, rawFundingRound),
+    ]);
+
+    const rawTotalSupply = numberToBalance(totalSupply, context, isDivisible);
+
+    this.addBatchTransaction(tx.asset.issue, {}, [tuple(rawTicker, rawTotalSupply)]);
```
monitz87, line 142: I forgot that batching different kinds of transactions isn't currently supported by the SDK. What you're doing here is adding two separate batches.

Until we support that, I suggest just using `addTransaction` here. You should refactor the code like so:

- create token
- if totalSupply exists and is non-zero, add an issue tx

monitz87 (changes requested):

src/api/procedures/__tests__/createSecurityToken.ts
```diff
@@ -264,21 +264,11 @@ describe('createSecurityToken procedure', () => {
       null
     );
 
-    const addBatchTransactionStub = procedureMockUtils.getAddBatchTransactionStub();
+    const createAssetIssuen = dsMockUtils.createTxStub('asset', 'issue');
```
monitz87, line 267: `issueTransaction`

monitz87 (changes requested):

monitz87 (approved):

PR #514, opened 2021-07-27
Feat/msdk 641 grouped instructions

monitz87:


monitz87:
I propose we ignore the duplicated lines for this case. Otherwise we'd have to do a bunch of refactoring which will serve no purpose after we delete the deprecated methods

VictorVicente (approved):

PR #515, opened 2021-07-29
feat(instruction): add `reschedule` method for failed instructions

monitz87:


src/api/procedures/rescheduleInstruction.ts
```diff
@@ -0,0 +1,103 @@
...
+  const {
+    storage: {
+      instructionDetails: {
+        venue: { id },
+      },
+    },
+  } = this;
+
+  return {
+    roles: [{ type: RoleType.VenueOwner, venueId: id }],
```
VictorVicente, line 75: Are you sure you need to be the venue owner to execute it? My guess is that everyone can reschedule a failed instruction.

VictorVicente (approved):

PR #527, opened 2021-08-10
feat: getPermissionGroup method implementation

shuffledex:


src/api/entities/Agent.ts
```diff
@@ -34,4 +41,29 @@ export class Agent extends Identity {
...
+      ticker,
+      did,
+    } = this;
+
+    const rawTicker = stringToTicker(ticker, context);
+    const rawIdentityId = stringToIdentityId(did, context);
+
+    const rawGroupPermissions = await externalAgents.groupOfAgent(rawTicker, rawIdentityId);
+
+    const agentGroup = rawGroupPermissions.unwrap();
```
monitz87, line 65: This could be `None`. Throw an error in that case. Something like "This agent is no longer part of a permission group"

src/api/entities/__tests__/Agent.ts
```diff
@@ -1,5 +1,10 @@
+import sinon from 'sinon';
+
+import { KnownPermissionGroup } from '~/api/entities/KnownPermissionGroup';
```
monitz87, line 3: Import from internal

src/api/entities/__tests__/Agent.ts
```diff
@@ -46,4 +51,24 @@ describe('Agent class', () => {
       expect(Agent.isUniqueIdentifiers({ did: 1 })).toBe(false);
     });
   });
+
+  describe('method: getPermissionGroup', () => {
+    test('should return te agent group associated with the actual Agent', async () => {
```
monitz87, line 56:
```suggestion
    test('should return the permission group associated with the Agent', async () => {
```

monitz87 (changes requested):

src/api/entities/Agent.ts
monitz87: Let's change the text to "This Identity is no longer an Agent for this Security Token"

monitz87 (approved):

PR #532, opened 2021-08-12
fix: remove deprecated PIA and CAA roles

VictorVicente:


src/api/entities/SecurityToken/CorporateActions/index.ts
```diff
@@ -93,18 +90,12 @@ export class CorporateActions extends Namespace<SecurityToken> {
    *
    * @note this action will leave the Security Token owner as the Corporate Actions Agent
    *
-   * @note required role:
-   *   - Security Token Owner
-   *
    * @deprecated
    */
   public removeAgent: ProcedureMethod<void, void>;
```
monitz87, line 95: Shouldn't we remove this procedure while we're at it?

src/api/entities/SecurityToken/CorporateActions/index.ts
```diff
@@ -81,9 +81,6 @@ export class CorporateActions extends Namespace<SecurityToken> {
    *   the corresponding Account. An Account or Identity can
    *   fetch its pending Authorization Requests by calling `authorizations.getReceived`
    *
-   * @note required role:
-   *   - Security Token Owner
-   *
    * @deprecated in favor of `inviteAgent`
    */
   public setAgent: ProcedureMethod<ModifyCorporateActionsAgentParams, void>;
```
monitz87, line 86: This one as well

monitz87 (changes requested):
Should we list the required account/agent permissions for every procedure? (Maybe as a separate task)

src/api/entities/SecurityToken/CorporateActions/index.ts
VictorVicente: Not yet. The idea to keep the backward compatibility is to give users time to replace these methods with External Agents calls. 
Given that EA is not completely finished, we should keep them for a while. To make it easier to users to update, they should be available at the same time so they can refactor their code.
We will be removing these methods later.

src/api/entities/SecurityToken/CorporateActions/index.ts
VictorVicente: Same as above

PR #533, opened 2021-08-13
Feat/msdk 661 remove current identity account

monitz87:


src/api/procedures/inviteAccount.ts
```diff
@@ -66,15 +67,6 @@ export async function prepareInviteAccount(
...
 
-  const isPresent = !!secondaryKeys.find(({ signer }) => signerToString(signer) === address);
-
-  if (isPresent) {
-    throw new PolymeshError({
-      code: ErrorCode.ValidationError,
-      message: 'The target Account is already a secondary key for this Identity',
-    });
-  }
-
```
VictorVicente, line 77: Why did you remove this validation?

src/api/procedures/leaveIdentity.ts
```diff
@@ -20,15 +20,20 @@ export async function prepareLeaveIdentity(
...
+
+  const currentIdentity = await account.getIdentity();
 
-  const currentIdentity = await context.getCurrentIdentity();
+  if (!currentIdentity) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'There is no Identity associated to this Account',
+    });
+  }
```
VictorVicente, line 33: This procedure should require that account is current Account too. Otherwise it could generate misleading behaviors

VictorVicente (changes requested):
I'm not sure if I'm understanding how the IdentityRole works. How does it ensure Identity is just the current?

src/api/procedures/inviteAccount.ts
VictorVicente: Ah ok. This error can not be thrown because if the key has any Identity, the previous error would throw first.

VictorVicente (approved):

PR #560, opened 2021-09-06
feat: create namespace for token permissions

monitz87:
create `tokenPermissions` namespace in the `Identity` Entity to handle all External Agents
  functionality

BREAKING CHANGE:
- move `identity.hasTokenPermissions` to `identity.tokenPermissions.hasPermissions`
- move `identity.getTokenPermissions` to `identity.tokenPermissions.get`
- move `identity.waivePermissions` to `identity.tokenPermissions.waive`
- change the type of `AgentWithGroup.agent` from `Agent` to `Identity`. This affects the return
  type of `securityToken.permissions.getAgents`
- remove the `Agent` entity:
    - move `agent.setPermissionGroup` to `identity.tokenPermissions.setGroup`. It now also
      receives the token as a parameter when not passing an existing group. It now throws
      an error if the target is not already an agent
    - move `agent.getPermissionGroup` to `identity.tokenPermissions.getGroup`. It now also
      receives the token as a parameter
    - move `agent.addedAt` to `identity.tokenPermissions.grantedAt`. It now also receives
      the token as a parameter


src/api/entities/Identity/TokenPermissions.ts
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  /**
+   * Retrieve the identifier data (block number, date and event index) of the event that was emitted when this Identity was granted
+   *   permissions over a specific Security Token
+   *
+   * @note uses the middleware
+   * @note there is a possibility that the data is not ready by the time it is requested. In that case, `null` is returned
+   */
+  public async grantedAt({
```
VictorVicente, line 262: I think it should be renamed to agentAddedAt or something similar to keep it similar to the event we are fetching.

VictorVicente (changes requested):

VictorVicente (approved):

PR #577, opened 2021-09-27
chore: 🤖 make setassetrequirements params more flexible

shuffledex:


src/types/index.ts
```diff
@@ -366,7 +366,7 @@ export type MultiClaimCondition = ConditionBase & {
 
 export type IdentityCondition = ConditionBase & {
   type: ConditionType.IsIdentity;
-  identity: Identity;
+  identity: Identity | string;
```
monitz87, line 369: This would affect the type when you're returning conditions to the user. The idea was to only change the type that is used for input

src/api/procedures/setAssetRequirements.ts
```diff
@@ -74,6 +75,29 @@ export async function prepareSetAssetRequirements(
     });
   }
 
+  const identityErr: string[] = [];
+
+  await P.each(flattenDeep<Condition>(requirements), async requirement => {
+    if (requirement.type === ConditionType.IsIdentity) {
+      let { identity } = requirement;
+      identity = identity instanceof Identity ? identity : new Identity({ did: identity }, context);
+      const exits = await identity.exists();
```
monitz87, line 84: exists

src/api/procedures/setAssetRequirements.ts
```diff
@@ -74,6 +75,29 @@ export async function prepareSetAssetRequirements(
     });
   }
 
+  const identityErr: string[] = [];
```
monitz87, line 78: invalidDids

src/api/procedures/setAssetRequirements.ts
```diff
@@ -74,6 +75,29 @@ export async function prepareSetAssetRequirements(
...
+      if (!exits) {
+        identityErr.push(identity.did);
+      }
+    }
+  });
+
+  if (identityErr.length) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'Some identities no longer exists',
```
monitz87, line 94: `Some of the passed "isIdentity" conditions reference Identities that do not exist`

Identities can't be deleted, so "no longer" doesn't apply here

src/utils/conversion.ts
```diff
@@ -2155,9 +2155,8 @@ export function requirementToComplianceRequirement(
       const { claims } = condition;
       conditionContent = claims.map(claim => claimToMeshClaim(claim, context));
     } else if (condition.type === ConditionType.IsIdentity) {
-      const {
-        identity: { did },
-      } = condition;
+      const { identity } = condition;
+      const did = identity instanceof Identity ? identity.did : identity;
```
monitz87, line 2159: Since the `Condition` type should remain the same, I suggest not making any changes to this function, instead converting the inputs in the procedure from `Identity | string` to `Identity` at the same time that you're checking for existence, and then pass the converted result to this function

monitz87 (changes requested):

src/api/entities/DividendDistribution/__tests__/index.ts
```diff
@@ -301,7 +301,7 @@ describe('DividendDistribution class', () => {
     });
   });
 
-  describe('method: getParticipants', () => {
+  describe.only('method: getParticipants', () => {
```
monitz87, line 304:
```suggestion
  describe('method: getParticipants', () => {
```

src/types/index.ts
```diff
@@ -379,6 +380,10 @@ export type Condition =
   | IdentityCondition
   | ExternalAgentCondition;
 
+export type RequirementCondition =
```
monitz87, line 383: ConditionLike

src/api/procedures/setAssetRequirements.ts
```diff
@@ -26,6 +26,28 @@ export type Params = SetAssetRequirementsParams & {
...
+        const { identity } = requirement;
+        return {
+          ...requirement,
+          identity:
+            identity instanceof Identity ? identity : new Identity({ did: identity }, context),
+        };
+      }
+      return requirement;
+    });
+  }) as unknown) as Condition[][];
```
monitz87, line 48: This type assertion is not necessary

src/api/procedures/setAssetRequirements.ts
```diff
@@ -26,6 +26,28 @@ export type Params = SetAssetRequirementsParams & {
   ticker: string;
 };
 
+/**
+ * @hidden
+ */
+function identityTransformation(
```
monitz87, line 32: You should create a `conditionLikeToCondition(conditionLike: ConditionLike, context: Context): Condition` util in the conversion file and use that

src/api/procedures/setAssetRequirements.ts
```diff
@@ -65,8 +87,10 @@ export async function prepareSetAssetRequirements(
     return !differenceWith(a, b, isEqual).length && a.length === b.length;
   };
 
+  const requirementsCondition = identityTransformation(requirements, context);
```
monitz87, line 90: Call the destructured variable `inputRequirements` and the converted one `requirements`

monitz87 (changes requested):

monitz87:
This is already being handled in #596

PR #582, opened 2021-09-30
Fix/msdk 716/update schema

VictorVicente:


src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -85,14 +85,13 @@ export async function prepareModifyInstructionAffirmation(
...
     case InstructionAffirmationOperation.Withdraw: {
-      excludeCriteria.push(AffirmationStatus.Pending, AffirmationStatus.Rejected);
+      excludeCriteria.push(AffirmationStatus.Pending);
       errorMessage = 'The instruction is not affirmed';
       transaction = settlementTx.withdrawAffirmation;
 
       break;
     }
     case InstructionAffirmationOperation.Reject: {
-      excludeCriteria.push(AffirmationStatus.Rejected);
```
monitz87, line 95: I think this becomes redundant without any exclude criteria, since the error would never be thrown. Should we include `Unknown` in the exclude criteria for every case, or would that already be covered by `assertInstructionValid`? If the latter is the case, then this switch case can be safely removed

VictorVicente: Agreed it's redundant. But we need a default error message value anyways. We could assign it when `errorMessage` is declared or change this `case` to `default`. 
But in terms of output, the result will be exactly the same.
As far as I understand, if Instruction is valid, it can't contain any `Unknown` affirmation.

monitz87: Hmm I don't like having this code branch suggesting there's a potential error when rejecting, when in fact it can never be thrown. I think the "correct" way of doing it is removing the branch, adding a ! to errorMessage when throwing, and a comment explaining that if the error was thrown it means that the operation had to be either affirm or withdraw, and so the message must be set

I know it sounds kinda ugly, but it's better IMO than having useless code

monitz87 (approved):

PR #583, opened 2021-09-30
feat: 🎸 export typeguards for every Entity

monitz87:
They are now exported from `/utils`

BREAKING CHANGE: 🧨 move `isEntity` and `isPolymeshError` from `/types` to `/utils`


VictorVicente:
> Something like `yarn generate:entity` which creates the class and test files, adds the export from `internal`, creates the relevant types in `entities/types` and also a typeguard that doesn't compile so that the developer is forced to implement it properly (or maybe this can be done with the linter)

I really like this idea. I'll create a task for the future and we'll see when we can tackle it.

PR #599, opened 2021-11-08
feat: 🎸 Add initial support for new Middleware V2

raycar5:
Adds support for middlewareV2 on the heartbeat, latestBlock and
investments queries.

✅ Closes: MSDK-721

The thought process for this is that we want to be able to use the old and new middleware at the same time to verify the accuracy of the new middleware in the real world, to achieve this:

1. All v2 query results will be mapped onto the v1 result "shape".
2. The sdk will try to use the v1 middleware if only it is available.
3. The sdk will try to use both middlewares if both are available and check that they return the same thing, if a difference is detected, it will be reported via the `middlewareDiffLogger` (which can be connected to sentry or other error tracking tools) and default to the response of the v1 middleware.
4. The sdk will use the v2 middleware if only it is available.

I'm unsure if this PR should be merged or be left as a feature branch until all queries are migrated what do you think? For now I'll leave it as a draft


src/middleware/queries.ts
```diff
@@ -612,9 +787,40 @@ export function investments(variables: QueryInvestmentsArgs): GraphqlQuery<Query
...
+      query: { query: queryv2, variables },
+      mapper: a => {
+        return {
+          investments: {
+            __typename: 'InvestmentResult',
+            totalCount: a.investments!.totalCount,
+            items: a.investments!.nodes,
+          },
+        };
+      },
```
monitz87, line 820:
```suggestion
      mapper: a => {
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        const { totalCount, nodes: items } = a.investments!
        return {
          investments: {
            __typename: 'InvestmentResult',
            totalCount,
            items,
          },
        };
      },
```

src/polkadot/definitions.ts
```diff
@@ -1,5 +1,3 @@
-/* istanbul ignore file */
-
```
monitz87, line 2: This should be re-added by hand (or maybe by script), otherwise the coverage reports will be skewed

src/types/internal.ts
```diff
@@ -165,6 +168,18 @@ export interface GraphqlQuery<Variables = undefined> {
...
 }
 
+export type MultiGraphqlQuery<
+  Args = undefined,
+  Q2 extends keyof QueryV2 = keyof QueryV2,
+  Q1 extends keyof Query = keyof Query
+> = {
+  v1: GraphqlQuery<Args>;
+  v2?: {
+    query: GraphqlQuery<Args>;
```
monitz87, line 178: What about `payload`, `request`, `data`?

monitz87 (changes requested):
We should add `**/middleware/types*.ts` to `sonar.cpd.exclusions` in `sonar-project.properties`

src/middleware/queries.ts
```diff
@@ -587,9 +647,122 @@ export function settlements(variables: QuerySettlementsArgs): GraphqlQuery<Query
...
+                  to: {
+                    __typename: 'Portfolio',
+                    did: leg.to.did,
+                    kind: leg.to.number === 0 ? 'Default' : leg.to.number.toString(),
+                  },
+                  direction:
+                    leg.from.did === variables.identityId &&
+                    // If this was !portfolioNumber, it could be 0,
+                    // but since variables.portfolioNumber is a string, this is fine.
+                    (!variables.portfolioNumber || portfolioNumber === leg.from.number)
```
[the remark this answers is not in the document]
raycar5: nope, it's in the test cases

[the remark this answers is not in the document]
monitz87: What is in the test cases?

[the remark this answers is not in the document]
raycar5: There exists a settlement in the integration tests which transfers tokens between portfolios of the same did.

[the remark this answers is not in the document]
monitz87: Still, in that case wouldn't the direction be irrelevant? If you're transferring to yourself, it's both incoming and outgoing

src/middleware/queries.ts
```diff
@@ -47,21 +85,109 @@ export function proposalVotes(
...
+        query: queryV2,
+        variables: {
+          ...variables,
+          pipId: variables.pipId.toString(),
+          orderBy: [mapProposalVotesOrder(variables.orderBy)],
+        },
+      },
+      mapper: ({ proposalVotes: results }) => {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const { nodes } = results!;
```
polymath-eric, line 129: null assertion doesn't seem right. 

GraphQL has not null in the schema, and I would have thought the types would be alright if we were returning an empty array. If it can be null, `.map` below will cause a run time error.

src/middleware/queries.ts
```diff
@@ -47,21 +85,109 @@ export function proposalVotes(
...
+        return {
+          proposalVotes: nodes.map(node => {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const { blockId, ...rest } = node!;
```
polymath-eric, line 133: I don't think this would be null in a loop

src/middleware/queries.ts
```diff
@@ -47,21 +85,109 @@ export function proposalVotes(
...
+      return ProposalVotesOrderBy.VoteDesc;
+    }
+  } else if (field === ProposalVotesOrderFields.Weight) {
+    if (order === Order.Asc) {
+      return ProposalVotesOrderBy.WeightAsc;
+    } else {
+      return ProposalVotesOrderBy.WeightDesc;
+    }
+  }
+  // Not smart enough type system
```
polymath-eric, line 172: I would have thought TS would be smart enough if the checks above were exhaustive.

src/middleware/queries.ts
```diff
@@ -141,10 +412,91 @@ export function eventByIndexedArgs(
...
+ */
+function createEventFilter(variables: QueryEventByIndexedArgsArgs): {
+  filter: string;
+  args: string;
+} {
+  let filter = '';
+  let args = ' ';
+  const { eventArg0, eventArg1, eventArg2 } = variables;
+  if (eventArg0 !== undefined) {
+    args += ', $eventArg0: String';
```
polymath-eric, line 486: building up strings seems like a smell to me. Why doesn't Apollo handle this for us?

polymath-eric (commented):
Left some comments, mainly around types.

Having the no-null-assertion in the queries makes me a bit uneasy.  If the GraphQL schema says they can be null, then we should handle the null case. If they really are not nullable, then the types should reflect that.

src/middleware/queries.ts
polymath-eric: It seems like graphql should always return with having an empty nodes array e.g. `proposalVotes: { nodes: [] }`

I think the no-non-null is OK here, but ideally our types would know that.

VictorVicente:
We'll close this PR until we define how Middleware V2 will be.

PR #609, opened 2021-11-30
feat: 🎸 Allow DID string for set compliance requirements

polymath-eric:
Allows for a DID string to be passed in Identity fields for
setAssetRequirements fields. An Identity entity can still be passed

There is [getDid](https://github.com/PolymathNetwork/polymesh-sdk/blob/master/src/utils/internal.ts#L113), but defaulting to the Context Identity seems dangerous in this context. Also the async makes it a breaking change for the conversion.

`getDid` is using the `signerToString` method, and that does what I need, but the name makes it feel weird since its not actually taking a signer in this context.

I was tempted to make a new method like `getDid`, but without the context default + validate the identity exists, but I didn't want to create clutter and have an awkward name. Thoughts on this?

### DIFF
--- a/src/api/entities/SecurityToken/Compliance/TrustedClaimIssuers.ts
+++ b/src/api/entities/SecurityToken/Compliance/TrustedClaimIssuers.ts
@@ -12,11 +12,7 @@ import {
 } from '~/internal';
 import { ProcedureMethod, SubCallback, UnsubCallback } from '~/types';
 import { TrustedClaimIssuerOperation } from '~/types/internal';
-import {
-  signerToString,
-  stringToTicker,
-  trustedIssuerToTrustedClaimIssuer,
-} from '~/utils/conversion';
+import { stringToTicker, trustedIssuerToTrustedClaimIssuer } from '~/utils/conversion';
 import { createProcedureMethod } from '~/utils/internal';
 
 /**
@@ -115,11 +111,11 @@ export class TrustedClaimIssuers extends Namespace<SecurityToken> {
 
     const assembleResult = (issuers: TrustedIssuer[]): DefaultTrustedClaimIssuer[] =>
       issuers.map(issuer => {
-        const { identity, trustedFor } = trustedIssuerToTrustedClaimIssuer(issuer, context);
-        return new DefaultTrustedClaimIssuer(
-          { did: signerToString(identity), ticker, trustedFor },
-          context
-        );
+        const {
+          identity: { did },
+          trustedFor,
+        } = trustedIssuerToTrustedClaimIssuer(issuer, context);
+        return new DefaultTrustedClaimIssuer({ did, ticker, trustedFor }, context);
       });
 
     if (callback) {
--- a/src/api/entities/SecurityToken/Compliance/TrustedClaimIssuers.ts
+++ b/src/api/entities/SecurityToken/Compliance/TrustedClaimIssuers.ts
@@ -12,7 +12,11 @@ import {
 } from '~/internal';
 import { ProcedureMethod, SubCallback, UnsubCallback } from '~/types';
 import { TrustedClaimIssuerOperation } from '~/types/internal';
-import { stringToTicker, trustedIssuerToTrustedClaimIssuer } from '~/utils/conversion';
+import {
+  signerToString,
+  stringToTicker,
+  trustedIssuerToTrustedClaimIssuer,
+} from '~/utils/conversion';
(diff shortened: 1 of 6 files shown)


src/types/index.ts
```diff
@@ -337,7 +337,7 @@ export interface ClaimScope {
 }
 
 export interface TrustedClaimIssuer {
-  identity: Identity;
+  identity: string | Identity;
```
monitz87, line 340: The way we handle these situations in the SDK is by creating a specific `Input` type (`InputTrustedClaimIssuer` or `TrustedClaimIssuerLike`)

Something like 
```
export type InputTrustedClaimIssuer = Modify<TrustedClaimIssuer, {
  identity: string | Identity;
}>;
```

The name is up to you

We can then use that type where users are required to input said values, without affectinig the return types of getters

This also applies to `IdentityCondition`, but leave that one as it was before because I'm working on it on my end and we'll have some pretty awful conflicts otherwise

src/api/entities/SecurityToken/Compliance/__tests__/Requirements.ts
```diff
@@ -200,7 +200,9 @@ describe('Requirements class', () => {
                   }),
                   issuers: [
                     dsMockUtils.createMockTrustedIssuer({
-                      issuer: dsMockUtils.createMockIdentityId(notDefaultClaimIssuer.identity.did),
+                      issuer: dsMockUtils.createMockIdentityId(
+                        (notDefaultClaimIssuer.identity as Identity).did
+                      ),
```
monitz87, line 205: This shouldn't be necessary when the types are separated

src/api/entities/SecurityToken/Compliance/TrustedClaimIssuers.ts
```diff
@@ -111,11 +115,11 @@ export class TrustedClaimIssuers extends Namespace<SecurityToken> {
...
-        const {
-          identity: { did },
-          trustedFor,
-        } = trustedIssuerToTrustedClaimIssuer(issuer, context);
-        return new DefaultTrustedClaimIssuer({ did, ticker, trustedFor }, context);
+        const { identity, trustedFor } = trustedIssuerToTrustedClaimIssuer(issuer, context);
+        return new DefaultTrustedClaimIssuer(
+          { did: signerToString(identity), ticker, trustedFor },
+          context
+        );
```
monitz87, line 122: This shouldn't be necessary when the types are separated

monitz87 (changes requested):

src/types/index.ts
polymath-eric: Thanks, this looks a lot better know.

src/types/index.ts
```diff
@@ -337,13 +338,21 @@ export interface ClaimScope {
...
    * an undefined value means that the issuer is trusted for all claim types.
    */
   trustedFor?: ClaimType[];
 }
 
+export type InputTrustedClaimIssuer = Modify<
+  TrustedClaimIssuer,
+  {
+    identity: string | Identity;
+    trustedFor?: ClaimType[];
```
monitz87, line 352: This isn't necessary

`Modify<T, R>` creates a type which has the properties of both `T` and `R`, and in case of an overlap, `R` takes precedence. So `Modify<TrustedClaimIssuer, { identity: string | Identity }>` will result in 

```
{
    identity: string | Identity;
    trustedFor?: ClaimType[];
}
```

Without the need to explicitly add `trustedFor?: ClaimType[];`

src/types/index.ts
```diff
@@ -353,7 +362,10 @@ export enum ConditionType {
   IsIdentity = 'IsIdentity',
 }
 
-export type ConditionBase = { target: ConditionTarget; trustedClaimIssuers?: TrustedClaimIssuer[] };
+export type ConditionBase = {
+  target: ConditionTarget;
+  trustedClaimIssuers?: InputTrustedClaimIssuer[];
```
monitz87, line 367: We're facing the same problem here, since conditions are also returned to users. Leave it as `TrustedClaimIssuer` for now. I'm working on the `Condition` side

monitz87 (changes requested):

src/types/index.ts
polymath-eric: I switched it back for now, but I wouldn't want to merge this in like this since its not enough to close the ticket.

monitz87 (approved):

PR #611, opened 2021-11-30
Msdk 731 fix instruction reschedule docs

polymath-eric:


prashantasdeveloper (commented):
LGTM

src/api/entities/Instruction/index.ts
```diff
@@ -403,7 +403,8 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
   public withdraw: NoArgsProcedureMethod<Instruction>;
 
   /**
-   * Schedule a failed Instructi oto rwaa
+   * Reschedules a failed Instruction to be tried again
+   * @throws if the Instruction status is not Failed
```
monitz87, line 407:
```suggestion
   * Reschedules a failed Instruction to be tried again
   *
   * @throws if the Instruction status is not `InstructionStatus.Failed`
```

(Nit, but just in case we end up changing the underlying enum value)

monitz87 (changes requested):

VictorVicente (approved):

PR #612, opened 2021-12-01
docs: ✏️ Add JSDoc comment to CreateCheckpointScheduleParams

polymath-eric:
Adds documents explaining the meaning of null values when creating a
CheckpointSchedule

### DIFF
--- a/src/api/procedures/createCheckpointSchedule.ts
+++ b/src/api/procedures/createCheckpointSchedule.ts
@@ -18,8 +18,17 @@ import {
 import { filterEventRecords } from '~/utils/internal';
 
 export interface CreateCheckpointScheduleParams {
+  /**
+   * The date for when to begin creating snapshots. A null value indicates immediately
+   */
   start: Date | null;
+  /**
+   * The cadence in which to make Checkpoints. A null value indicates to create only one Checkpoint, regardless of repetitions specified
+   */
   period: CalendarPeriod | null;
+  /**
+   * The number of snapshots to take. A null value indicates snapshots should be made indefinitely
+   */
   repetitions: number | null;
 }
 
--- a/src/api/procedures/createCheckpointSchedule.ts
+++ b/src/api/procedures/createCheckpointSchedule.ts
@@ -19,11 +19,12 @@ import { filterEventRecords } from '~/utils/internal';
 
 export interface CreateCheckpointScheduleParams {
   /**
-   * The date for when to begin creating snapshots. A null value indicates immediately
+   * The date from which to begin creating snapshots. A null value indicates immediately
    */
   start: Date | null;
   /**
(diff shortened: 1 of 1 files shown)


src/api/procedures/createCheckpointSchedule.ts
```diff
@@ -18,8 +18,17 @@ import {
...
 
 export interface CreateCheckpointScheduleParams {
+  /**
+   * The date for when to begin creating snapshots. A null value indicates immediately
+   */
   start: Date | null;
+  /**
+   * The cadence in which to make Checkpoints. A null value indicates to create only one Checkpoint, regardless of repetitions specified
+   */
   period: CalendarPeriod | null;
```
polymath-eric, line 28: To me a null period doesn't make sense when there is an API to create a single Checkpoint.

monitz87: This allows you to schedule its creation in the future

monitz87: But yeah, null date + null period is the same as just creating a Checkpoint directly

src/api/procedures/createCheckpointSchedule.ts
```diff
@@ -18,8 +18,17 @@ import {
 import { filterEventRecords } from '~/utils/internal';
 
 export interface CreateCheckpointScheduleParams {
+  /**
+   * The date for when to begin creating snapshots. A null value indicates immediately
```
monitz87, line 22:
```suggestion
   * The date from which to begin creating snapshots. A null value indicates immediately
```

src/api/procedures/createCheckpointSchedule.ts
```diff
@@ -18,8 +18,17 @@ import {
...
+   */
   start: Date | null;
+  /**
+   * The cadence in which to make Checkpoints. A null value indicates to create only one Checkpoint, regardless of repetitions specified
```
monitz87, line 26: I think "cadence with which to make..." is the correct usage, but I'm actually not sure and I couldn't find it anywhere, so I leave it up to you

monitz87 (changes requested):

src/api/procedures/createCheckpointSchedule.ts
polymath-eric: Oh, I didn't think about that. I guess it does give new functionality then.

src/api/procedures/createCheckpointSchedule.ts
polymath-eric: I think you are right.

src/api/procedures/createCheckpointSchedule.ts
polymath-eric: I moved it to a @note, + specified that use case.

monitz87 (approved):

PR #613, opened 2021-12-03
feat: 🎸 MSDK-754 Allow non-entities as checkpoints while creating a DividendDistribution

prashantasdeveloper:
### JIRA Link

https://polymath.atlassian.net/browse/MSDK-754

### Change Description

This allows passing an object with `id` and `type` as checkpoint value while creating `DividendDistribution` along with already supported types `Checkpoint` , `CheckpointSchedule` and `Date`.
The `type` can be have either of two values -
* `Existing` - if `id` value defines a `Checkpoint`
* `Scheduled`- if `id` value defines a `CheckpointSchedule`

### DIFF
--- a/src/api/entities/SecurityToken/Checkpoints/types.ts
+++ b/src/api/entities/SecurityToken/Checkpoints/types.ts
@@ -1,0 +1,21 @@
+import { BigNumber } from 'bignumber.js';
+
+import { Checkpoint, CheckpointSchedule } from '~/types';
+
+export enum CaCheckpointType {
+  Existing = 'Existing',
+  Schedule = 'Schedule',
+}
+
+export type CaCheckpointTypeParams =
+  | Checkpoint
+  | CheckpointSchedule
+  | Date
+  | {
+      type: CaCheckpointType.Existing;
+      id: BigNumber;
+    }
+  | {
+      type: CaCheckpointType.Schedule;
+      id: BigNumber;
+    };
--- a/src/utils/internal.ts
+++ b/src/utils/internal.ts
@@ -15,7 +15,10 @@ import stringify from 'json-stable-stringify';
 import { chunk, differenceWith, groupBy, isEqual, map, mapValues, padEnd } from 'lodash';
 import { ModuleName, TxTag } from 'polymesh-types/types';
 
+import { assertCaCheckpointValid } from '~/api/procedures/utils';
 import {
+  Checkpoint,
+  CheckpointSchedule,
   Context,
   Identity,
   PolymeshError,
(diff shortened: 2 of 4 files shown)


src/utils/internal.ts
```diff
@@ -813,3 +818,34 @@ export function conditionsAreEqual(
...
+  checkpoint: CaCheckpointTypeParams,
+  token: string | SecurityToken,
+  context: Context
+): Promise<Checkpoint | CheckpointSchedule | Date> {
+  if (
+    checkpoint instanceof Checkpoint ||
+    checkpoint instanceof CheckpointSchedule ||
+    checkpoint instanceof Date
+  ) {
+    assertCaCheckpointValid(checkpoint);
```
polymath-eric, line 837: I am a little unsure about this assertion in a generic conversion function like this. It's asserting a date must be in the future, but I am not sure if that's going to be the case for everything that might call this method.

It might be reasonable that all references to checkpoints by dates will be in the future, so it might be OK to leave in. I think the exists check makes sense in that.

polymath-eric (commented):
LGTM, left a comment around the assert, but that might be alright. Let me know what you think.

src/utils/internal.ts
```diff
@@ -813,3 +818,34 @@ export function conditionsAreEqual(
 
   return equalClaims && equalClaimIssuers;
 }
+
+/**
+ * @hidden
+ *
+ * Transforms CaCheckpointTypeParams values to instance either Checkpoint or CheckpointSchedule
```
monitz87, line 825:
```suggestion
 * Transforms input CA Checkpoint values to `Checkpoint | CheckpointSchedule | Date` for easier processing
```

monitz87 (changes requested):

src/api/entities/SecurityToken/Checkpoints/types.ts
```diff
@@ -2,7 +2,7 @@ import { BigNumber } from 'bignumber.js';
 
 import { Checkpoint, CheckpointSchedule } from '~/types';
 
-export enum CaCheckpointType {
+export enum InputCaCheckpoint {
```
monitz87, line 5: My suggestion was to rename `CaCheckpointTypeParams`, not this enum

monitz87 (changes requested):

src/utils/internal.ts
monitz87: The idea was to remove the assertion, since the validation is already being done in the CA creation procedure. The `await` comment was just academic

monitz87 (changes requested):
While you're at it, please also change `InitiateCorporateActionParams` to accept non-entity values for the checkpoint (should still support undefined)

polymath-eric:
@prashantasdeveloper  

I pushed a commit addressing @monitz87  recent comments. I took out the assert in `getCheckpointValue`, updated the names, and changed `InitiateCorporateActionParams`

polymath-eric (approved):

monitz87 (approved):

PR #616, opened 2021-12-06
Msdk 743 update to 4.1.0 types

polymath-eric:


src/polkadot/definitions.ts
```diff
@@ -1,5 +1,3 @@
-/* istanbul ignore file */
-
```
VictorVicente, line 2: This file must be ignored, please restore this line

VictorVicente (changes requested):
Hey @polymath-eric, well done.
I think you need to fix the test case for the versioning control. That's the reason why all tests are failing.

monitz87 (approved):

VictorVicente (approved):

PR #626, opened 2021-12-14
fix: 🐛 Transfer Ownership method

polymath-eric:
Pads the ticker that is passed to the token.transferOwnership method to
prevent an incorrect length error from being thrown


monitz87 (approved):
LGTM, gj

PR #635, opened 2021-12-16
feat: 🎸 MSDK-756 Change `totalSupply` to `initialSupply` on asset creation

prashantasdeveloper:
BREAKING CHANGE: 🧨 `TickerReservation.createToken` now takes `initialSupply` instead of
`totalSupply` parameter for specifying amount of tokens that will be
minted on creation


polymath-eric (approved):
LGTM

I think this name makes it a lot more clear what the value is for.

monitz87 (approved):

PR #637, opened 2021-12-17
feat: 🎸 Add reserved to AccountBalance

polymath-eric:


monitz87 (changes requested):
`reserved` shouldn't really be exposed, since it doesn't have a use in the chain and as such it will just confuse users. If it ever becomes a thing, we'll add it. For now we're just adding it to the total to avoid a potential bug in a way that doesn't clutter the interface

monitz87 (approved):

prashantasdeveloper (commented):
LGTM

PR #638, opened 2021-12-17
refactor: 💡 Create interface for getGroups method

polymath-eric:


src/types/index.ts
```diff
@@ -939,6 +939,14 @@ export interface Permissions {
  */
 export type GroupPermissions = Pick<Permissions, 'transactions' | 'transactionGroups'>;
 
+/**
+ * This represents all Permission Groups of a Security Token
+ */
+export interface AllPermissionGroups {
```
monitz87, line 945: What about

```
/**
 * This represents all Permission Groups of a specific Security Token, separated by category
 */
export interface CategorizedPermissionGroups {
```

monitz87 (changes requested):
Minor naming suggestion. Other than that, all good, and good catch on the spelling mistakes

src/types/index.ts
polymath-eric: What about just `PermissionGroups`, it looks like that won't conflict with anything.

I think Categorized is better than All, but `known` and `custom` feel like pretty broad categories to me.

polymath-eric: Pushed a commit with Categorized, but I can change it to just `PermissionGroups` also. 

Let me know what you prefer.

monitz87: `PermissionGroups` is better I think

polymath-eric: Updated to PermissionGroups

monitz87 (approved):

PR #642, opened 2021-12-21
chore: 🤖 MSDK-752 Add `hasSameElements` for array comparison

prashantasdeveloper:
#### JIRA Link

https://polymath.atlassian.net/browse/MSDK-752

#### Changelog

This adds a common function `isSameSet` to check if two arrays are equal. 
It defaults to `isEqual` function of `lodash`, if no comparator is provided to the function.


src/utils/internal.ts
```diff
@@ -847,3 +844,11 @@ export async function getCheckpointValue(
     ).schedule;
   }
 }
+
+export function isSameSet<T>(
```
polymath-eric, line 848: I think this should this be named `isSameArray` or `arrayDeepEqual`. 

imo Using the term Set is misleading with Array arguments since JS has an actual Set type.

monitz87: The thing is that order matters in arrays, so that name would be misleading. We could call it `isSameCollection`

monitz87 (changes requested):
There's a linting error (missing docs for isSameSet)

src/utils/internal.ts
polymath-eric: Good point. I like both names Jere suggested.

polymath-eric (approved):

monitz87 (approved):

monitz87:
GJ, merging

PR #643, opened 2021-12-22
fix: 🐛 Correct `agIdSequence` -> `aGIdSequence`

prashantasdeveloper:


polymath-eric (approved):
The casing looks werid, but if this is the fix, then I guess its OK

monitz87 (approved):

PR #645, opened 2021-12-22
Msdk 684 validate authorizations

polymath-eric:


src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+        error = err;
+      }
+      const expectedError = new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'Subsidizer Account does not have a valid CDD Claim',
+      });
+      expect(error).toEqual(expectedError);
+    });
+
+    test('with a beneficiary that does not have an Identity', async () => {
```
monitz87, line 1180:
```suggestion
    test('should throw with a beneficiary that does not have an Identity', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      let error;
+      try {
+        await assertAuthorizationRequestValid(mockContext, auth);
+      } catch (err) {
+        error = err;
+      }
+      expect(error).toBe(undefined);
+    });
+
+    test('with a beneficiary that does not have a CDD Claim', async () => {
```
monitz87, line 1097:
```suggestion
    test('should throw with a beneficiary that does not have a CDD Claim', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      transactions: null,
+      transactionGroups: [],
+      portfolios: null,
+    };
+    const rawData = {
+      type: AuthorizationType.JoinIdentity as AuthorizationType.JoinIdentity,
+      value: permissions,
+    };
+    const data = rawData as Authorization;
+    test('with a valid request', async () => {
```
monitz87, line 1009:
```suggestion

    test('should not throw with a valid request', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      let error;
+      try {
+        await assertAuthorizationRequestValid(mockContext, auth);
+      } catch (err) {
+        error = err;
+      }
+      expect(error).toBe(undefined);
+    });
+
+    test('without a valid CDD', async () => {
```
monitz87, line 1031:
```suggestion
    test('should throw without a valid CDD', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      const subsidy = {
+        beneficiary,
+        subsidizer,
+        allowance,
+        remaining: allowance,
+      };
+      const data = {
+        type: AuthorizationType.AddRelayerPayingKey as AuthorizationType.AddRelayerPayingKey,
+        value: subsidy,
+      };
```
monitz87, line 1112: Same here

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+        error = err;
+      }
+      const expectedError = new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'Beneficiary Account does not have a valid CDD Claim',
+      });
+      expect(error).toEqual(expectedError);
+    });
+
+    test('with a Subsidizer that does not have a CDD Claim', async () => {
```
monitz87, line 1136:
```suggestion
    test('should throw with a Subsidizer that does not have a CDD Claim', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+        error = err;
+      }
+      const expectedError = new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'Beneficiary Account does not have an Identity',
+      });
+      expect(error).toEqual(expectedError);
+    });
+
+    test('with a Subsidizer that does not have an Identity', async () => {
```
monitz87, line 1219:
```suggestion
    test('should throw with a Subsidizer that does not have an Identity', async () => {
```

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      const beneficiary = getAccountInstance({
+        getIdentity: getIdentityInstance({ hasValidCdd: true }),
+      });
+      // getIdentityInstance modifies the prototype, which prevents two mocks from returning different values
```
monitz87, line 1223: Same here

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+      const subsidy = {
+        beneficiary,
+        subsidizer,
+        allowance,
+        remaining: allowance,
+      };
+      const data = {
+        type: AuthorizationType.AddRelayerPayingKey as AuthorizationType.AddRelayerPayingKey,
+        value: subsidy,
+      };
```
monitz87, line 1237: Same here

src/api/procedures/consumeAuthorizationRequests.ts
```diff
@@ -119,6 +125,8 @@ export async function getAuthorization(
     if (!accept) {
       did = await fetchDid();
       condition = condition || did === issuer.did;
+    } else {
+      await assertAuthorizationRequestValid(context, authRequest);
```
monitz87, line 129: These sorts of validations should go in the prepare method. The `getAuthorization` method is only for permissions and roles

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  if (authRequest.isExpired()) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'The Authorization Request has expired',
+      data: {
+        expiry: authRequest.expiry,
+      },
+    });
+  }
+  switch (authRequest.data.type) {
```
monitz87, line 350: Destructure data into a variable so you can reuse it below

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+      return assertJoinIdentityAuthorizationValid(authRequest);
+    case AuthorizationType.AddRelayerPayingKey:
+      return assertAddRelayerPayingKeyAuthorizationValid(authRequest.data);
+    default:
+      throw new UnreachableCaseError(authRequest.data); // ensures switch statement covers all values
+  }
+}
+
+/**
+ * asserts valid  primary key rotation authorization
```
monitz87, line 378:
```suggestion
 * @hidden
 *
 * Assert valid primary key rotation authorization
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  if (authRequest.target instanceof Identity) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'An Identity can not join another Identity',
+    });
+  }
+}
+
+/**
+ * asserts valid attest primary key authorization
```
monitz87, line 392:
```suggestion
 * @hidden
 *
 * Assert valid attest primary key authorization
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+): Promise<void> {
+  if (!(await authRequest.issuer.isCddProvider())) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Issuer must be a CDD provider',
+    });
+  }
+}
+
+/** asserts transfer ticker authorization is valid */
```
monitz87, line 405:
```suggestion
/** 
 * @hidden
 *
 * Assert transfer ticker authorization is valid
 */
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  }
+  if (status === TickerReservationStatus.TokenCreated) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'The ticker has already been used to create an Asset',
+    });
+  }
+}
+
+/** asserts valid transfer asset ownership authorization */
```
monitz87, line 426:
```suggestion
/**
 * @hidden
 *
 * Assert valid transfer asset ownership authorization
 */
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  const token = new SecurityToken({ ticker: data.value }, context);
+  if (!(await token.exists()))
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'The Asset does not exist',
+    });
+}
+
+/**
+ * asserts valid join identity authorization request
```
monitz87, line 440:
```suggestion
 * @hidden
 *
 * Assert valid join identity authorization request
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+ */
+export async function assertJoinIdentityAuthorizationValid(
```
monitz87, line 442: This one should also check that the target is not an identity and is not already part of one

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  authRequest: AuthorizationRequest
+): Promise<void> {
+  // https://github.com/PolymathNetwork/Polymesh/blob/5fec16fdfb05dc9022508880503e835ae9c1776c/pallets/identity/src/keys.rs#L470
+  const identity = authRequest.issuer;
+  if (!(await identity.hasValidCdd())) {
```
monitz87, line 447: We avoid inline awaits for better readability (there are quite a few in this PR)

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Issuing Identity does not have a valid CDD claim',
+    });
+  }
+}
+
+/** asserts valid add relayer paying key authorization */
```
monitz87, line 455:
```suggestion
/**
 * @hidden
 *
 * Assert valid add relayer paying key authorization 
 */
```

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  }
+  if (!(await subsidizerIdentity.hasValidCdd())) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Subsidizer Account does not have a valid CDD Claim',
+    });
+  }
+}
+
+/** A helper class to ensure a code path is unreachable. For example this can be used for ensuring switch statements are exhaustive */
```
monitz87, line 489:
```suggestion
/**
 * @hidden
 *
 * Helper class to ensure a code path is unreachable. For example, this can be used for ensuring switch statements are exhaustive 
 */
```

src/types/index.ts
```diff
@@ -1037,26 +1037,46 @@ export interface Subsidy {
...
+  value: Permissions;
+};
+export type PortfolioCustodyAuthorizationData = {
+  type: AuthorizationType.PortfolioCustody;
+  value: NumberedPortfolio | DefaultPortfolio;
+};
+export type BecomeAgentAuthorizationData = {
+  type: AuthorizationType.BecomeAgent;
+  value: KnownPermissionGroup | CustomPermissionGroup;
+};
```
monitz87, line 1052:
```suggestion
export type RotatePrimaryKeyAuthorizationData = { type: AuthorizationType.RotatePrimaryKey };

export type JoinIdentityAuthorizationData = {
  type: AuthorizationType.JoinIdentity;
  value: Permissions;
};

export type PortfolioCustodyAuthorizationData = {
  type: AuthorizationType.PortfolioCustody;
  value: NumberedPortfolio | DefaultPortfolio;
};

export type BecomeAgentAuthorizationData = {
  type: AuthorizationType.BecomeAgent;
  value: KnownPermissionGroup | CustomPermissionGroup;
};
```

monitz87: Just empty lines in between

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
     });
   }
 }
+
+/**
+ * @hidden
+ */
+export async function assertAuthorizationRequestValid(
```
monitz87, line 337: We should also validate (if we're not already, which I think is the case) that the authorization still exists by the time it is being accepted (it could have been removed or already accepted before)

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+/**
+ * asserts valid  primary key rotation authorization
+ */
+export async function assertPrimaryKeyRotationAuthorizationValid(
+  authRequest: AuthorizationRequest
+): Promise<void> {
+  if (authRequest.target instanceof Identity) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'An Identity can not join another Identity',
```
monitz87, line 386: An Identity can not become the primary key of another Identity

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  if (status === TickerReservationStatus.Free) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'The Ticker is not reserved',
+    });
+  }
+  if (status === TickerReservationStatus.TokenCreated) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'The ticker has already been used to create an Asset',
```
monitz87, line 421: Ticker should be uppercase here (we're not being very consistent accross the codebase though)

src/api/procedures/utils.ts
monitz87: Now that I think about it, a lot of authorizations are meant for either an identity or an account, but not both. We should probably have `assertTargetIsIdentity` and `assertTargetIsAccount`, and call them appropriately depending on the auth type

src/api/procedures/utils.ts
```diff
@@ -322,3 +330,166 @@ export function assertRequirementsNotTooComplex(
...
+  if (!(await identity.hasValidCdd())) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Issuing Identity does not have a valid CDD claim',
+    });
+  }
+}
+
+/** asserts valid add relayer paying key authorization */
+export async function assertAddRelayerPayingKeyAuthorizationValid(
```
monitz87, line 456: We should probably parallelize the calls where possible here (fetch both identities in parallel, check and throw errors if they don't exist, and then parallelize the cdd calls)

monitz87 (changes requested):
Mostly nitpicks with some substance in between. GJ overall, I liked the giant switch approach

src/api/procedures/__tests__/utils.ts
```diff
@@ -10,22 +11,40 @@ import {
...
+  PolymeshError,
+} from '~/internal';
 import { dsMockUtils, entityMockUtils } from '~/testUtils/mocks';
-import { getInstructionInstance } from '~/testUtils/mocks/entities';
+import {
+  getAccountInstance,
+  getIdentityInstance,
+  getInstructionInstance,
+  getNumberedPortfolioInstance,
+} from '~/testUtils/mocks/entities';
```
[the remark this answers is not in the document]
polymath-eric: Other places in this file were using them directly. I'll update those as well.

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
     ).not.toThrow();
   });
 });
+
+describe('authorization request validations', () => {
+  let mockContext: Context;
+  let target: Identity;
+  let issuer: Identity;
+  let expiry: Date;
```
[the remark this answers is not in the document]
polymath-eric: Good catch. I set trim trailing whitespace in VSCode now.

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+  afterEach(() => {
+    dsMockUtils.cleanup();
+  });
+
+  afterAll(() => {
+    dsMockUtils.cleanup();
+  });
+
+  describe('assertAuthorizationRequestValid', () => {
+    test('with an expired request', async () => {
```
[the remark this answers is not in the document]
polymath-eric: This is a nit on my part, but if we are writing the cases with "should", then why aren't we using the `it`?

I looked it up and `it` aliases `test`, and it makes the should read more naturally.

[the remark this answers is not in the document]
monitz87: Agreed, we could just replace all `test` with `it`. Can you create a task for it?

src/api/procedures/__tests__/utils.ts
```diff
@@ -627,3 +675,611 @@ describe('assertRequirementsNotTooComplex', () => {
...
+    test('with a Subsidizer that does not have a CDD Claim', async () => {
+      const beneficiary = getAccountInstance({
+        getIdentity: getIdentityInstance({ hasValidCdd: true }),
+      });
+      // getIdentityInstance modifies the prototype, which prevents two mocks from returning different values
+      const subsidizer = ({
+        getIdentity: () => {
+          return { hasValidCdd: () => false };
+        },
+      } as unknown) as Account;
```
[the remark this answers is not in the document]
monitz87: I think chaining methods in sinon has unexpected behavior in some cases. Can you try

```
const hasValidCddStub = entityMockUtils.getIdentityHasValidCddStub(); // remember not to import separately
hasValidCddStub.onFirstCall().resolves(true);
hasValidCddStub.onSecondCall().resolves(false);
// ... etc
```

If that doesn't work, try it together with specifying `getIdentity: entityMockUtils.getIdentityInstance()` in the args for `getAccountInstance`

If THAT doesn't work, just go back to your original approach

src/api/procedures/utils.ts
monitz87: What I mean is that after accepting or removing an auth, the user still has access to the authorization request object, so they might try to accept it once again

src/api/procedures/__tests__/utils.ts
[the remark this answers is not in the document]
polymath-eric: https://polymath.atlassian.net/browse/MSDK-790

src/api/procedures/consumeAuthorizationRequests.ts
```diff
@@ -58,15 +58,16 @@ export async function prepareConsumeAuthorizationRequests(
 
     const idsPerType: Record<AllowedAuthType, [u64][]> = mapValues(typesToExtrinsics, () => []);
 
-    liveRequests.forEach(authRequest => {
+    for (const authRequest of liveRequests) {
```
monitz87, line 61: You can keep the forEach, push the assertion promise into an array in each iteration, and then await the array with Promise.all outside the forEach

src/api/procedures/utils.ts
monitz87: Actually, as of version 3.7, you can do

```
function assertTargetIsIdentity(target: Signer): asserts target is Identity {
  // throw if not identity
}

// in function
assertTargetIsIdentity(target);

// if this code is reached, then the compiler knows `target` is an `Identity`
console.log(target.did);
```

More on assertion functions: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

monitz87 (changes requested):

src/api/procedures/utils.ts
polymath-eric: That makes sense I will add that in

src/api/procedures/consumeAuthorizationRequests.ts
polymath-eric: True, and then they will be in parallel if its a big batch.

src/api/procedures/utils.ts
polymath-eric: I didn't know that was possible. I made an assertIsSigner function with that.

src/api/procedures/__tests__/utils.ts
```diff
@@ -679,21 +679,34 @@ describe('authorization request validations', () => {
...
     dsMockUtils.cleanup();
+    entityMockUtils.cleanup();
   });
 
   describe('assertAuthorizationRequestValid', () => {
     test('should throw with an expired request', async () => {
       const oldDate = new Date('1987-01-01');
-      const data = { type: AuthorizationType.RotatePrimaryKey } as Authorization;
+      const data: Authorization = { type: AuthorizationType.RotatePrimaryKey };
       const auth = new AuthorizationRequest(
```
monitz87, line 710: You should be using an entity mock instead of a real instance in these tests

src/api/procedures/utils.ts
```diff
@@ -533,6 +536,20 @@ export async function assertAddRelayerPayingKeyAuthorizationValid(
   }
 }
 
+/**
+ * @hidden
+ *
+ * Asserts a target is a Account
+ */
+function assertIsSigner(target: Identity | Account): asserts target is Account {
```
monitz87, line 544: Shouldn't it be `assertIsAccount`? Also, you can use the `Signer` type instead of `Identity | Account`

src/api/procedures/utils.ts
```diff
@@ -533,6 +536,20 @@ export async function assertAddRelayerPayingKeyAuthorizationValid(
   }
 }
 
+/**
+ * @hidden
+ *
+ * Asserts a target is a Account
```
monitz87, line 542:
```suggestion
 * Assert a target is an Account
```

monitz87 (changes requested):
Nit: most procedure assertion utils have the context as the last argument, except for `assertRequirementsNotTooComplex` and the ones you introduce in this PR. Could you make it so that all of them have it as the last arg?

src/api/procedures/__tests__/utils.ts
```diff
@@ -22,6 +22,7 @@ import {
   PolymeshError,
 } from '~/internal';
 import { dsMockUtils, entityMockUtils } from '~/testUtils/mocks';
+import { getAuthorizationRequestInstance } from '~/testUtils/mocks/entities';
```
monitz87, line 25: Should use `entityMockUtils`

monitz87 (changes requested):
Just a nit

monitz87 (approved):

PR #650, opened 2021-12-23
Msdk 723 return authorizations

polymath-eric:


src/api/procedures/inviteExternalAgent.ts
```diff
@@ -88,6 +90,14 @@ export async function prepareInviteExternalAgent(
 
   const { ticker, target, permissions, expiry } = args;
 
+  const issuer = await context.getCurrentIdentity();
+  let targetIdentity: Identity;
+  if (typeof target === 'string') {
+    targetIdentity = new Identity({ did: target }, context);
+  } else {
+    targetIdentity = target;
+  }
```
monitz87, line 99: We should probably create something like `context.getIdentity(identity: Identity | string)` for this, since it's being repeated quite a bit accross the entire project. If the passed value is an identity, we just return it. If it's a DID, we instantiate an identity, check if it `exists()`, return if so, throw if not.

Then, the `getIdentity` method in  the Polymesh class could use that behind the scenes

monitz87 (changes requested):

src/api/procedures/inviteExternalAgent.ts
```diff
@@ -117,16 +119,27 @@ export async function prepareInviteExternalAgent(
...
       ticker,
       permissions,
     })) as PostTransactionValue<CustomPermissionGroup>;
+    postTransactionAuthorization = createGroupResult.transform(customPermissionGroup => {
+      return {
+        type: AuthorizationType.BecomeAgent,
+        value: customPermissionGroup,
+      };
+    });
     rawAuthorizationData = createGroupResult.transform(customPermissionGroup =>
```
monitz87, line 143: I was thinking more of something like "chaining" the transformations. So instead of having `authorizationDataResolver` that goes directly from permission to authorizationData, you split it into two steps. One that goes from permission to Authorization, and then just `authorizationToAuthorizationData`

You can do the same with transform, using `transform` on the result of the first `transform` call to get the second result

src/api/procedures/setCustodian.ts
```diff
@@ -78,27 +78,19 @@ export async function prepareSetCustodian(
...
-    {
-      type: AuthorizationType.PortfolioCustody as AuthorizationType.PortfolioCustody,
-      value: portfolio,
-    },
-    context
-  );
+  const authRequest = {
+    type: AuthorizationType.PortfolioCustody as AuthorizationType.PortfolioCustody,
+    value: portfolio,
+  };
```
monitz87, line 84:
```suggestion
  const authRequest: Authorization = {
    type: AuthorizationType.PortfolioCustody,
    value: portfolio,
  };
```

src/api/procedures/setCustodian.ts
```diff
@@ -78,27 +78,19 @@ export async function prepareSetCustodian(
...
+  const rawAuthorizationData = authorizationToAuthorizationData(authRequest, context);
 
   const rawExpiry = expiry ? dateToMoment(expiry, context) : null;
```
monitz87, line 87: While you're at it, could you change this to use `optionize`?

src/api/procedures/transferTickerOwnership.ts
```diff
@@ -73,26 +74,18 @@ export async function prepareTransferTickerOwnership(
...
-      value: ticker,
-    },
-    context
-  );
+  const authReq: Authorization = {
+    type: AuthorizationType.TransferTicker,
+    value: ticker,
+  };
+  const rawAuthorizationData = authorizationToAuthorizationData(authReq, context);
   const rawExpiry = expiry ? dateToMoment(expiry, context) : null;
```
monitz87, line 82: Same here with optionize

src/api/procedures/transferTokenOwnership.ts
```diff
@@ -55,26 +55,18 @@ export async function prepareTransferTokenOwnership(
...
-    {
-      type: AuthorizationType.TransferAssetOwnership as AuthorizationType.TransferAssetOwnership,
-      value: ticker,
-    },
-    context
-  );
+  const authRequest = {
+    type: AuthorizationType.TransferAssetOwnership as AuthorizationType.TransferAssetOwnership,
+    value: ticker,
+  };
```
monitz87, line 61:
```suggestion
  const authRequest: Authorization = {
    type: AuthorizationType.TransferAssetOwnership,
    value: ticker,
  };
```

src/api/procedures/inviteExternalAgent.ts
```diff
@@ -91,12 +98,7 @@ export async function prepareInviteExternalAgent(
   const { ticker, target, permissions, expiry } = args;
 
   const issuer = await context.getCurrentIdentity();
-  let targetIdentity: Identity;
-  if (typeof target === 'string') {
-    targetIdentity = new Identity({ did: target }, context);
-  } else {
-    targetIdentity = target;
-  }
+  const targetIdentity = await context.getIdentity(target);
```
monitz87, line 101: Aren't there more places in this PR where this can be used?

src/base/Context.ts
```diff
@@ -523,6 +523,30 @@ export class Context {
...
+    if (typeof identity === 'string') {
+      id = new Identity({ did: identity }, this);
+    } else {
+      id = identity;
+    }
+    const exists = await id.exists();
+
+    if (!exists) {
+      throw new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
```
monitz87, line 542: I'd say this should be a `DataUnavailable` error, although I can understand the impulse of using `UnmetPrerequisite`

src/base/Context.ts
```diff
@@ -523,6 +523,30 @@ export class Context {
     return invalidDids;
   }
 
+  /**
+   * Returns an Identity when given a did string
+   *
+   * @throws if the Identity does not exist
+   */
+  public async getIdentity(identity: Identity | string): Promise<Identity> {
```
monitz87, line 531: This should also be used in the `getIdentity` method in the root class

src/base/Context.ts
```diff
@@ -523,6 +523,30 @@ export class Context {
...
+    let id;
+    if (typeof identity === 'string') {
+      id = new Identity({ did: identity }, this);
+    } else {
+      id = identity;
+    }
```
monitz87, line 537: Here we should just return. If the user has an Identity instance, it's because it was already checked for existence somewhere else, or directly returned from the chain. Since Identities cannot be removed, the check would be redundant

monitz87 (changes requested):

src/api/procedures/__tests__/inviteExternalAgent.ts
```diff
@@ -155,19 +163,29 @@ describe('inviteExternalAgent procedure', () => {
...
+            {
+              agent: { did: 'otherDid' } as Identity,
+              group: entityMockUtils.getCustomPermissionGroupInstance(),
+            },
+          ],
+        }),
+      }
+    );
 
+    type AuthCallback = () => AuthorizationData;
```
polymath-eric, line 180: used this to try to clean up the call sig a bit. I needed the transform from transform for the chaining case.

src/base/Context.ts
polymath-eric: Update that to call this method

src/__tests__/Polymesh.ts
```diff
@@ -673,22 +673,13 @@ describe('Polymesh Class', () => {
...
 
-      const result = await polymesh.getIdentity(params);
       const context = dsMockUtils.getContextInstance();
+      const identity = new Identity(params, context);
+      context.getIdentity.onFirstCall().resolves(identity);
 
-      expect(result).toMatchObject(new Identity(params, context));
-    });
-
-    test('should throw an error if the Identity does not exist', async () => {
```
polymath-eric, line 682: Took this case since its covered in the real implementation which is now called.

src/api/procedures/inviteExternalAgent.ts
polymath-eric: I see. That does clean up the file a bit.

The mock in the test case is a bit more complicated, but I think its worth it.

src/api/procedures/inviteExternalAgent.ts
```diff
@@ -127,7 +111,7 @@ export async function prepareInviteExternalAgent(
       type: AuthorizationType.BecomeAgent,
       value: permissions,
     };
-    rawAuthorizationData = authorizationDataResolver(permissions, context);
+    rawAuthorizationData = authorizationToAuthorizationData(postTransactionAuthorization, context);
```
monitz87, line 114: This is almost perfect, but we're still repeating most of

```
{
  type: AuthorizationType.BecomeAgent,
  value: permissions,
};
```

My idea was having something like 

```
const createBecomeAgentData = (value: KnownPermissionGroup | CustomPermissionGroup) => ({
  type: AuthorizationType.BecomeAgent,
  value,
});
```

And then use that in both branches

monitz87 (changes requested):
Just a minor detail

src/api/procedures/__tests__/utils.ts
```diff
@@ -672,13 +671,13 @@ describe('createAuthorizationResolver', () => {
     expect(authRequest.authId).toEqual(new BigNumber(3));
   });
 
-  test('it returns a function that creates an AuthorizationRequest with a PostTransaction Authorization', async () => {
+  test('should return a function that creates an AuthorizationRequest with a PostTransaction Authorization', async () => {
     const mockContext = dsMockUtils.getContextInstance();
 
     const authData = {
```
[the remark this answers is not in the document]
monitz87: Yeah, the `isRotatePrimaryKey` threw me off

src/api/procedures/inviteExternalAgent.ts
polymath-eric: Makes sense, I made the change.

monitz87 (approved):

PR #651, opened 2021-12-28
feat: 🎸MSDK-525 Procedure to `modify` a Compliance Requirement

prashantasdeveloper:
#### JIRA Link

https://polymath.atlassian.net/browse/MSDK-525

#### Changelog

Add a `modify` procedure to Requirements that allows to modify conditions of an existing Compliance Requirement.


src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,106 @@
...
+  ticker: string;
+};
+
+/**
+ * @hidden
+ */
+export async function prepareModifyComplianceRequirement(
+  this: Procedure<Params, SecurityToken>,
+  args: Params
+): Promise<SecurityToken> {
```
polymath-eric, line 34: I am not sure if SecurityToken is the right return value.

Stuff like `modifyCorporateActionsAgent` returns `Promise<void>`, while `modifyInstruction` returns an Instruction since that was what was being changed.

I am thinking we should either return ComplianceRequirements or void here.

(createCheckpoint has an example of using a resolver to get at post transaction values)

prashantasdeveloper: `void` makes more sense to me. Returning the same data is of no use/ is just redundant IMO.

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,105 @@
...
+import { requirementToComplianceRequirement, stringToTicker } from '~/utils/conversion';
+import { conditionsAreEqual, hasSameElements } from '~/utils/internal';
+
+export type ModifyComplianceRequirementParams = {
+  /**
+   * ID of the Compliance Requirement
+   */
+  id: number;
+  /**
+   * array of conditions to replace the existing array of conditions for the requirement(identified by `id`).
```
monitz87, line 14:
```suggestion
   * array of conditions to replace the existing array of conditions for the requirement (identified by `id`).
```

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,105 @@
...
+  const { conditions: existingConditions } = existingRequirement;
+
+  if (hasSameElements(conditions, existingConditions, conditionsAreEqual)) {
+    throw new PolymeshError({
+      code: ErrorCode.NoDataChange,
+      message: 'The supplied condition list is equal to the current one',
+    });
+  }
+
+  assertRequirementsNotTooComplex(context, conditions, defaultTrustedClaimIssuers.length);
```
monitz87, line 72: Here you should pass the resulting conditions. That means the existing conditions - the one being replaced + the new condition

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * @hidden
+ */
+export function getAuthorization(
+  this: Procedure<Params, void>,
+  { ticker }: Params
+): ProcedureAuthorization {
+  return {
+    roles: [{ type: RoleType.TickerOwner, ticker }],
```
monitz87, line 92: This role should not be a requirement for this procedure

src/api/procedures/__tests__/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,189 @@
...
+      conditions: fakeConditions,
+    };
+    const proc = procedureMockUtils.getInstance<Params, void>(mockContext);
+
+    return expect(prepareModifyComplianceRequirement.call(proc, args)).rejects.toThrow(
+      'The Compliance Requirement does not exist'
+    );
+  });
+
+  test('should throw an error if the supplied requirement conditions has no change', () => {
```
monitz87, line 122:
```suggestion
  test('should throw an error if the supplied requirement conditions have no change', () => {
```

src/api/procedures/__tests__/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,189 @@
...
+      conditions,
+    };
+    const proc = procedureMockUtils.getInstance<Params, void>(mockContext);
+
+    return expect(prepareModifyComplianceRequirement.call(proc, args)).rejects.toThrow(
+      'The supplied condition list is equal to the current one'
+    );
+  });
+
+  test('should add an modify compliance requirement transaction to the queue', async () => {
```
monitz87, line 135:
```suggestion
  test('should add a modify compliance requirement transaction to the queue', async () => {
```

monitz87 (changes requested):

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const unchangedRequirements = currentRequirements.filter(
+    ({ id: currentRequirementId }) => id !== currentRequirementId
+  );
+
+  const unchangedConditions = map(unchangedRequirements, 'conditions');
+
+  assertRequirementsNotTooComplex(
+    context,
+    [...flatten(unchangedConditions), ...newConditions],
```
polymath-eric, line 82: On line 78, could that just be `flatMap` instead?

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -62,16 +64,29 @@ export async function prepareModifyComplianceRequirement(
...
-  if (hasSameElements(conditions, existingConditions, conditionsAreEqual)) {
+  if (hasSameElements(newConditions, existingConditions, conditionsAreEqual)) {
     throw new PolymeshError({
       code: ErrorCode.NoDataChange,
       message: 'The supplied condition list is equal to the current one',
     });
   }
 
-  assertRequirementsNotTooComplex(context, conditions, defaultTrustedClaimIssuers.length);
+  const unchangedRequirements = currentRequirements.filter(
```
monitz87, line 74: Instead of going through the array twice, you can use lodash `remove` to get `existingRequirement` and then `currentRequirements` will only contain the unchanged ones

monitz87 (changes requested):

monitz87 (approved):

PR #653, opened 2022-01-03
feat: 🎸 Make `currentFundingRound` return `null` when empty

prashantasdeveloper:
BREAKING CHANGE: 🧨 `currentFundingRound` now returns null for empty funding round, thus
changing the return type `Promise<string | null>`

### DIFF
--- a/src/api/entities/SecurityToken/index.ts
+++ b/src/api/entities/SecurityToken/index.ts
@@ -5,6 +5,7 @@ import {
   AgentGroup,
   AssetName,
   Counter,
+  FundingRoundName,
   IdentityId,
   SecurityToken as MeshSecurityToken,
   Ticker,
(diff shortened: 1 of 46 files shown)


polymath-eric (approved):

src/utils/conversion.ts
```diff
@@ -1601,8 +1601,8 @@ export function stringToFundingRoundName(roundName: string, context: Context): F
 /**
  * @hidden
  */
-export function fundingRoundNameToString(roundName: FundingRoundName): string {
-  return roundName.toString();
+export function fundingRoundNameToString(roundName: FundingRoundName): string | null {
```
monitz87, line 1604: I'd rather you perform the isEmpty check in `currentFoundingRound` and leave this as it was. It'll be easier to adapt when core changes the return type to `Option`

monitz87 (changes requested):

src/api/entities/SecurityToken/index.ts
```diff
@@ -320,12 +320,12 @@ export class SecurityToken extends Entity<UniqueIdentifiers, string> {
...
     if (callback) {
       return asset.fundingRound(rawTicker, round => {
-        callback(fundingRoundNameToString(round));
+        callback(fundingRoundNameToString(round) || null);
       });
     }
 
     const fundingRound = await asset.fundingRound(rawTicker);
-    return fundingRoundNameToString(fundingRound);
+    return fundingRoundNameToString(fundingRound) || null;
```
monitz87, line 329: This behavior should probably be encapsulated into a function

```
const assembleResult = (roundName: FundingRoundName): string | null => fundingRoundNameToString(roundName) || null;

if (callback) {
  return asset.fundingRound(rawTicker, round => {
    callback(assembleResult(round));
  });
}

const fundingRound = await asset.fundingRound(rawTicker);
return assembleResult(fundingRound);
```

Easier to change when we move to Options, less prone to bugs

monitz87 (changes requested):

src/api/entities/SecurityToken/index.ts
prashantasdeveloper: I infact did that and then reverted it 😅

monitz87 (approved):

PR #659, opened 2022-01-05
feat: 🎸 Add identities namespace

polymath-eric:
Creates an identities namespace and moves methods into it

- sdk.registerIdentity -> sdk.identities.registerIdentity
- sdk.getIdentity -> sdk.identities.getIdentitiy
- sdk.isIdentityValid -> sdk.identities.isIdentityValid

BREAKING CHANGE: 🧨 Moves Identity methods


prashantasdeveloper (commented):
LGTM

polymath-eric:
@monitz87 Prashant gave an approval already for this one.

prashantasdeveloper (approved):

src/Identities.ts
```diff
@@ -0,0 +1,54 @@
+import { Context, Identity, registerIdentity, RegisterIdentityParams } from '~/internal';
+import { ProcedureMethod } from '~/types';
+import { createProcedureMethod } from '~/utils/internal';
+
+/**
+ * Handles all Identity related functionality
+ */
+export class Identities {
```
monitz87, line 8: Shouldn't `getCurrentIdentity` be in this namespace as well?

monitz87 (changes requested):

src/Identities.ts
polymath-eric: To me the idea of "currentIdentity" has to do more with SDK configuration than with interacting with Identities. 

In a sense the corresponding "setter" is the `setSigner` method and I think they should be kept together. The usage of the "currentIdentity" is different than all of the other methods here imo.

That said I seem to be in the minority here so I can move it if you still think it belong in the namespace.

monitz87: You make a very good point. Let's leave it as is for now

monitz87 (approved):

PR #662, opened 2022-01-08
Feat/msdk 586 batch different transactions

monitz87:


src/base/PolymeshTransactionBase.ts
```diff
@@ -383,13 +348,43 @@ export abstract class PolymeshTransactionBase<
     /* eslint-enable default-case */
   }
 
+  /**
+   * Return whether the transaction can be subsidized. If the result is false
+   *   AND the caller is being subsidized by a third party, the transaction can't be executed and trying
+   *   to do so will result in an error
```
polymath-eric, line 354: I thought the caller would still be able to make the transaction, but would have to cover the cost themselves.

Does this mean a subsidized account can do less things, even if they have the funds to pay for them?

monitz87: That's right. They can unilaterally exit the subsidizer relationship though if they need to run transactions that can't be subsidized

I proposed changes to core to address this in a cleaner way (https://polymathnetwork.slack.com/archives/C01RLA867C5/p1640281223354100?thread_ts=1640121955.343700&cid=C01RLA867C5), but there wasn't much interest

In any case, we're a long way from this becoming a real problem (if it ever does), and there's already a workaround

polymath-eric: Thanks, I wanted to be sure I was understanding this correctly.

Making a second key isn't too bad of a work around.

src/api/procedures/__tests__/reserveTicker.ts
```diff
@@ -174,29 +174,25 @@ describe('reserveTicker procedure', () => {
...
-      rawTicker
+        args: [rawTicker],
+      })
     );
     expect(result).toBe(reservation);
 
     entityMockUtils.getTickerReservationDetailsStub().resolves({
       owner: entityMockUtils.getIdentityInstance(),
-      expriy: new Date(3000, 12, 12),
+      expiry: new Date(3000, 12, 12),
```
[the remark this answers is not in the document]
prashantasdeveloper:
```suggestion
      expiryDate: new Date(3000, 12, 12),
```

src/base/TransactionQueue.ts
```diff
@@ -423,9 +453,23 @@ export class TransactionQueue<
    *   are sufficient to cover this queue's fees
    */
   private async assertFeesCovered(): Promise<void> {
-    const breakdown = await this.getMinFees();
+    const breakdown = await this._getMinFees();
+
+    const { accountBalance, accountFees, thirdPartyFees, subsidizedTransactions } = breakdown;
 
-    const { accountBalance, accountFees, thirdPartyFees } = breakdown;
+    const unsupportedTransactions = subsidizedTransactions.filter(
```
prashantasdeveloper, line 460: Since there is no further need of unsupportedTransactions, IMO we should use `.some` instead of `.filter`. In that case we can directly check the boolean value instead of checking the length

prashantasdeveloper (changes requested):

src/base/TransactionQueue.ts
monitz87: It's being used as part of the error data

src/api/procedures/__tests__/reserveTicker.ts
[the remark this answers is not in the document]
monitz87: Typescript doesn't really support "exact types". We could avoid using Partial, but it's useful sometimes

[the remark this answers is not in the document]
polymath-eric: The partial is define for the argument in `getTickerReservationDetailsStub`, but this is being passed into `resolves` which is `any`, so I am not sure if we can do anything here.

I thought even with partial TS would complain about recognized property in this case.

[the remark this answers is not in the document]
polymath-eric: I looked at this a bit more, and it looks like we could parameterize the SinonStubs, right now they are
`SinonStub`, but looks like you can say SinonStub<Args, Return>.

VictorVicente (approved):

PR #663, opened 2022-01-10
refactor: 💡 MSDK-667 Change portfolio name uniqueness validations

prashantasdeveloper:


src/utils/conversion.ts
```diff
@@ -3362,3 +3364,32 @@ export function agentGroupToPermissionGroup(
...
 }
+
+/**
+ * @hidden
+ */
+export async function portfolioNameToNumber(
+  rawIdentityId: IdentityId,
+  rawName: PortfolioName,
+  query: QueryableStorage<'promise'>
+): Promise<BigNumber | undefined> {
```
polymath-eric, line 3375: I think this should be null:
```suggestion
): Promise<BigNumber | null> {
```

In JS `undefined` means not initialized, where as `null` communicates there wasn't actually a value.

I think the behavior is pretty similar though.

(test case + return would need to be updated as well)

src/utils/conversion.ts
```diff
@@ -3362,3 +3364,32 @@ export function agentGroupToPermissionGroup(
...
   }
 }
+
+/**
+ * @hidden
+ */
+export async function portfolioNameToNumber(
+  rawIdentityId: IdentityId,
+  rawName: PortfolioName,
+  query: QueryableStorage<'promise'>
```
polymath-eric, line 3374: I'd prefer passing in the Context so the signature looks more consistent and so a caller doesn't need to know about the destructing stuff.

src/api/procedures/renamePortfolio.ts
```diff
@@ -34,35 +39,26 @@ export async function prepareRenamePortfolio(
...
+  if (existingPortfolioNumber) {
+    if (id.eq(existingPortfolioNumber)) {
+      throw new PolymeshError({
+        code: ErrorCode.NoDataChange,
+        message: 'New name is the same as current name',
+      });
+    } else {
+      throw new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'A portfolio with that name already exists',
```
polymath-eric, line 56: We aren't consistent, but I think we want entities to be capitalized like the rest API, so:

```suggestion
        message: 'A Portfolio with that name already exists',
```

src/api/procedures/__tests__/createPortfolio.ts
```diff
@@ -75,15 +58,17 @@ describe('createPortfolio procedure', () => {
...
-    ).rejects.toThrow('A portfolio with that name already exists');
+    return expect(prepareCreatePortfolio.call(proc, { name: newPortfolioName })).rejects.toThrow(
+      'A portfolio with that name already exists'
+    );
   });
 
   test('should add a create portfolio transaction to the queue', async () => {
     const proc = procedureMockUtils.getInstance<Params, NumberedPortfolio>(mockContext);
     const createPortfolioTransaction = dsMockUtils.createTxStub('portfolio', 'createPortfolio');
+    portfolioNameToNumberStub.returns(undefined);
```
polymath-eric, line 71:
```suggestion
    portfolioNameToNumberStub.returns(null);
```

(see other comment)

polymath-eric (changes requested):
overall looks good, left some small comments

src/utils/conversion.ts
```diff
@@ -3362,3 +3363,36 @@ export function agentGroupToPermissionGroup(
...
+  } = context;
+
+  const rawPortfolioNumber = await portfolio.nameToNumber(rawIdentityId, rawName);
+
+  const portfolioId = u64ToBigNumber(rawPortfolioNumber);
+
+  // TODO @prashantasdeveloper remove this logic once nameToNumber returns Option<PortfolioNumber>
+  if (portfolioId.eq(new BigNumber(1))) {
+    /**
+     * since nameToNumber returns for non-existing portfolios,
```
polymath-eric, line 3388:
```suggestion
     * since nameToNumber returns 1 for non-existing portfolios,
```

nit, but I think it makes a difference in the readability.

polymath-eric (approved):
LGTM. (left a nit)

@monitz87 ready for you to take a pass.

src/utils/conversion.ts
```diff
@@ -3362,3 +3363,36 @@ export function agentGroupToPermissionGroup(
     }
   }
 }
+
+/**
+ * @hidden
+ */
+export async function portfolioNameToNumber(
```
monitz87, line 3370: Feels weird to have this as a conversion util, since it's not actually converting data, but rather fetching. I'd move it to `src/utils/internal.ts` and rename it to `getPortfolioIdByName`

monitz87 (changes requested):

monitz87 (approved):

PR #665, opened 2022-01-10
feat: 🎸 Return type change for `getPrimaryKey`

prashantasdeveloper:
JIRA Link - https://polymath.atlassian.net/browse/MSDK-760

BREAKING CHANGE: 🧨
* `SecondaryKey` interface has been renamed to `SigningKey`
* `identity.getPrimaryKey` now returns interface of type `SigningKey` instead of `Account`

### DIFF
--- a/src/api/entities/Account.ts
+++ b/src/api/entities/Account.ts
@@ -304,9 +304,9 @@ export class Account extends Entity<UniqueIdentifiers, string> {
       return false;
     }
 
-    const primaryKey = await identity.getPrimaryKey();
+    const { signer } = await identity.getPrimaryKey();
 
-    if (address === primaryKey.address) {
+    if (address === signerToString(signer)) {
       return false;
     }
 
@@ -321,12 +321,12 @@ export class Account extends Entity<UniqueIdentifiers, string> {
 
     const currentIdentity = await context.getCurrentIdentity();
 
-    const [primaryKey, secondaryKeys] = await Promise.all([
+    const [{ signer: primaryKeySigner }, secondaryKeys] = await Promise.all([
       currentIdentity.getPrimaryKey(),
       currentIdentity.getSecondaryKeys(),
     ]);
 
-    if (address === primaryKey.address) {
+    if (address === signerToString(primaryKeySigner)) {
       return {
         tokens: null,
         transactions: null,
--- a/src/api/entities/Identity/index.ts
+++ b/src/api/entities/Identity/index.ts
@@ -6,7 +6,6 @@ import { CddStatus, DidRecord } from 'polymesh-types/types';
 
 import { assertPortfolioExists } from '~/api/procedures/utils';
 import {
-  Account,
   Context,
   Entity,
   Instruction,
@@ -30,7 +29,8 @@ import {
   Order,
   ResultSet,
   Role,
-  SecondaryKey,
+  SignerType,
+  SigningKey,
   SubCallback,
   UnsubCallback,
 } from '~/types';
@@ -52,6 +52,7 @@ import {
   signerValueToSigner,
   stringToIdentityId,
   stringToTicker,
+  transactionPermissionsToTxGroups,
   u64ToBigNumber,
 } from '~/utils/conversion';
 import { calculateNextKey, getTicker, removePadding } from '~/utils/internal';
@@ -256,11 +257,13 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
    *
    * @note can be subscribed to
    */
-  public async getPrimaryKey(): Promise<Account>;
-  public async getPrimaryKey(callback: SubCallback<Account>): Promise<UnsubCallback>;
+  public async getPrimaryKey(): Promise<SigningKey>;
+  public async getPrimaryKey(callback: SubCallback<SigningKey>): Promise<UnsubCallback>;
 
   // eslint-disable-next-line require-jsdoc
-  public async getPrimaryKey(callback?: SubCallback<Account>): Promise<Account | UnsubCallback> {
+  public async getPrimaryKey(
+    callback?: SubCallback<SigningKey>
+  ): Promise<SigningKey | UnsubCallback> {
     const {
       context: {
         polymeshApi: {
@@ -271,8 +274,19 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
       context,
     } = this;
 
-    const assembleResult = ({ primary_key: primaryKey }: DidRecord): Account => {
-      return new Account({ address: accountIdToString(primaryKey) }, context);
+    const assembleResult = ({ primary_key: primaryKey }: DidRecord): SigningKey => {
+      return {
+        signer: signerValueToSigner(
+          { type: SignerType.Account, value: accountIdToString(primaryKey) },
+          context
+        ),
+        permissions: {
+          tokens: null,
+          portfolios: null,
+          transactions: null,
+          transactionGroups: transactionPermissionsToTxGroups(null),
+        },
+      };
     };
 
     const rawDid = stringToIdentityId(did, context);
@@ -650,13 +664,13 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
    *
    * @note can be subscribed to
    */
-  public async getSecondaryKeys(): Promise<SecondaryKey[]>;
-  public async getSecondaryKeys(callback: SubCallback<SecondaryKey[]>): Promise<UnsubCallback>;
+  public async getSecondaryKeys(): Promise<SigningKey[]>;
+  public async getSecondaryKeys(callback: SubCallback<SigningKey[]>): Promise<UnsubCallback>;
 
   // eslint-disable-next-line require-jsdoc
   public async getSecondaryKeys(
-    callback?: SubCallback<SecondaryKey[]>
-  ): Promise<SecondaryKey[] | UnsubCallback> {
+    callback?: SubCallback<SigningKey[]>
+  ): Promise<SigningKey[] | UnsubCallback> {
     const {
       did,
       context,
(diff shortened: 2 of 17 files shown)


src/types/index.ts
```diff
@@ -1159,7 +1159,7 @@ export interface CheckpointWithData {
   totalSupply: BigNumber;
 }
 
-export interface SecondaryKey {
+export interface SigningKey {
   signer: Signer;
```
polymath-eric, line 1163: To me I would have though `Account` type should be sufficient here since I don't think we'd get an `Identity` from the related methods, and `Account` should be accepted by methods wanting the `Signer` since its a part of the union type of `Signer`.

I tried this change and it breaks some tests and is going to be more work. I am going to ask about it in Slack though.

polymath-eric (approved):
Left a comment about potentially narrowing the signer type to just `Account`. Seems like that will be a bigger change so out of scope for this PR imo.

Also as an FYI / heads up, with my Account Management namespace I'll be naming all of the `Key` -> `Account`.

src/types/index.ts
polymath-eric: I looked at the card and it said this should become an `Account`

https://polymath.atlassian.net/browse/MSDK-760

So this is actually in scope for this PR.

polymath-eric (changes requested):
Update the signer from `Signer` -> `Account` + fix the test cases that will break please.

src/api/entities/Identity/index.ts
```diff
@@ -667,9 +681,18 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
 
-    const assembleResult = ({ secondary_keys: secondaryKeys }: DidRecord): SecondaryKey[] => {
+    const assembleResult = ({
+      secondary_keys: secondaryKeys,
+    }: DidRecord): PermissionedAccount[] => {
       return secondaryKeys.map(({ signer: rawSigner, permissions }) => ({
-        signer: signerValueToSigner(signatoryToSignerValue(rawSigner), context),
+        account: new Account(
+          {
+            address: signerToString(
```
polymath-eric, line 690: This looks like a lot of nested helper function calls. I think a new function `signerToAddress` could make this nicer.

src/api/procedures/__tests__/removeSecondaryKeys.ts
```diff
@@ -31,9 +33,24 @@ describe('removeSecondaryKeys procedure', () => {
     addTransactionStub = procedureMockUtils.getAddTransactionStub();
     mockContext = dsMockUtils.getContextInstance();
 
-    const signers = [entityMockUtils.getAccountInstance({ address: 'someFakeAccount' })];
+    primaryKeyAccount = entityMockUtils.getAccountInstance({ address: 'primaryKey' });
+
+    console.log('here');
```
polymath-eric, line 38: remove

src/api/procedures/removeSecondaryKeys.ts
```diff
@@ -23,17 +23,20 @@ export async function prepareRemoveSecondaryKeys(
...
-  const { signers } = args;
+  const { accounts } = args;
 
   const identity = await context.getCurrentIdentity();
 
-  const [primaryKey, secondaryKeys] = await Promise.all([
-    identity.getPrimaryKey(),
-    identity.getSecondaryKeys(),
-  ]);
+  console.log(await identity.getPrimaryKey());
```
polymath-eric, line 30: stray console

polymath-eric (commented):
Overall LGTM, few stray console log, and I left a comment about reducing the nested signer helper functions.

src/testUtils/mocks/dataSources.ts
```diff
@@ -584,7 +584,17 @@ const defaultContextOptions: ContextOptions = {
     next: 1,
     count: 1,
   },
-  primaryKey: 'primaryKey',
+  primaryKey: {
+    account: ({
+      address: 'primaryKey',
+    } as unknown) as Account,
```
polymath-eric, line 590: I think it would be nice to not have this type assertion here.

Not sure how feasible it is though.

src/api/entities/Identity/index.ts
monitz87: Actually, since Identities can no longer be secondary keys, we can assume the rawSigner here always represents an Account

This means we can do one of the following:

- Use a type assertion `account: signerValueToSigner(signatoryToSignerValue(rawSigner), context) as Account`
- Create a `signatoryToAccount` conversion util and use it here

In both cases we should add a comment explaining that even though the chain types support identities as secondary accounts, in practice it doesn't occur and it will be removed

```
/**
 * @hidden
 */
export function signatoryToAccount(signatory: Signatory, context: Context): Account {
  if (signatory.isAccount) {
    return new Account({ address: accountIdToString(signatory.asAccount) }, context);
  }

  throw new PolymeshError({ code: ErrorCode.UnexpectedError, message: 'Received an Identity where an Account was expected. Please report this issue to the Polymath team' });
}
```

I suggest the second approach. More explicit and robust

prashantasdeveloper: Added the new util method

src/testUtils/mocks/dataSources.ts
prashantasdeveloper: Such type assertions are used at other places as well, hence I used the approach here. It works and doesn't impact anything so seems OK to me.

polymath-eric: True, OK to leave.

src/api/entities/Account.ts
```diff
@@ -322,12 +323,17 @@ export class Account extends Entity<UniqueIdentifiers, string> {
 
     const currentIdentity = await context.getCurrentIdentity();
 
-    const [primaryAccount, secondaryAccount] = await Promise.all([
+    const [
+      {
+        account: { address: primaryAccountAddress },
+      },
+      secondaryKeys,
```
polymath-eric, line 330:
```suggestion
      secondaryAccounts,
```

src/api/entities/Account.ts
```diff
@@ -336,8 +342,9 @@ export class Account extends Entity<UniqueIdentifiers, string> {
       };
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const account = secondaryAccount.find(({ signer }) => address === signerToString(signer))!;
+    const account = secondaryKeys.find(
+      ({ account: { address: secondaryKeyAddress } }) => address === secondaryKeyAddress
```
polymath-eric, line 346:
```suggestion
      ({ account: { address: secondaryAccountAddress } }) => address === secondaryAccountAddress
```

src/api/procedures/modifySignerPermissions.ts
```diff
@@ -36,29 +36,23 @@ export async function prepareModifySignerPermissions(
...
   const identity = await context.getCurrentIdentity();
 
-  const secondaryAccounts = await identity.getSecondaryAccounts();
-  const signerValues = signers.map(({ signer, permissions }) => {
-    return {
-      signer: signerToSignerValue(signer),
-      permissions,
-    };
-  });
+  const existingSecondaryKeys = await identity.getSecondaryAccounts();
```
polymath-eric, line 43:
```suggestion
  const existingSecondaryAccounts = await identity.getSecondaryAccounts();
```

polymath-eric (commented):
Left two nitty account vs key in variable name comments.

Basically we want to use account as much as possible, except in the polkadot layer thats communicating directly with the chain.

polymath-eric (approved):
LGTM

Should be good for another look from @monitz87

src/api/procedures/__tests__/removeSecondaryAccounts.ts
```diff
@@ -85,36 +101,35 @@ describe('removeSecondaryAccounts procedure', () => {
...
     const proc = procedureMockUtils.getInstance<RemoveSecondaryAccountsParams, void>(mockContext);
 
     return expect(
       prepareRemoveSecondaryAccounts.call(proc, {
         ...args,
         identity: entityMockUtils.getIdentityInstance({
-          getPrimaryAccount: entityMockUtils.getAccountInstance({ address: 'primaryAccount' }),
+          getPrimaryAccount: primaryAccount,
           getSecondaryAccounts: [],
         }),
```
monitz87, line 131: This shouldn't be necessary either. This procedure's args don't have an `identity` property

src/api/procedures/modifySignerPermissions.ts
```diff
@@ -14,7 +14,7 @@ export interface ModifySignerPermissionsParams {
    * list of secondary accounts
    */
   secondaryAccounts: {
```
monitz87, line 16: This can be `Modify<PermissionedAccount, { permissions: PermissionsLike }>[]`

src/testUtils/mocks/dataSources.ts
```diff
@@ -317,8 +317,8 @@ interface ContextOptions {
   getIdentity?: Identity;
   getIdentityClaimsFromChain?: ClaimData[];
   getIdentityClaimsFromMiddleware?: ResultSet<ClaimData>;
-  primaryAccount?: string;
-  secondaryAccounts?: SecondaryAccount[];
+  primaryAccount?: PermissionedAccount;
```
monitz87, line 320: Considering this will always have the same permissions, I'd change it to be `string` again and just wrap it in an object with permissions inside the configure method to avoid having to write the same permissions over and over in tests

monitz87 (changes requested):

src/Polymesh.ts
```diff
@@ -136,13 +136,20 @@ export class Polymesh {
...
+      accountMnemonic,
+      signer,
+      middleware,
+    } = params;
     let context: Context;
 
     /* istanbul ignore next: part of configuration, doesn't need to be tested */
     const wsp = new WebSocketAsPromised(nodeUrl, {
-      createWebSocket: url => new W3CWebSocket(url) as unknown as WebSocket,
+      createWebSocket: url => (new W3CWebSocket(url) as unknown) as WebSocket,
```
polymath-eric, line 152: FYI: I bumped the prettier version, so merging alpha, running `yarn` and then `yarn format` should clean up the diff

polymath-eric (approved):

monitz87 (approved):

PR #666, opened 2022-01-14
feat: 🎸 Add Instruction methods to Settlements namespace

polymath-eric:
Add methods to settlement namespace so an Instruction lookup isn't
needed when the ID is already known


polymath-eric:
Turns out the current setup is pretty good for exposing methods in different areas.

src/api/procedures/__tests__/addInstruction.ts
```diff
@@ -451,7 +451,7 @@ describe('addInstruction procedure', () => {
       let boundFunc = getAuthorization.bind(proc);
 
       let result = await boundFunc({
-        venue,
+        venueId: venue.id,
```
prashantasdeveloper, line 454:
```suggestion
        venueId,
```

src/api/procedures/__tests__/addInstruction.ts
```diff
@@ -222,7 +222,7 @@ describe('addInstruction procedure', () => {
 
     venue = entityMockUtils.getVenueInstance({ id: venueId });
```
prashantasdeveloper, line 223: This can be removed

src/api/procedures/addInstruction.ts
```diff
@@ -72,11 +71,15 @@ export interface AddInstructionsParams {
...
 }
 
+export interface AddInstructionParamsWithVenueId extends AddInstructionParams {
+  venueId: BigNumber;
+}
+
 /**
  * @hidden
  */
 export type Params = AddInstructionsParams & {
```
prashantasdeveloper, line 81:
```suggestion
export type Params = AddInstructionsWithVenueIdParams;
```

prashantasdeveloper (changes requested):

src/api/procedures/addInstruction.ts
prashantasdeveloper: Oh I got confused between two separate methods. You can leave this as it is.

prashantasdeveloper (approved):
LGTM. @monitz87 you can review this now.

src/Settlements.ts
```diff
@@ -75,4 +100,18 @@ export class Settlements {
...
+  public addInstruction: ProcedureMethod<
+    AddInstructionWithVenueIdParams,
+    Instruction[],
+    Instruction
+  >;
+
+  /**
+   * Affirm an instruction (authorize)
+   */
+  public affirmInstruction: ProcedureMethod<BigNumber, Instruction>;
```
monitz87, line 116: This one should have `args: { id: BigNumber }` as parameters (consistency with all other methods)

monitz87 (changes requested):
We don't seem to be checking if the venue exists in addInstruction. Can you please add a check? Also, please check if modifyInstructionAffirmation checks for instruction existence, and add that too if not

monitz87:
> Turns out the current setup is pretty good for exposing methods in different areas.

Yeah, we designed it with reusability in mind. Good to see it paid off

src/api/procedures/utils.ts
```diff
@@ -15,7 +16,9 @@ import {
   PolymeshError,
   PostTransactionValue,
   TickerReservation,
+  Venue,
 } from '~/internal';
+import { VenueId } from '~/polkadot';
```
monitz87: This should be removed

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -52,8 +55,12 @@ export async function prepareModifyInstructionAffirmation(
   const { operation, id } = args;
 
   const instruction = new Instruction({ id }, context);
+  const { venue } = await instruction.details();
 
-  await assertInstructionValid(instruction, context);
+  await Promise.all([
+    assertVenueExists(venue.id, context),
```
monitz87, line 61: This is not necessary. Venues can't be removed, so if an instruction was created on one (as reflected by the fact that instruction.details returns it), then the Venue must exist. What I meant to say in my comment before was that we should validate whether the instruction exists (not knowing if it was already being done). Since it is already being done by `assertInstructionValid`, then we shouldn't do anything else

src/Settlements.ts
```diff
@@ -113,5 +114,5 @@ export class Settlements {
   /**
    * Affirm an instruction (authorize)
```
monitz87, line 115: Uppercase Instruction

monitz87 (changes requested):

monitz87 (approved):

PR #668, opened 2022-01-17
feat: 🎸 Add Claim typeguards 

polymath-eric:
Add typeguard functions to cast Claims into specific types

BREAKING CHANGE: 🧨 isSingleClaimCondition, isMultiClaimCondition,  isPortfolioCustodianRole, isVenueOwnerRole, isCddProviderRole, isTickerOwnerRole, isIdentityRole moved from `src/types/index.ts` to `src/utils/typeguards.ts`


prashantasdeveloper (approved):
LGTM

prashantasdeveloper:
@monitz87  you can review this now

src/types/index.ts
```diff
@@ -251,58 +216,52 @@ export enum ClaimType {
...
 export type InvestorUniquenessClaim = {
   type: ClaimType.InvestorUniqueness;
   scope: Scope;
   cddId: string;
   scopeId: string;
 };
 
+export type NoDataClaim = { type: ClaimType.NoData };
+
 export type InvestorUniquenessV2Claim = {
```
monitz87, line 246: All of these should be interfaces. Even though in current typescript versions there aren't many differences between type and interface, there still exist some, and we're using interface wherever possible. Better be consistent

src/types/index.ts
```diff
@@ -251,58 +216,52 @@ export enum ClaimType {
...
+
+export type AffiliateClaim = { type: ClaimType.Affiliate; scope: Scope };
+
+export type BuyLockupClaim = { type: ClaimType.BuyLockup; scope: Scope };
+
+export type SellLockupClaim = { type: ClaimType.SellLockup; scope: Scope };
+
 export type CddClaim = { type: ClaimType.CustomerDueDiligence; id: string };
 
+export type KnowYourCustomerClaim = { type: ClaimType.KnowYourCustomer; scope: Scope };
```
monitz87, line 229: Maybe `KycClaim`? Otherwise the other one should be `CustomerDueDiligenceClaim`

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+ * Return whether a Claim is a ScopedClaim
+ */
+export function isScopedClaim(claim: Claim): claim is ScopedClaim {
+  const { type } = claim;
+
+  return (
+    ![ClaimType.NoData, ClaimType.CustomerDueDiligence, ClaimType.InvestorUniquenessV2].includes(
+      type
+    ) && 'scope' in claim
+  );
```
monitz87, line 200: It should be enough to check that the scope property exists

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+/**
+ * Return whether a Claim is an UnscopedClaim
+ */
+export function isUnscopedClaim(claim: Claim): claim is UnscopedClaim {
+  const { type } = claim;
+  return [
+    ClaimType.NoData,
+    ClaimType.CustomerDueDiligence,
+    ClaimType.InvestorUniquenessV2,
+  ].includes(type);
```
monitz87, line 212: It should be enough to check that the scope property *doesn't* exist

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+export function isCustomerDueDiligenceClaim(claim: Claim): claim is CddClaim {
+  return claim.type === ClaimType.CustomerDueDiligence;
+}
+
+/**
+ * Return whether Claim is a KnowYourCustomerClaim
+ */
+export function isKnowYourCustomerClaim(claim: Claim): claim is KnowYourCustomerClaim {
+  return claim.type === ClaimType.KnowYourCustomer;
+}
```
monitz87, line 255: Make sure the comments/functions/types have matching names

src/types/index.ts
```diff
@@ -251,58 +216,52 @@ export enum ClaimType {
...
+
+export type JurisdictionClaim = { type: ClaimType.Jurisdiction; code: CountryCode; scope: Scope };
+
+export type ExemptClaim = { type: ClaimType.Exempted; scope: Scope };
```
monitz87, line 233: ExemptedClaim

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+ * Return whether Claim is a JurisdictionClaim
+ */
+export function isJurisdictionClaim(claim: Claim): claim is JurisdictionClaim {
+  return claim.type === ClaimType.Jurisdiction;
+}
+
+/**
+ * Return whether Claim is a ExemptClaim
+ */
+export function isExemptClaim(claim: Claim): claim is ExemptClaim {
```
monitz87, line 267: Match name and comment to type

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+
+/**
+ * Return whether Claim is a NoDataClaim
+ */
+export function isNoDataClaim(claim: Claim): claim is NoDataClaim {
+  return claim.type === ClaimType.NoData;
+}
+
+/**
+ *  Return whether a Claim is an InvestorUniquenessV2Claim
```
monitz87, line 293:
```suggestion
 * Return whether a Claim is an InvestorUniquenessV2Claim
```

src/utils/typeguards.ts
```diff
@@ -156,3 +186,165 @@ export function isVenue(value: unknown): value is Venue {
...
+}
+
+/**
+ * Return whether Condition has multiple Claims
+ */
+export function isMultiClaimCondition(
+  condition: InputCondition
+): condition is InputConditionBase & MultiClaimCondition {
+  return [ConditionType.IsAnyOf, ConditionType.IsNoneOf].includes(condition.type);
+}
```
monitz87, line 315: These two should probably check for the existence of the `claim` and `claims` properties instead of the type. Shouldn't matter much, but it's a bit more future proof in case we start adding more condition types

monitz87 (changes requested):

src/types/index.ts
polymath-eric: Interfaces lack the operations like `&` and `|`, which we use for `ScopedClaim`, `UnscopedClaim` and `Claim`.

We can express the base types as interfaces, but the mentioned ones need to be types.

I think its more consistent to leave them all as types, unless there is a way to define the unions as interfaces, which I don't think is the case. I checked a few blogs and all the examples use type for the union part.

polymath-eric:
Looks like the merge commit + auto format caused a bunch of brackets to be added in. Looking into it.

src/types/index.ts
polymath-eric: I went with the long form since it looked more consistent.

src/types/index.ts
monitz87: Even though it's more consistent within the claim types to use `type`, it's more consistent with the code base at large to use `interface`

The rule of thumb we're using is "use interface unless it's not possible" (for example, when using unions, as you said)

polymath-eric: I'll change the ones I can to interface, but leave the compound ones as types.

Personally I think having them all types feels more consistent, but the rule of thumb is reasonable enough. Especially when there isn't much difference between them.

monitz87:
Other than the type/interface thing, LGTM

src/types/index.ts
monitz87: There used to be bigger differences. For example, types didn't show up by name in compilation errors sometimes, among other things. It's more of a legacy thing + the current typescript manual recommends it so in a certain way it could be considered language convention 😅

monitz87 (approved):

PR #670, opened 2022-01-18
feat: 🎸 Add RotatePrimaryKeyToSecondary Authorization

polymath-eric:


src/api/procedures/consumeJoinIdentityAuthorization.ts
```diff
@@ -25,6 +26,8 @@ export interface Storage {
 
 /**
  * @hidden
+ *
+ * Consumes JoinIdentity and RotatePrimaryKeyToSecondaryKey Authorizations
```
polymath-eric, line 30: Initially I made a duplicate procedure, but Sonar Cloud didn't like that. I like the consistency of having the name line up with the auth so its easy to fine. That said I think the duplicate line check is right in that its a fair amount of extra code.

I was thinking of updating this name, but in some sense the rotate auth is a join identity auth, except the added key become primary.

I also was thinking of putting the insides in a shared function, but since the consumes would be siblings in a rather big directory there isn't an obvious place for it. Not exactly a util function.

For now I am thinking either share the function like this makes the most sense.

src/api/procedures/consumeJoinIdentityAuthorization.ts
```diff
@@ -82,17 +108,17 @@ export async function getAuthorization(
   const {
     storage: { currentAccount, calledByTarget },
   } = this;
-
   let hasRoles = calledByTarget;
 
+  const { type } = authRequest.data;
```
prashantasdeveloper, line 113:
```suggestion
  const { data: { type } } = authRequest;
```

src/api/procedures/consumeJoinIdentityAuthorization.ts
prashantasdeveloper: Sharing sounds good to me but only thing is that can we change the name that reflects both the actions something like `AddOrRotateAccount` (wherever JoinIdentity is referred here). 
This is the best I could relate to (`JoinIdentity` eventually adds an account, considering we are referring everything as account and not key, and `RotatePrimaryKeyToSecondaryKey` shortened to `RotateAccount` IDK but changing the name would be better.

src/api/procedures/utils.ts
```diff
@@ -556,6 +529,35 @@ function assertIsAccount(target: Signer): asserts target is Account {
   }
 }
 
+/**
+ * @hidden
+ *
+ * Asserts valid request for becoming a Primary Account (JoinIdentity and RotatePrimaryKeyToSecondary)
+ */
+async function assertJoinIdentityAuthorizationValid(
```
prashantasdeveloper, line 537: If we consider name change this could be `assertAddOrRotateAccountAuthorizationValid` which sounds more clear to me. IDK maybe it sounds good only in my head 😄

prashantasdeveloper (commented):
Left some name change suggestions, overall looks good

src/api/procedures/consumeJoinIdentityAuthorization.ts
polymath-eric: About the auth names, the "Key" part got left in so it matches with whats on chain. We could rename the variables to something else, but keep the values the same.

I think `AddOrRotateAccount` makes sense as a name for this.

src/api/procedures/__tests__/consumeJoinIdentityAuthorization.ts
```diff
@@ -9,13 +9,13 @@ import {
...
-import { Account, AuthorizationRequest, Context, Identity } from '~/internal';
+import { Account, AuthorizationRequest, Context, Identity, PolymeshError } from '~/internal';
 import { dsMockUtils, entityMockUtils, procedureMockUtils } from '~/testUtils/mocks';
 import { Mocked } from '~/testUtils/types';
-import { Authorization, AuthorizationType, Signer, TxTags } from '~/types';
+import { Authorization, AuthorizationType, ErrorCode, Signer, TxTags } from '~/types';
 import * as utilsConversionModule from '~/utils/conversion';
 
-describe('consumeJoinSignerAuthorization procedure', () => {
+describe('consumeJoinIdentityAuthorization procedure', () => {
```
prashantasdeveloper, line 18: Can you please also change the same in `consumeAddMultiSigSignerAuthorization` test as well?

src/api/procedures/consumeJoinIdentityAuthorization.ts
polymath-eric: Went with `JoinOrRotate`. Almost the same, but I like the `Join` as a bit of a hint.

src/api/procedures/utils.ts
```diff
@@ -556,6 +529,35 @@ function assertIsAccount(target: Signer): asserts target is Account {
...
 
+/**
+ * @hidden
+ *
+ * Asserts valid authorization for JoinIdentity and RotatePrimaryKeyToSecondary types
+ */
+async function assertJoinOrRotateAuthorizationValid(
+  authRequest: AuthorizationRequest
+): Promise<void> {
+  const issuer = authRequest.issuer;
```
monitz87, line 540: This and `target` can be destructured together

src/api/entities/AuthorizationRequest.ts
```diff
@@ -70,17 +70,18 @@ export class AuthorizationRequest extends Entity<UniqueIdentifiers, HumanReadabl
...
-   * | Add Relayer Paying Key      | Beneficiary, Relayer, Allowance |
-   * | Become Agent                | Permission Group                |
-   * | Attest Primary Key Rotation | DID                             |
-   * | Rotate Primary Key          | DID                             |
-   * | Transfer Ticker             | Ticker                          |
-   * | Add MultiSig Signer         | Account                         |
-   * | Transfer Token Ownership    | Ticker                          |
-   * | Join Identity               | DID                             |
-   * | Portfolio Custody           | Portfolio                       |
+   * | Type                            | Data                            |
```
monitz87, line 73: Please modify this table so that:
- Rotate Primary Key's data is "N/A"
- Rotate Primary Key to Secondary's data is "Permissions"
- Join Identity's data is "Permissions"

monitz87 (changes requested):
The `RotatePrimaryKey` authorization is being consumed as part of the `consumeAuthorizationRequests` Procedure. I think it makes sense to move it to the new Procedure you created

src/api/procedures/consumeJoinOrRotateAuthorization.ts
```diff
@@ -90,6 +94,12 @@ export async function prepareConsumeJoinOrRotateAuthorization(
       paidForBy: issuer,
       args: [rawAuthId],
     });
+  } else if (type === AuthorizationType.RotatePrimaryKey) {
+    this.addTransaction({
+      transaction: identity.acceptPrimaryKey,
+      paidForBy: issuer,
+      args: [rawAuthId, null],
+    });
   } else {
```
monitz87, line 103: These last two are very alike, we could avoid repeating the entire addTransaction call and only differentiate the value of `transaction` with a ternary

monitz87 (changes requested):

monitz87 (approved):

PR #672, opened 2022-01-21
feat: 🎸 Check for Portfolio existance in getName function

polymath-eric:
BREAKING CHANGE: 🧨 Portfolio getName can throw if Portfolio does not exist


src/api/procedures/deletePortfolio.ts
```diff
@@ -31,17 +31,7 @@ export async function prepareDeletePortfolio(
...
-    numberedPortfolio.getTokenBalances(),
-  ]);
-
-  if (!exists) {
-    throw new PolymeshError({
-      code: ErrorCode.DataUnavailable,
-      message: "The Portfolio doesn't exist",
-    });
-  }
+  const portfolioBalances = await numberedPortfolio.getTokenBalances();
```
polymath-eric, line 34: `getTokenBalances` performs the exist check, so the one here is redundant.

src/api/entities/__tests__/NumberedPortfolio.ts
```diff
@@ -89,20 +90,42 @@ describe('NumberedPortfolio class', () => {
...
+    const did = 'someDid';
+    const portfolioName = 'someName';
     test('should return the name of the Portfolio', async () => {
-      const id = new BigNumber(1);
-      const did = 'someDid';
-      const portfolioName = 'someName';
-      const rawPortfolioName = dsMockUtils.createMockText(portfolioName);
       const numberedPortfolio = new NumberedPortfolio({ id, did }, context);
-
+      const spy = jest.spyOn(numberedPortfolio, 'exists').mockResolvedValue(true);
```
polymath-eric, line 98: Looks like we don't use jest.spyOn, but its already installed, and this seems to be a good use case for it vs setting up the boiler plate thats in the exists check.

src/api/entities/NumberedPortfolio.ts
```diff
@@ -75,9 +81,17 @@ export class NumberedPortfolio extends Portfolio {
...
     } = this;
-
-    const rawPortfolioName = await portfolio.portfolios(did, numberToU64(id, context));
-
+    const [rawPortfolioName, exists] = await Promise.all([
+      portfolio.portfolios(did, numberToU64(id, context)),
+      this.exists(),
+    ]);
+
+    if (!exists) {
```
polymath-eric, line 89: Another approach would be to check the name being being non-empty.

It would save us a call, but duplicates the logic. Since its in parallel I think consistency is better.

prashantasdeveloper (approved):
LGTM

monitz87 (approved):

PR #675, opened 2022-01-25
feat: 🎸 Add `trustedFor` method in DefaultTrustedClaimIssuer

prashantasdeveloper:
BREAKING CHANGE: 🧨 
- `TrustedClaimIssuer` interface allows null value for `trustedFor` for specifying all claim types.
- `TrustedClaimIssuers.get` now returns `TrustedClaimIssuer<true>[]`
- `DefaultTrustedClaimIssuer.trustedFor` has been converted to an async method returning claim types for which Claim Issuer is trusted. 
- Allow `null` value for `trustedFor` parameter in
`ModifyTokenTrustedClaimIssuersAddSetParams.claimIssuers` to specify all
claim types for an identity


polymath-eric (approved):
LGTM, left a potential nit, but might be OK to leave.

Minor detail in the commit message, but technically "static property" means it's defined on the class itself and not instances. I know what you mean, but it's an overloaded word in the context.

src/api/procedures/modifyTokenTrustedClaimIssuers.ts
```diff
@@ -18,7 +18,10 @@ export interface ModifyTokenTrustedClaimIssuersAddSetParams {
   /**
    * array of Identity IDs
    */
-  claimIssuers: { identity: string | Identity; trustedFor?: ClaimType[] }[];
+  claimIssuers: {
+    identity: string | Identity;
+    trustedFor: ClaimType[] | null;
+  }[];
```
monitz87, line 24: This can be `InputTrustedClaimIssuer[]`

monitz87 (changes requested):
You forgot to list the changes to the `TrustedClaimIssuer` interface as breaking changes (and the methods that are affected by the interface change)

src/api/entities/DefaultTrustedClaimIssuer.ts
```diff
@@ -76,4 +70,37 @@ export class DefaultTrustedClaimIssuer extends Identity {
...
+    const claimIssuers = await complianceManager.trustedClaimIssuer(rawTicker);
+
+    const claimIssuer = claimIssuers
+      .map(issuer => trustedIssuerToTrustedClaimIssuer(issuer, context))
+      .find(({ identity: { did: issuerDid } }) => did === issuerDid);
+
+    if (!claimIssuer) {
+      throw new PolymeshError({
+        code: ErrorCode.DataUnavailable,
+        message: `The Identity with DID "${did}" is no longer a trusted issuer for "${ticker}"`,
```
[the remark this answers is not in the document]
prashantasdeveloper: As explained by Jere, I suppose no change is required here.

polymath-eric (approved):

src/types/index.ts
```diff
@@ -335,10 +336,14 @@ export interface ClaimScope {
...
-  identity: Identity;
+/**
+ * @param IsDefault - whether the Identity is a default trusted claim issuer for an asset or just
+ * for a specific compliance condition. Defaults to false
+ */
+export interface TrustedClaimIssuer<IsDefault extends boolean = false> {
+  identity: IsDefault extends true ? DefaultTrustedClaimIssuer : Identity;
   /**
-   * a null value means that the issuer is trusted for all claim types.
+   * an undefined value means that the issuer is trusted for all claim types.
```
monitz87, line 346:
```suggestion
   * a null value means that the issuer is trusted for all claim types
```

monitz87 (approved):

PR #677, opened 2022-01-27
fix: 🐛 Allow `modifyClaims` to modify claim with cdd_id `0x00`

prashantasdeveloper:


polymath-eric (approved):

VictorVicente (changes requested):
@prashantasdeveloper I left a comment regarding CddIds.
Also, could you check we are covering all these cases with unit tests:
1) `currentCddId` is the default and `newCddId` is different (should success)
2) `newCddId` is the default and `currentCddId` is different (should success)
3) `currentCddId` is some_value and `newCddId` is the another_value (should be invalid)

src/utils/constants.ts
```diff
@@ -134,3 +134,8 @@ export const SYSTEM_VERSION_RPC_CALL = {
   method: 'system_version',
   params: [],
 };
+
+/**
+ * Default cdd_id associated with an Identity on chain
```
monitz87, line 139:
```suggestion
 * Default CDD ID associated with an Identity on chain. Used for Identities onboarded without PUIS
```

monitz87 (approved):

VictorVicente (approved):

PR #680, opened 2022-02-02
fix: 🐛 Set `txHash` before setting transaction status to `Running`

prashantasdeveloper:


polymath-eric (approved):
LGTM

VictorVicente (approved):

VictorVicente (commented):
Approved, but is this being tested anywhere? Is it possible to add a test checking that txHash is present when `TransactionStatus.Running` is set?

monitz87:
> Approved, but is this being tested anywhere? Is it possible to add a test checking that txHash is present when `TransactionStatus.Running` is set? 

Agreed. This should be tested. I think it's possible (not at my computer right now)

PR #682, opened 2022-02-03
feat: 🎸 Use nonce: -1 for signAndSend

polymath-eric:
Using this option will account for pending transactions when auto
incrementing the account nonce. This allows for a better experience when
submitting multiple transactions for a single account. Each transaction
needs time to end up in the transaction pool, so some wait time may
still be needed.


monitz87 (approved):
I think this works for now, but we should expose the nonce to the user as part of `ProcdedureOptions`, and have a better value than "-1" for this case (probably null)

polymath-eric:
Made a card for exposing the nonce to users: https://polymath.atlassian.net/browse/MSDK-814

PR #684, opened 2022-02-08
feat: 🎸 Add modify allowance procedures to `Subsidy` 

prashantasdeveloper:
Add `setAllowance`, `increaseAllowance` & `decreaseAllowance` procedures
to `Subsidy` entity allowing the subsidizer to modify allowance of a
Subsidy relationship

### DIFF
--- a/src/api/entities/Subsidy/__tests__/index.ts
+++ b/src/api/entities/Subsidy/__tests__/index.ts
@@ -4,6 +4,7 @@ import sinon from 'sinon';
 import { Context, Entity, Subsidy, TransactionQueue } from '~/internal';
 import { dsMockUtils, entityMockUtils, procedureMockUtils } from '~/testUtils/mocks';
 import { Mocked } from '~/testUtils/types';
+import { AllowanceOperation } from '~/types';
 
 jest.mock(
   '~/base/Procedure',
@@ -83,6 +84,84 @@ describe('Subsidy class', () => {
         .resolves(expectedQueue);
 
       const queue = await subsidy.quit();
+
+      expect(queue).toBe(expectedQueue);
+    });
+  });
+
+  describe('method: setAllowance', () => {
+    afterAll(() => {
+      sinon.restore();
+    });
+
+    test('should prepare the setAllowance procedure with the correct arguments and context, and return the resulting transaction queue', async () => {
+      const args = { allowance: new BigNumber(50) };
+
+      const expectedQueue = 'mockQueue' as unknown as TransactionQueue<void>;
+
+      procedureMockUtils
+        .getPrepareStub()
+        .withArgs(
+          { args: { ...args, subsidy, operation: AllowanceOperation.Set }, transformer: undefined },
+          context
+        )
+        .resolves(expectedQueue);
+
+      const queue = await subsidy.setAllowance(args);
+
+      expect(queue).toBe(expectedQueue);
+    });
+  });
+
+  describe('method: increaseAllowance', () => {
+    afterAll(() => {
+      sinon.restore();
+    });
+
+    test('should prepare the increaseAllowance procedure with the correct arguments and context, and return the resulting transaction queue', async () => {
+      const args = { allowance: new BigNumber(50) };
+
+      const expectedQueue = 'mockQueue' as unknown as TransactionQueue<void>;
+
+      procedureMockUtils
+        .getPrepareStub()
+        .withArgs(
+          {
+            args: { ...args, subsidy, operation: AllowanceOperation.Increase },
+            transformer: undefined,
+          },
+          context
+        )
+        .resolves(expectedQueue);
+
+      const queue = await subsidy.increaseAllowance(args);
+
+      expect(queue).toBe(expectedQueue);
+    });
+  });
+
+  describe('method: decreaseAllowance', () => {
+    afterAll(() => {
+      sinon.restore();
+    });
+
+    test('should prepare the decreaseAllowance procedure with the correct arguments and context, and return the resulting transaction queue', async () => {
+      const args = { allowance: new BigNumber(50) };
+
+      const expectedQueue = 'mockQueue' as unknown as TransactionQueue<void>;
+
+      procedureMockUtils
+        .getPrepareStub()
+        .withArgs(
+          {
+            args: { ...args, subsidy, operation: AllowanceOperation.Decrease },
+            transformer: undefined,
+          },
+          context
+        )
+        .resolves(expectedQueue);
+
+      const queue = await subsidy.decreaseAllowance(args);
 
       expect(queue).toBe(expectedQueue);
     });
--- a/src/api/entities/Subsidy/index.ts
+++ b/src/api/entities/Subsidy/index.ts
@@ -1,7 +1,15 @@
 import BigNumber from 'bignumber.js';
 
-import { Account, Context, Entity, PolymeshError, quitSubsidy } from '~/internal';
-import { ErrorCode, NoArgsProcedureMethod } from '~/types';
+import {
+  Account,
+  Context,
+  Entity,
+  modifyAllowance,
+  ModifyAllowanceParams,
+  PolymeshError,
+  quitSubsidy,
+} from '~/internal';
+import { AllowanceOperation, ErrorCode, NoArgsProcedureMethod, ProcedureMethod } from '~/types';
 import { createProcedureMethod, toHumanReadable } from '~/utils/internal';
 
 export interface UniqueIdentifiers {
@@ -55,6 +63,60 @@ export class Subsidy extends Entity<UniqueIdentifiers, HumanReadable> {
       { getProcedureAndArgs: () => [quitSubsidy, { subsidy: this }], voidArgs: true },
       context
     );
+
+    this.setAllowance = createProcedureMethod<
+      Pick<ModifyAllowanceParams, 'allowance'>,
+      ModifyAllowanceParams,
+      void
+    >(
+      {
+        getProcedureAndArgs: args => [
+          modifyAllowance,
+          {
+            ...args,
+            subsidy: this,
+            operation: AllowanceOperation.Set,
+          } as ModifyAllowanceParams,
+        ],
+      },
+      context
+    );
+
+    this.increaseAllowance = createProcedureMethod<
+      Pick<ModifyAllowanceParams, 'allowance'>,
+      ModifyAllowanceParams,
+      void
+    >(
+      {
+        getProcedureAndArgs: args => [
+          modifyAllowance,
+          {
+            ...args,
+            subsidy: this,
+            operation: AllowanceOperation.Increase,
+          } as ModifyAllowanceParams,
+        ],
+      },
+      context
+    );
+
+    this.decreaseAllowance = createProcedureMethod<
+      Pick<ModifyAllowanceParams, 'allowance'>,
+      ModifyAllowanceParams,
+      void
+    >(
+      {
+        getProcedureAndArgs: args => [
+          modifyAllowance,
+          {
+            ...args,
+            subsidy: this,
+            operation: AllowanceOperation.Decrease,
+          } as ModifyAllowanceParams,
+        ],
+      },
+      context
+    );
   }
 
   /**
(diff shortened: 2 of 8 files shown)


src/api/entities/Subsidy/index.ts
```diff
@@ -64,6 +126,31 @@ export class Subsidy extends Entity<UniqueIdentifiers, HumanReadable> {
...
+   * @note Only the subsidizer is allowed to increase the allowance
+   */
+  public increaseAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;
+
+  /**
+   * Decrease allowance for this Subsidy relationship.
+   *
+   * @note Only the subsidizer is allowed to decrease the allowance
+   *
+   * @throws if the allowance to decrease is not less than the existing allowance
```
polymath-eric, line 150:
```suggestion
   * @throws if the amount to decrease by is more than the existing allowance
```

(nit) I think "more than" is easier to understand than "not less than".

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,145 @@
...
+  this.addTransaction({
+    transaction,
+    args: [rawBeneficiaryAccount, rawAllowance],
+  });
+}
+
+/**
+ * @hidden
+ *
+ * To modify the allowance for a Subsidy, the caller should be the subsidizer
```
polymath-eric, line 110:
```suggestion
 * To modify the allowance for a Subsidy, the caller must be the subsidizer
```

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  const currentAccount = context.getCurrentAccount();
+
+  const hasRoles = subsidizer.isEqual(currentAccount);
+
+  const transactionMap = new Map<AllowanceOperation, RelayerTx>([
+    [AllowanceOperation.Increase, TxTags.relayer.IncreasePolyxLimit],
+    [AllowanceOperation.Decrease, TxTags.relayer.DecreasePolyxLimit],
+    [AllowanceOperation.Set, TxTags.relayer.UpdatePolyxLimit],
+  ]);
```
polymath-eric, line 130:
```suggestion
    const transactionMap = {
    [AllowanceOperation.Increase]: TxTags.relayer.IncreasePolyxLimit,
    [AllowanceOperation.Decrease]: TxTags.relayer.DecreasePolyxLimit,
    [AllowanceOperation.Set]: TxTags.relayer.UpdatePolyxLimit,
  };
```

With an object TS will be able to tell it matches all cases and you won't need the no null assertion

polymath-eric (commented):
LGTM, only left nitty comments around docs + a way to avoid the not null assertion

src/api/entities/Subsidy/types.ts
```diff
@@ -21,3 +21,9 @@ export interface SubsidyWithAllowance {
   subsidy: Subsidy;
   allowance: BigNumber;
 }
+
+export enum AllowanceOperation {
```
monitz87, line 25: This should be exported from `~/types/internal`

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,144 @@
+import BigNumber from 'bignumber.js';
+import { TxTags } from 'polymesh-types/types';
```
monitz87, line 2: This should be imported from `~/types`. There are a couple of other places where this can be corrected

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,144 @@
...
+  operation: AllowanceOperation.Set;
+}
+
+export type ModifyAllowanceParams = (
+  | IncreaseAllowanceParams
+  | DecreaseAllowanceParams
+  | SetAllowanceParams
+) & {
+  subsidy: Subsidy;
+};
```
monitz87, line 39: Since you're using `ModifyAllowanceParams` in the actual endpoints, the user won't see the separate doc comments for each case. I suggest exporting each interface and using those in the endpoints

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,144 @@
...
+      });
+    }
+
+    transaction = tx.relayer.updatePolyxLimit;
+  }
+
+  if (operation === AllowanceOperation.Decrease) {
+    if (currentAllowance.lte(allowance)) {
+      throw new PolymeshError({
+        code: ErrorCode.NoDataChange,
```
monitz87, line 93: Should be a ValidationError

src/api/entities/Subsidy/index.ts
```diff
@@ -64,6 +126,31 @@ export class Subsidy extends Entity<UniqueIdentifiers, HumanReadable> {
...
+  public increaseAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;
+
+  /**
+   * Decrease allowance for this Subsidy relationship.
+   *
+   * @note Only the subsidizer is allowed to decrease the allowance
+   *
+   * @throws if the amount to decrease by is more than the existing allowance
+   */
+  public decreaseAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;
```
monitz87, line 152:
```suggestion
  /**
   * Set allowance for this Subsidy relationship
   *
   * @note Only the subsidizer is allowed to set the allowance
   *
   * @throws if the allowance to set is equal to the current allowance
   */
  public setAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;

  /**
   * Increase allowance for this Subsidy relationship
   *
   * @note Only the subsidizer is allowed to increase the allowance
   */
  public increaseAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;

  /**
   * Decrease allowance for this Subsidy relationship
   *
   * @note Only the subsidizer is allowed to decrease the allowance
   *
   * @throws if the amount to decrease by is more than the existing allowance
   */
  public decreaseAllowance: ProcedureMethod<Pick<ModifyAllowanceParams, 'allowance'>, void>;
```

monitz87: (Just removed some periods at the ends of sentences)

src/api/procedures/__tests__/modifyAllowance.ts
```diff
@@ -0,0 +1,206 @@
...
+  let allowance: BigNumber;
+  let rawBeneficiaryAccountId: AccountId;
+  let rawAllowance: Balance;
+
+  let increasePolyxLimitTransaction: sinon.SinonStub;
+
+  beforeAll(() => {
+    dsMockUtils.initMocks();
+    procedureMockUtils.initMocks();
+    entityMockUtils.initMocks();
```
monitz87, line 41: Probably better to explicitly set the current allowance, it's a bit more readable

monitz87 (changes requested):

src/api/procedures/__tests__/modifyAllowance.ts
```diff
@@ -0,0 +1,210 @@
...
+    procedureMockUtils.cleanup();
+    dsMockUtils.cleanup();
+  });
+
+  test('should throw an error if the Subsidy does not exist', async () => {
+    const proc = procedureMockUtils.getInstance<ModifyAllowanceParams, void>(mockContext);
+
+    let error;
+
+    try {
```
polymath-eric, line 83: FYI, I know we've done this pattern a lot, but there is a more succinct way of testing exceptions.
```suggestion
    expect(() => prepareModifyAllowance.call(...)).rejects.toThrowError(expectedError)
```
`.rejects` and `.resolves` are for promises

You might not need to do the anonymous function with a promise, but you do need it for non promises. You can also use `.not.toThrow()` to test for no exception.

The test doesn't need to be async if you return the expect, but I think await also works.

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,143 @@
+import BigNumber from 'bignumber.js';
+
+import { PolymeshError, Procedure, Subsidy } from '~/internal';
+import { ErrorCode, TxTags } from '~/types';
+import { AllowanceOperation, ProcedureAuthorization } from '~/types/internal';
+import { bigNumberToBalance, stringToAccountId } from '~/utils/conversion';
+
+export interface IncreaseAllowanceParams {
+  /**
+   * amount of POLYX to be increased
```
polymath-eric, line 10:
```suggestion
   * amount of POLYX to be increase the allowance by
```

nit, but the phrasing seems a weird to me without a noun / subject

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  allowance: BigNumber;
+  operation: AllowanceOperation.Increase;
+}
+
+export interface DecreaseAllowanceParams {
+  /**
+   * amount of POLYX to be decreased
```
polymath-eric, line 18:
```suggestion
   * amount of POLYX to be decrease the allowance by
```

src/api/procedures/modifyAllowance.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  allowance: BigNumber;
+  operation: AllowanceOperation.Decrease;
+}
+
+export interface SetAllowanceParams {
+  /**
+   * amount of POLYX to be set
```
polymath-eric, line 26:
```suggestion
   * amount of POLYX to set the allowance to
```

polymath-eric (commented):
LGTM, left a few nitty comments, but nothing substantial.

src/api/procedures/modifyAllowance.ts
```diff
@@ -79,7 +78,7 @@ export async function prepareModifyAllowance(
   if (operation === AllowanceOperation.Set) {
     if (currentAllowance.eq(allowance)) {
       throw new PolymeshError({
-        code: ErrorCode.NoDataChange,
+        code: ErrorCode.ValidationError,
```
monitz87, line 81: This one was ok, shouldn't have been changed

monitz87 (changes requested):

src/api/procedures/__tests__/modifyAllowance.ts
```diff
@@ -76,62 +80,36 @@ describe('modifyAllowance procedure', () => {
...
+      prepareModifyAllowance.call(proc, {
         ...args,
         subsidy: entityMockUtils.getSubsidyInstance({ exists: false }),
-      });
-    } catch (err) {
-      error = err;
-    }
-
-    expect(error.message).toBe('The Subsidy no longer exists');
+      })
```
monitz87, line 87: No need for the arrow function, and you can just `return expect` and have the method not be async. Using `await expect` is for when you need to test several things in the same `test` call. That can be changed in multiple places I think

monitz87 (changes requested):

src/api/entities/Subsidy/__tests__/index.ts
```diff
@@ -88,6 +89,84 @@ describe('Subsidy class', () => {
     });
   });
 
+  describe('method: setAllowance', () => {
+    afterAll(() => {
+      sinon.restore();
+    });
+
+    test('should prepare the setAllowance procedure with the correct arguments and context, and return the resulting transaction queue', async () => {
```
polymath-eric, line 97: I think we want to use `it` now, considering this is going into alpha

src/api/procedures/__tests__/modifyAllowance.ts
```diff
@@ -0,0 +1,183 @@
...
+    procedureMockUtils.reset();
+    dsMockUtils.reset();
+  });
+
+  afterAll(() => {
+    procedureMockUtils.cleanup();
+    dsMockUtils.cleanup();
+  });
+
+  test('should throw an error if the Subsidy does not exist', () =>
```
polymath-eric, line 82:
```suggestion
  it('should throw an error if the Subsidy does not exist', () =>
```

As well as others in this file

src/api/entities/Subsidy/__tests__/index.ts
polymath-eric: This looks like its still `test` instead of `it`

prashantasdeveloper: My bad. The file somehow didn't save. Pushed up a small change for this

polymath-eric (approved):
LGTM

I wonder what happened to the formatting though. I thought we updated prettier and had most files without the brackets already.

prashantasdeveloper:
> I wonder what happened to the formatting though. I thought we updated prettier and had most files without the brackets already.

Seems somehow they got added back in the PR https://github.com/PolymathNetwork/polymesh-sdk/pull/703

polymath-eric:
> > I wonder what happened to the formatting though. I thought we updated prettier and had most files without the brackets already.
> 
> Seems somehow they got added back in the PR #703

Whoops, guess that was my bad.

For the dependency bump I've been working off of beta, and I think I forgot to run yarn when switching branches.

monitz87 (approved):

PR #687, opened 2022-02-15
Msdk 678 fix entity compare

polymath-eric:


prashantasdeveloper:
Overall LGTM, but I agree with more coupling introduced. Also it would be great if you can add the tests for `getIdentity` function

src/api/procedures/consumeAuthorizationRequests.ts
```diff
@@ -111,26 +111,22 @@ export async function getAuthorization(
 ): Promise<ProcedureAuthorization> {
   const { context } = this;
 
-  let did: string;
+  const identity = await context.getCurrentIdentity();
```
monitz87, line 114: The idea behind `fetchDid` was to avoid making the call to the chain if not necessary. We should preserve that optimization I think

src/utils/internal.ts
```diff
@@ -134,6 +134,29 @@ export async function getDid(
...
+  } else {
+    return intoIdentity(value, context);
+  }
+}
+
+/**
+ * @hidden
+ * Given a DID return the corresponding Identity, given an Identity return the Identity
+ */
+export function intoIdentity(value: string | Identity, context: Context): Identity {
```
monitz87, line 156: Maybe `asIdentity` could be a better name, although I'm not sold either

src/api/procedures/modifyClaims.ts
```diff
@@ -136,9 +136,9 @@ const findInvalidCddClaims = async (
     });
 
     newCddClaims.forEach(({ target, claim }) => {
-      const did = signerToString(target);
-      const issuedClaimsForTarget = issuedCddClaims.data.filter(
-        ({ target: issuedTarget }) => issuedTarget.did === did
+      const targetId = intoIdentity(target, context);
```
monitz87, line 139: targetIdentity

src/testUtils/mocks/dataSources.ts
```diff
@@ -672,6 +673,7 @@ function configureContext(opts: ContextOptions): void {
       },
     }),
     getSecondaryAccounts: sinon.stub().resolves(opts.secondaryAccounts),
+    getCurrentIdentity: sinon.stub().resolves(opts.getCurrentIdentity),
```
monitz87, line 676: This shouldn't be necessary (you added a `getCurrentIdentity` method to the current Identity instance returned by the context mock)

monitz87: If you want to override the result of `context.getCurrentIdentity`, you should use the `getCurrentIdentity` stub defined in this same function.

Something like:
```
if (opts.getCurrentIdentity) {
  // set the getCurrentIdentityStub to resolve to that value
} else {
  // define identity and use that
}
```

Keep in mind that that would override most of the ContextOptions

src/utils/__tests__/internal.ts
```diff
@@ -8,6 +8,7 @@ import sinon from 'sinon';
 import { Asset, Context, PostTransactionValue, Procedure } from '~/internal';
 import { ClaimScopeTypeEnum } from '~/middleware/types';
 import { dsMockUtils, entityMockUtils } from '~/testUtils/mocks';
+import { getIdentityInstance } from '~/testUtils/mocks/entities';
```
monitz87, line 11: Use `entityMockUtils.getIdentityInstance`

monitz87 (changes requested):

src/api/entities/__tests__/Checkpoint.ts
```diff
@@ -216,13 +216,7 @@ describe('Checkpoint class', () => {
...
       });
 
-      const assetBalance = new BigNumber(10);
-
-      entityMockUtils.configureMocks({
-        identityOptions: {
-          getAssetBalance: assetBalance,
-        },
-      });
+      const assetBalance = new BigNumber(1000);
```
polymath-eric, line 219: I am not sure why this changed, the old setup doesn't seem to work, as in it still returns `1000` now.

monitz87: I'm not sure I understand

src/api/procedures/consumeAuthorizationRequests.ts
```diff
@@ -122,11 +123,13 @@ export async function getAuthorization(
...
     } else {
-      condition = target.isEqual(identity);
+      did = await fetchDid();
+      condition = did === target.did;
     }
 
     if (!accept) {
-      condition = condition || issuer.isEqual(identity);
+      did = await fetchDid();
+      condition = condition || did === issuer.did;
```
monitz87, line 132: I meant you should adapt the optimization to using `isEqual`. Probably the easiest way to go about it is to do something like

```
let identity: Identity;
const fetchIdentity = async () => identity || context.getCurrentIdentity();
```

And use that in a similar manner

src/api/procedures/modifyClaims.ts
```diff
@@ -25,7 +25,7 @@ import {
   stringToScopeId,
   stringToTicker,
 } from '~/utils/conversion';
-import { assembleBatchTransactions, intoIdentity } from '~/utils/internal';
+import { asIdentity,assembleBatchTransactions } from '~/utils/internal';
```
monitz87, line 28:
```suggestion
import { asIdentity, assembleBatchTransactions } from '~/utils/internal';
```

src/utils/__tests__/internal.ts
```diff
@@ -853,9 +852,7 @@ describe('getIdentity', () => {
   let context: Context;
 
   beforeEach(() => {
-    context = dsMockUtils.getContextInstance({
-      getCurrentIdentity: entityMockUtils.getIdentityInstance(),
-    });
+    context = dsMockUtils.getContextInstance({});
```
monitz87, line 855:
```suggestion
    context = dsMockUtils.getContextInstance();
```

monitz87 (changes requested):

src/utils/__tests__/internal.ts
polymath-eric: I tried that, and it gives different behavior. 

When passing `undefined` configureContext won't get called:

```
export function getContextInstance(opts?: ContextOptions): MockContext {
  if (opts) {
    configureContext({ ...defaultContextOptions, ...opts });
  }
  return mockInstanceContainer.contextInstance;
}
```

src/api/entities/__tests__/Checkpoint.ts
polymath-eric: The old setup of "10" wasn't working, but now Identity is coming from a different call, so now I see why its happening.

src/utils/__tests__/internal.ts
monitz87: You're missing the `beforeAll` with `dsMockUtils.initMocks()`

monitz87 (changes requested):

monitz87 (approved):

PR #694, opened 2022-02-24
refactor: 💡 Replace jest `test` with `it`

prashantasdeveloper:


polymath-eric (commented):
I think you missed the cases when the string is double quoted

searching for `test("` brings up 50 results. Could you update those as well?

LGTM otherwise.

Good catch with the STO -> Offering rename.

polymath-eric (approved):

PR #701, opened 2022-02-24
Msdk 821 upgrade dependencies

polymath-eric:
I went through all the dependabot alerts. I used the suggestion when it was possible, otherwise I bumped package versions, or resolved stuff manually when I could.

There are still going to be a few warnings, but mainly on dev dependencies. 

Two noteworthy examples are 
- eslint which has a transitive dependency, and I wan't able to update that. It would give a CLIEngine undefined. I tried looking into it, but it seems like there is an issue with "vue-eslint" (I am surprised we have that), which doesn't play nice with the newer versions.

- typedoc has a regex vulnerability, when I tried bumping the version `yarn build:docs` broke and left a temp directory. I didn't spend much time on this.

I also opted for ^ dependencies, I think this will give a bit more flexibility and keep us on newer packages. I can switch them to pinned ones, but I think we'll be OK with the lock file + test coverage.


polymath-eric:
I switched this to draft for now.

I was able to get webpack 5 working, but I needed to change two files imports in node_modules:

- node_modules/promise-controller/src/index.js
- node_modules/websocket-as-promised/src/index.js

```js
 require('./utils')
```

to

```js
require('./utils.js')
```

It seems like I should be able to do this from the webpack config, but I am not sure why the imports for used libraries got changed.

I am going to continue to look into this so I pushed what I have.

The alternative is to stay on old webpack, I am not sure on the benefits besides newer, but it seems like we'll eventually need to upgrade.

polymath-eric:
I am fairly happy with the replacement function, but not sure on a good way of testing it yet.

I tried to make a mock with event emitter and trigger it that way, and also to try invoking the `onmessage` function directly. However it seems like I am not getting the same instance in my test as the implementation.

src/Polymesh.ts
```diff
@@ -193,32 +190,17 @@ export class Polymesh {
...
 
-    /* istanbul ignore next: part of configuration, doesn't need to be tested */
-    const wsp = new WebSocketAsPromised(nodeUrl, {
-      createWebSocket: url => (new W3CWebSocket(url) as unknown) as WebSocket,
-      packMessage: data => JSON.stringify(data),
-      unpackMessage: data => JSON.parse(data.toString()),
-      attachRequestId: (data, requestId) => Object.assign({ id: requestId }, data),
-      extractRequestId: data => data && data.id,
+    await assertExpectedChainVersion(nodeUrl).catch(error => {
+      if (
```
polymath-eric, line 194: I figured since I was making a change here, I might as well make it a warning.

https://polymathnetwork.slack.com/archives/C026L2FKT25/p1646310363628679?thread_ts=1646304939.113329&cid=C026L2FKT25

webpack.config.dev.js
```diff
@@ -33,14 +34,16 @@ const devConfig = {
         test: /\.mjs$/,
         include: /node_modules/,
         type: 'javascript/auto',
+        resolve: {
```
polymath-eric, line 37: Technically, this resolve can be scoped to just `graphql` lib, but it seemed the cleanest to include here. It seems OK to leave as false.

https://webpack.js.org/configuration/module/#resolvefullyspecified

webpack.config.dev.js
```diff
@@ -52,19 +55,26 @@ const devConfig = {
...
     publicPath: '/',
   },
-  node: {
-    fs: 'empty',
-  },
   resolve: {
     extensions: ['.ts', '.js'],
     plugins: [new TsconfigPathsPlugin()],
+    fallback: {
+      crypto: require.resolve('crypto-browserify'),
```
polymath-eric, line 74: We need to explicitly add in these polyfills, the old version would automatically include them which would inflate the bundle size.

src/testUtils/mocks/dataSources.ts
```diff
@@ -187,18 +186,80 @@ function createApolloClient(): Mutable<ApolloClient<NormalizedCacheObject>> {
...
+function createWebSocket(): MockWebSocket {
+  return new MockWebSocket();
+}
+
+/**
+ * Creates mock websocket class. Contains additional methods for tests to control it
+ */
+export class MockWebSocket {
+  /**
+   *@hidden
```
monitz87, line 200: Add a space between the asterisk and tag (same for the others)

src/utils/__tests__/internal.ts
```diff
@@ -660,3 +671,49 @@ describe('isModuleOrTagMatch', () => {
...
+    const signal = assertExpectedChainVersion('ws://example.com');
+    client.sendVersion('3.0.0');
+    const expectedError = new PolymeshError({
+      code: ErrorCode.FatalError,
+      message: 'Unsupported Polymesh version. Please upgrade the SDK',
+    });
+    return expect(signal).rejects.toThrowError(expectedError);
+  });
+
+  it('should throw an error if the version does not match', () => {
```
monitz87, line 710: The description doesn't match what is being tested

src/testUtils/mocks/dataSources.ts
```diff
@@ -187,18 +186,80 @@ function createApolloClient(): Mutable<ApolloClient<NormalizedCacheObject>> {
...
+  close(): void {
+    // stub for close
+  }
+
+  /**
+   * @hidden
+   */
+  send(msg: string): void {
+    const response = { data: '{ "result": "4.1.1" }' };
+    setTimeout(() => this.onmessage(response), 10);
```
monitz87, line 240: Why the timeout here?

src/utils/internal.ts
```diff
@@ -710,3 +714,57 @@ export function toHumanReadable<T>(obj: T): HumanReadableType<T> {
...
 }
+
+/**
+ * @hidden
+ *
+ * Checks chain version. This function uses a websocket as it's intended to be called during initialization
+ * @param nodeUrl - URL for the chain node
+ * @returns A promise that resolves if the version is in the expected range, otherwise it will reject
+ */
+export function assertExpectedChainVersion(nodeUrl: string): Promise<unknown> {
```
monitz87, line 725: I'd recommend doing everything from inside the Promise constructor. It's more readable IMO. If not, at least define the callbacks and create the promise at the top of the function and use `resolve`, `reject`, `_resolve`, `_reject` for names

Also, you're mixing arrow notation with `function` in the event callbacks

src/utils/internal.ts
```diff
@@ -710,3 +714,57 @@ export function toHumanReadable<T>(obj: T): HumanReadableType<T> {
...
+        message: 'Unsupported Polymesh version. Please upgrade the SDK',
+        data: {
+          polymeshVersion: version,
+          supportedVersionRange: SUPPORTED_VERSION_RANGE,
+        },
+      });
+      fail(error);
+    } else {
+      success();
+      fail();
```
monitz87, line 760: One of these shouldn't be here, right?

monitz87 (changes requested):

polymath-eric:
I merged alpha into the branch from muscle memory and was wondering why there was so many conflicts. Going to take out the merge commit.

polymath-eric:
I merged alpha by mistake, going to force push to get that merge out of this history.

src/utils/__tests__/internal.ts
polymath-eric: My bad, you are right, will update

monitz87 (approved):

PR #708, opened 2022-03-04
Msdk 831 improve alpha quality

polymath-eric:


polymath-eric:
the if statement was getting +4 for being so nested, by bringing it up a level it should only get a +2 for nesting, which brings the total to the allowed 15 score.

There is also a similar comparator function in `getMissingPortfolioPermissions` below, so in some sense its a bit more consistent also.

Screen Shot 2022-03-04 at 1 59 37 PM

monitz87 (approved):

PR #709, opened 2022-03-04
Msdk 831 improve sonar issues

polymath-eric:
This resolves the Sonar Cloud issues that are not already addressed in alpha


src/api/procedures/modifyCorporateActionsAgent.ts
```diff
@@ -26,8 +26,6 @@ export type Params = { ticker: string } & ModifyCorporateActionsAgentParams;
 
 /**
  * @hidden
- *
- * @deprecated in favor of `inviteAgent`
```
polymath-eric, line 30: I figured these are "internal" procedures so we don't need the deprecated comment on them.

src/base/__tests__/Context.ts
```diff
@@ -56,7 +56,6 @@ jest.mock(
   require('~/testUtils/mocks/entities').mockSubsidyModule('~/api/entities/Subsidy')
 );
 
-// TODO: refactor tests (too much repeated code)
```
polymath-eric, line 59: If anything this should be an issue, but I think they are OK even if a bit verbose.

src/utils/internal.ts
```diff
@@ -214,7 +214,6 @@ export function unwrapValues<T extends unknown[]>(values: MapMaybePostTransactio
  */
 type EventData<Event> = Event extends AugmentedEvent<'promise', infer Data> ? Data : never;
 
-// TODO @monitz87: use event enum instead of string when it exists
```
polymath-eric, line 217: Looks like its not using a string anymore.

src/api/entities/Asset/__tests__/Settlements.ts
```diff
@@ -236,17 +236,17 @@ describe('Settlements class', () => {
     });
 
     beforeEach(() => {
-      fromPortfolio = entityMockUtils.getDefaultPortfolioInstance({
```
polymath-eric, line 239: It doesn't make sense to me to refactor these into a shared method with those from the `canSettle` case, but by changing the order they shouldn't be the same block as above from the perspective of static analysis.

src/Polymesh.ts
```diff
@@ -347,7 +347,6 @@ export class Polymesh {
     this.context.setPair(signerToString(signer));
   }
 
-  // TODO @monitz87: remove when the dApp team no longer needs it
```
polymath-eric, line 350: Seems like this is part of the API now, this comment as been in for a while now.

src/testUtils/index.ts
```diff
@@ -10,3 +13,16 @@ export async function fakePromise(): Promise<void> {
   await new Promise(resolve => setImmediate(() => resolve(null)));
   await new Promise(resolve => setImmediate(() => resolve(null)));
 }
+
+/**
+ * @hidden
+ *
+ * Awaits multiple promises and advances jest time after each one
+ */
+export async function manyFakePromises(): Promise<void> {
```
prashantasdeveloper, line 22:
```suggestion
export async function fakePromises(): Promise<void> {
```
Just a suggestion, but you can leave this as it as well.

prashantasdeveloper (approved):
Overall LGTM, just left one renaming suggestion.

monitz87 (approved):

monitz87 (approved):

PR #710, opened 2022-03-08
fix: 🐛 Explictly check claim type in isScopedClaim typeguard

polymath-eric:
Before isScopedClaim would only check for `scope` field being defined,
which allowed for incorrectly defined claims to pass the check


monitz87 (changes requested):
There's a linting error. LGTM otherwise

polymath-eric:
Fixed. Weird it was in an untouched file though. 

I am guessing it got added in from a beta to alpha merge. Not sure how else it would slip through since we need feature branches to be at the head of alpha.

prashantasdeveloper (approved):
LGTM

monitz87 (approved):

PR #712, opened 2022-03-09
docs: ✏️ Add note to checkpoint allBalances method

polymath-eric:
Adds a note indicating current holders not in the checkpoint will be
included in the allBalances results with a balance of 0

This behavior seems strange to me, I'd expect not to receive the balance at all, but makes sense to add a note if it has to be this way. The ticket does say: 

> We should also add some comments explaining the logic used to fetch the balances and why it is that way

But I am not sure why its this was either.


monitz87:
Checkpoints on chain work in a very peculiar manner so as to optimize storage use. There are three relevant storage entries for checkpoints: `balanceUpdates(ticker, did) -> Vec<CheckpointId>`, `balance([ticker, checkpointId], did) -> Balance` and `checkpointIdSequence(ticker) -> CheckpointId` (which holds the id of the last created checkpoint).

The following is all within the scope of a specific asset (so as not to have to repeat "for the asset" 100 times):

Whenever an identity's balance is going to change for any reason, if `balance` at the current checkpoint ID is empty, the current balance (before the change) is stored for that checkpoint ID, and the current checkpoint ID is added to the `balanceUpdates` array for that identity. This is so you can create a bunch of checkpoints, but storage size will not grow unless there is actual change in balances. What this means though, is that `balance` does not reflect the real balance of a user at a specific checkpoint

In order to calculate the real balance of `did` at checkpoint `N`, you need to:
- Fetch `balanceUpdates(ticker, did)`
- Find the closest checkpoint `M` in that array such that `M >= N`
  - If `M` exists, return `balance([ticker, M], did)`
  - Otherwise return the current balance

As a consequence to this, if an identity's balance hasn't changed since a checkpoint was created, their balance at that checkpoint will be their current balance, which is why when fetching all balances at a checkpoint we need to start with all current holders and iterate over them. The reason we can't filter 0 balances out is because of pagination

Let me know if the explanation is clear enough (this took me quite a bit of time to wrap my head around when I had to implement it)

src/api/entities/Checkpoint.ts
```diff
@@ -101,6 +101,7 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
   /**
    * Retrieve all Asset Holder balances at this Checkpoint
    *
+   * @note current Asset holders who didn't hold any tokens when the Checkpoint was created will be listed with a balance of 0
```
monitz87, line 104: Maybe sprinkle "due to chain limitations" at the beginning to make it clear that this isn't an arbitrary senseless decision

src/api/entities/Checkpoint.ts
```diff
@@ -119,6 +121,7 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
 
     const rawTicker = stringToTicker(ticker, context);
 
+    // Get one page of current Asset balances
```
polymath-eric, line 124: I am assuming paginated asset.balanceOf returns 0 for people who have sold all of their tokens but were included in a checkpoint.

Otherwise I think people who have had a balance in a checkpoint, but currently don't hold any now would be skipped.

monitz87: Yeah, you could potentially get several pages full of zeroes

src/api/entities/Checkpoint.ts
```diff
@@ -210,14 +220,18 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
       u64ToBigNumber(checkpointId).gte(id)
     );
 
+    // If there has been a balance change since the Checkpoint was created, then we can query the Checkpoint storage
+    // If there hasn't been a balance update, the storage will not have an entry for the Identity. Instead the current balance should be queried instead.
```
monitz87, line 224:
```suggestion
    /*
     * If there has been a balance change since the Checkpoint was created, then we can query the Checkpoint storage.
     * Otherwise, the storage will not have an entry for the Identity. The current balance should be queried instead.
     */
```

src/api/entities/Checkpoint.ts
```diff
@@ -210,14 +220,18 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
...
     let balance: BigNumber;
     if (firstUpdatedCheckpoint) {
+      // if a balance has happened after the checkpoint has been created, then we can query the Checkpoint storage for the balance
```
monitz87, line 227: I think this comment and the next one are a bit redundant, considering the previous block comment says the same thing

monitz87 (changes requested):

src/api/entities/Checkpoint.ts
```diff
@@ -128,6 +130,7 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
     const currentDidBalances: { did: string; balance: BigNumber }[] = [];
     const balanceUpdatesMultiParams: [Ticker, IdentityId][] = [];
 
+    // Prepare the query for balance updates for balance updates. Push to currentDidBalances in case there are none
```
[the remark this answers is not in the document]
polymath-eric: I mean "in case there are no updates"

polymath-eric:
I committed some multiSig files I had by mistake so I force pushed the recent commit.

monitz87 (changes requested):

PR #714, opened 2022-03-15
Add support for exempted identities in non-PUIS Asset transfer restrictions

monitz87:


src/api/entities/Identity/__tests__/index.ts
```diff
@@ -704,6 +705,7 @@ describe('Identity class', () => {
     let rawScopeId: ScopeId;
 
     let stringToTickerStub: sinon.SinonStub<[string, Context], Ticker>;
+    let scopeIdOfStub: sinon.SinonStub;
```
VictorVicente, line 708: There is a warning here as 'scopeIdOfStub' is defined but never used

src/types/index.ts
```diff
@@ -1216,16 +1216,16 @@ export interface ProcedureAuthorizationStatus {
...
   /**
-   * array of Scope IDs that are exempted from the Restriction
+   * array of Scope/Identity IDs that are exempted from the Restriction
+   *
+   * @note if the Asset requires investor uniqueness, Scope IDs are used. Otherwise, we use Identity IDs. More on Scope IDs and investor uniqueness
+   *   {@link https://developers.polymesh.network/introduction/identity#polymesh-unique-identity-system-puis | here} and
+   *   {@link https://developers.polymesh.network/polymesh-docs/primitives/confidential-identity | here}
    */
-  exemptedScopeIds?: string[];
+  exemptedIds?: string[];
```
VictorVicente, line 1225: Is it ok we are not supporting `Identity[]` here?

monitz87: The value could be a scopeId. Even if we check whether the asset has PUIS enabled or not, the agent could have exempted a ScopeId, invalid DID, or anything that fits the format

VictorVicente (approved):

PR #715, opened 2022-03-28
Da 35/add multisig entity

polymath-eric:


polymath-eric:
Note: If beta gets updated, then this should point there instead of alpha.

src/api/procedures/createMultiSig.ts
```diff
@@ -0,0 +1,68 @@
...
+      polymeshApi: { tx },
+    },
+    context,
+  } = this;
+  const { signers, signaturesRequired } = args;
+
+  if (signaturesRequired.gt(signers.length)) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'The number of signatures required should not exceed the number of signers',
```
monitz87, line 44:
```suggestion
      message: 'The number of required signatures should not exceed the number of signers',
```

src/api/entities/MultiSig.ts
```diff
@@ -0,0 +1,95 @@
...
+
+interface MultiSigDetails {
+  signers: Signer[];
+  signaturesRequired: BigNumber;
+}
+
+/**
+ * Represents a MultiSig Account. MultiSig functions like an { @link Account } except submitted extrinsics need to be approved by a number of signers
+ */
+export class MultiSig extends Account {
```
monitz87, line 16: This class should override the `getIdentity` method using `multiSig.multiSigToIdentity` to fetch the DID

src/testUtils/mocks/dataSources.ts
```diff
@@ -3487,3 +3489,54 @@ export const createMockSignedBlock = (
...
+
+/**
+ * @hidden
+ * NOTE: `isEmpty` will be set to true if no value is passed
+ */
+export const createMockProposalData = (proposalData?: {
+  args: unknown[];
+  method: string;
+  section: string;
+}): ProposalData => {
```
monitz87, line 3501: This type should be `Proposal`, not `ProposalData` (`ProposalData` is an enum)

src/testUtils/mocks/dataSources.ts
```diff
@@ -3487,3 +3489,54 @@ export const createMockSignedBlock = (
...
+    !proposalData
+  ) as ProposalData;
+};
+
+export const createMockProposalDetails = (proposalDetails?: {
+  approvals: string;
+  rejections: string;
+  status: string;
+  expiry: Option<Moment> | null;
+  autoClose: boolean;
```
monitz87, line 3523: Please use the proper types for the arguments. Look at `createMockAuthorization` for guidance (note that options must be passed as options)

src/utils/typeguards.ts
```diff
@@ -366,3 +367,18 @@ export function isPolymeshTransaction(value: unknown): value is PolymeshTransact
 export function isPolymeshTransactionBatch(value: unknown): value is PolymeshTransactionBatch {
   return value instanceof PolymeshTransactionBatch;
 }
+
+/**
+ * @hidden
+ */
+export function isProposalStatus(status: string): status is ProposalStatus {
```
monitz87, line 374: This shouldn't be necessary after turning `ProposalStatus` into an enum

monitz87 (changes requested):
Can you please also add `isMultiSig` to the typeguards file?

src/api/entities/MultiSig.ts
monitz87: Seems like I misunderstood what `multiSig.multiSigToIdentity` does

src/api/entities/MultiSig/MultiSigProposal.ts
```diff
@@ -0,0 +1,139 @@
...
+      status,
+      expiry,
+      autoClose,
+      args: args.map(a => a.toString()),
+      txTag: `${section}.${method}` as TxTag,
+    };
+  }
+
+  /**
+   * Determines whether or not this Proposal exists on chain
```
prashantasdeveloper, line 110:
```suggestion
   * Determines whether this Proposal exists on chain
```

src/api/entities/__tests__/MultiSig.ts
```diff
@@ -0,0 +1,144 @@
...
+    it('should return an empty array if no proposals are pending', async () => {
+      dsMockUtils.createQueryStub('multiSig', 'proposalIds', {
+        entries: [],
+      });
+
+      const result = await multiSig.getProposals();
+
+      const expectedProposals: MultiSigProposal[] = [];
+
+      expect(result).toEqual(expectedProposals);
```
prashantasdeveloper, line 141:
```suggestion
      expect(result).toEqual([]);
```

src/api/entities/__tests__/MultiSigProposal.ts
```diff
@@ -0,0 +1,159 @@
...
+
+      const result = await proposal.exists();
+      expect(result).toBe(false);
+    });
+  });
+
+  describe('method: toHuman', () => {
+    it('should return a human readable representation of the entity', () => {
+      const result = proposal.toHuman();
+      expect(result).toEqual('{"multiSigAddress":"someAddress","id":"1"}');
```
prashantasdeveloper, line 156: This will be updated once `toHuman` method is altered

src/testUtils/mocks/dataSources.ts
```diff
@@ -4200,3 +4218,54 @@ export const createMockAssetTransferCompliance = (
...
+  return createMockCodec(
+    {
+      args: createMockCodec(args, false),
+      method,
+      section,
+    },
+    !callArgs
+  ) as MockCodec<Call>;
+};
+
```
prashantasdeveloper, line 4246:
```suggestion
/**
 * @hidden
 * NOTE: `isEmpty` will be set to true if no value is passed
 */
```

prashantasdeveloper (changes requested):

src/api/entities/MultiSig/MultiSig.ts
```diff
@@ -0,0 +1,85 @@
...
+      context,
+      address,
+    } = this;
+    const rawAddress = stringToAccountId(address, context);
+
+    const rawProposals = await multiSig.proposalIds.entries(rawAddress);
+
+    return rawProposals.map(([, rawId]) => {
+      const id = u64ToBigNumber(rawId.unwrap());
+      return new MultiSigProposal({ multiSigAddress: address, id: new BigNumber(id) }, context);
```
prashantasdeveloper, line 82:
```suggestion
      return new MultiSigProposal({ multiSigAddress: address, id }, context);
```
`id` is already of type `BigNumber`

src/api/entities/MultiSig/MultiSigProposal.ts
```diff
@@ -0,0 +1,134 @@
...
+      throw new PolymeshError({
+        code: ErrorCode.DataUnavailable,
+        message: `Proposal with ID: "${id}" was not found. It may have already been executed`,
+      });
+    } else {
+      const value = proposal.unwrap();
+      args = value.args;
+      method = value.method;
+      section = value.section;
+    }
```
prashantasdeveloper, line 84:
```suggestion
    } 
    
    const { args, method, section } = proposal.unwrap();
```

src/api/entities/MultiSig/MultiSigProposal.ts
```diff
@@ -0,0 +1,134 @@
...
+        },
+      },
+      multiSigAddress,
+      id,
+      context,
+    } = this;
+    const rawId = bigNumberToU64(id, context);
+    const rawMultiSignAddress = stringToAccountId(multiSigAddress, context);
+    const rawProposal = await multiSig.proposals([rawMultiSignAddress, rawId]);
+    return !rawProposal.isEmpty;
```
prashantasdeveloper, line 120:
```suggestion
    return rawProposal.isSome;
```
I prefer going for the true value instead of !false

src/api/entities/__tests__/MultiSig.ts
```diff
@@ -114,19 +113,15 @@ describe('MultiSig class', () => {
     });
   });
 
-  describe('method: getPendingProposals', () => {
+  describe('method: getProposals', () => {
     const id = new BigNumber(1);
-    it('should get pending proposals', async () => {
+    it('should get proposals', async () => {
```
monitz87, line 118: Blank line here

src/api/procedures/__tests__/createMultiSig.ts
```diff
@@ -54,17 +53,17 @@ describe('createMultiSig procedure', () => {
     const mockAccount = entityMockUtils.getAccountInstance();
 
     const signers = [mockAccount];
-    const signaturesRequired = new BigNumber(1);
+    const requiredSignatures = new BigNumber(1);
 
     const signatories = signers.map(s => utilsConversionModule.signerToSignatory(s, mockContext));
     const convertedSignersRequired = utilsConversionModule.bigNumberToU64(
```
monitz87, line 59: rawRequiredSignatures

src/api/procedures/__tests__/createMultiSig.ts
```diff
@@ -54,17 +53,17 @@ describe('createMultiSig procedure', () => {
     const mockAccount = entityMockUtils.getAccountInstance();
 
     const signers = [mockAccount];
-    const signaturesRequired = new BigNumber(1);
+    const requiredSignatures = new BigNumber(1);
 
     const signatories = signers.map(s => utilsConversionModule.signerToSignatory(s, mockContext));
```
monitz87, line 58: (nit) rawSignatories. I think it's easier to identify at a glance which values are Codecs and which aren't if we prefix them with raw. Although to be honest I don't think `raw` is the correct word here, so you might want to think of a different prefix for the future and do a mass rename

monitz87: Btw, why aren't you using `dsMockUtils` for these?

src/api/procedures/createMultiSig.ts
```diff
@@ -36,16 +26,16 @@ export async function prepareCreateMultiSigAccount(
...
+  if (requiredSignatures.gt(signers.length)) {
     throw new PolymeshError({
       code: ErrorCode.ValidationError,
-      message: 'The number of signatures required should not exceed the number of signers',
+      message: 'The number of required signatures should not exceed the number of signers',
     });
   }
 
-  const rawSignaturesRequired = bigNumberToU64(signaturesRequired, context);
+  const rawSignaturesRequired = bigNumberToU64(requiredSignatures, context);
```
monitz87, line 38: rawRequiredSignatures

src/testUtils/mocks/dataSources.ts
```diff
@@ -4224,32 +4240,36 @@ export const createMockProposalData = (proposalData?: {
...
   };
   return createMockCodec(
     {
       approvals,
       rejections,
       status,
       expiry,
-      auto_close: autoClose,
+      autoClose,
     },
```
monitz87, line 4272: This should be

```
{
    approvals: createMockU64(approvals),
    rejections: createMockU64(rejections),
    status: createMockProposalStatus(status),
    autoClose: createMockBool(autoClose),
    expiry,
  }

**Review by monitz87 — CHANGES_REQUESTED** [review]

**Comment by polymath-eric** [comment]
The sonar cloud smell:

> This assertion is unnecessary since it does not change the type of the expression

Should be ignored, TS needs the assertion to not complain

File: src/api/entities/MultiSig/MultiSigProposal.ts [hunk 18]
```diff
@@ -0,0 +1,131 @@
...
+        rejections: rawRejections,
+        status: rawStatus,
+        expiry: rawExpiry,
+        autoClose: rawAutoClose,
+      },
+      proposal,
+    ] = await Promise.all([
+      multiSig.proposalDetail([rawMultiSignAddress, rawId]),
+      multiSig.proposals([rawMultiSignAddress, rawId]),
+    ]);
```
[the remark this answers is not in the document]
polymath-eric: I left this as Promise.all since the types and code is cleaner and the extra overhead should be minimal.

Thoughts @prashantasdeveloper ? I left a similar comment on asset metadata PR. Maybe we should have a rule like more than 3 calls need to go to multi.

I'd be more open to using multi more if we didn't have to assert all the types.

src/api/entities/MultiSigProposal.ts
```diff
@@ -0,0 +1,147 @@
...
+  multiSigAddress: string;
+  id: BigNumber;
+}
+
+export interface HumanReadable {
+  multiSigAddress: string;
+  id: string;
+}
+
+export type ProposalStatus =
```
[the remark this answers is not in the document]
polymath-eric: Going to look into splitting them in the morning

src/utils/conversion.ts
```diff
@@ -4069,7 +4069,26 @@ export function inputStatTypeToMeshStatType(
...
+  switch (type) {
+    case 'ActiveOrExpired':
+      if (!expiry || expiry > new Date()) {
+        return ProposalStatus.Active;
+      } else {
+        return ProposalStatus.Expired;
+      }
+    case 'Invalid':
+      return ProposalStatus.Invalid;
+    case 'ExecutionSuccessful':
```
prashantasdeveloper, line 4085: If you see its rare that `type` is used out of any enum. For conversion corresponding `.is<EnumValue>` is checked like `if (status.isActiveOrExpired)`

src/utils/conversion.ts
```diff
@@ -4069,7 +4069,26 @@ export function inputStatTypeToMeshStatType(
...
+      return ProposalStatus.Successful;
+    case 'ExecutionFailed':
+      return ProposalStatus.Failed;
+    case 'Rejected':
+      return ProposalStatus.Rejected;
+    default:
+      throw new UnreachableCaseError(type);
```
prashantasdeveloper, line 4092:
```suggestion
       throw new PolymeshError({
    code: ErrorCode.UnexpectedError,
    message: 'Unsupported Multisig Proposal Status. Please contact the Polymesh team',
    data: {
      auth: JSON.stringify(status, null, 2),
    },
  });
```

prashantasdeveloper (changes requested):

prashantasdeveloper:
The type assertion error reported by sonar isn't present in alpha. We can skip it for now and when we merge alpha to beta that will get resolved at that time

src/utils/conversion.ts
polymath-eric: The `UnreachableCaseError` takes an argument of type `never`. By using it we ensure all cases are checked, so if a new one is added we will get a compile error.

We still throw in case someone is using JavaScript, but in TypeScript, the compiler ensures it won't get triggered.

This is also why I checked the strings, but using the `isActiveOrExpired` doesn't narrow the type down of status, where as using the `type` will.

src/utils/conversion.ts
polymath-eric: The methods don't narrow the type down, which makes the UnrechableCaseError not work. It forces a non exhaustive match into a compile error.

prashantasdeveloper (approved):

PR #716, opened 2022-03-29
fix: 🐛 Expose non-proxied polkadot client

prashantasdeveloper:
#### JIRA Link

https://polymath.atlassian.net/browse/DA-244

### DIFF
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -438,7 +438,7 @@ export class Context {
   /**
    * @hidden
    *
-   * Retrieve the non-proxied Polkadot client
+   * Retrieve the polkadot.js promise client
    */
   public getPolymeshApi(): ApiPromise {
     return this._polymeshApi;
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -438,6 +438,15 @@ export class Context {
   /**
    * @hidden
    *
+   * Retrieve the non-proxied Polkadot client
+   */
+  public getPolymeshApi(): ApiPromise {
+    return this._polymeshApi;
+  }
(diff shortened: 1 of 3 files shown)


src/base/Context.ts
```diff
@@ -435,6 +435,15 @@ export class Context {
     return identity;
   }
 
+  /**
+   * @hidden
+   *
+   * Retrieve the non-proxied Polkadot client
```
polymath-eric, line 441: (nit)

```suggestion
   *  Retrieve the polkadot.js promise client
```

Very nit, but if someone is getting the "regular" version they shouldn't have to worry about what a proxy version is.

I had to look at the code to figure out what a Proxy meant in this context. I first thought proxy server where the connection is going through some jump box or VPN.

polymath-eric (commented):
LGTM. Small comment nit.

I am surprised sonar cloud gives a check mark at 0%. Is this because it knows there is no logic in the method?

I am not sure if we need the proxy even, but thats might be a larger discussion.

src/base/Context.ts
prashantasdeveloper: That makes sense. I have changed the docs as suggested.

polymath-eric (approved):
LGTM

prashantasdeveloper:
I was bit in dilemma whether to write a test case for a one liner with almost no logic. And the build also passed so I skipped it. Probably, for such logics sonar is also chill about it.

In all the procedures proxied client is used, which throws error if signing account is not present. That is why I didn't remove it.

VictorVicente (approved):

PR #724, opened 2022-04-07
feat: 🎸 Add createMultiSig procedure

polymath-eric:
Adds createMultiSig procedure into AccountManagement

### DIFF
--- a/src/AccountManagement.ts
+++ b/src/AccountManagement.ts
@@ -1,9 +1,10 @@
 import { MultiSig } from '~/api/entities/MultiSig';
-import { createMultiSigAccount, CreateMultiSigParams } from '~/api/procedures/createMultiSig';
 import {
   Account,
   AuthorizationRequest,
   Context,
+  createMultiSigAccount,
+  CreateMultiSigParams,
   inviteAccount,
   InviteAccountParams,
   leaveIdentity,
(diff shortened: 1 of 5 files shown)


polymath-eric:
I am not sure why sonar cloud think `this` is being unused when I copied the format from other procedures. It does seem weird to have a function parameter that shadows this.

I am tempted to mark it as not an issue since there seems to be precedence for this. Not sure if I am missing something though.

prashantasdeveloper (changes requested):
Left some nit comments about renaming and docs

prashantasdeveloper:
> I am not sure why sonar cloud think `this` is being unused when I copied the format from other procedures. It does seem weird to have a function parameter that shadows this.
> 
> I am tempted to mark it as not an issue since there seems to be precedence for this. Not sure if I am missing something though.

I have also wondered about why we use the `this` param, felt kind of odd. I'm also not sure, if something is wrong but could you try importing `MultiSig` from `~/internal` instead of `~/api/entites/MultiSig`? If that doesn't work maybe Jere can help us out here

prashantasdeveloper (approved):
LGTM. But still not sure why sonar is not giving the pass on the `this` thing.

polymath-eric:
@monitz87 let me know what you think about the code smell when you review this

PR #725, opened 2022-04-08
feat: 🎸 Add `getOne` method in `IdentityAuthorizations`

prashantasdeveloper:
#### JIRA Link

https://polymath.atlassian.net/browse/DA-252

#### Changelog

`IdentityAuthorizations.getOne` - This method overrides the `getOne` method from base class `Authorizations` and will fetch Authorization Request, targeting or issued by the parent Identity, by its ID

### DIFF
--- a/src/api/entities/Identity/IdentityAuthorizations.ts
+++ b/src/api/entities/Identity/IdentityAuthorizations.ts
@@ -1,7 +1,9 @@
+import BigNumber from 'bignumber.js';
+
 import { AuthorizationRequest, Authorizations, Identity } from '~/internal';
 import { PaginationOptions, ResultSet } from '~/types';
 import { QueryReturnType, tuple } from '~/types/utils';
-import { signatoryToSignerValue, stringToIdentityId } from '~/utils/conversion';
+import { bigNumberToU64, signatoryToSignerValue, stringToIdentityId } from '~/utils/conversion';
 import { requestPaginated } from '~/utils/internal';
 
 /**
@@ -50,5 +52,40 @@ export class IdentityAuthorizations extends Authorizations<Identity> {
       data,
       next,
     };
+  }
+
+  /**
+   * Retrieve a single Authorization Request targeting or issued by this Identity by its ID
+   *
+   * @throws if there is no Authorization Request with the passed ID targeting or issued by this Identity
+   */
+  public override async getOne(args: { id: BigNumber }): Promise<AuthorizationRequest> {
+    const {
+      context,
+      parent: { did },
+      context: {
+        polymeshApi: {
+          query: { identity },
+        },
+      },
+    } = this;
+
+    const { id } = args;
+
+    const rawId = bigNumberToU64(id, context);
+
+    const targetSignatory = await identity.authorizationsGiven(
+      stringToIdentityId(did, context),
+      rawId
+    );
+
+    if (!targetSignatory.isEmpty) {
+      const auth = await identity.authorizations(targetSignatory, rawId);
+      const signerValue = signatoryToSignerValue(targetSignatory);
+
+      return this.createAuthorizationRequests([{ auth: auth.unwrap(), target: signerValue }])[0];
+    }
+
+    return super.getOne({ id });
   }
 }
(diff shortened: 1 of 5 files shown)


polymath-eric (approved):
LGTM, left some small comments

src/api/entities/Identity/IdentityAuthorizations.ts
```diff
@@ -51,4 +53,41 @@ export class IdentityAuthorizations extends Authorizations<Identity> {
       next,
     };
   }
+
+  /**
+   * Retrieve a single Authorization Request targeting or issued by this Identity by its ID
+   *
+   * `authorizations` storage only returns results for the authorization target, so `authorizationsGiven` needs to be queried first to find the relevant target if present
```
monitz87, line 60: This seems more of an implementation detail that should be in a comment inside the method rather than in the JSDoc

src/api/entities/Identity/IdentityAuthorizations.ts
```diff
@@ -51,4 +53,41 @@ export class IdentityAuthorizations extends Authorizations<Identity> {
...
+    );
+
+    if (!targetSignatory.isEmpty) {
+      const auth = await identity.authorizations(targetSignatory, rawId);
+      const target = signatoryToSignerValue(targetSignatory);
+
+      return this.createAuthorizationRequests([{ auth: auth.unwrap(), target }])[0];
+    }
+
+    return super.getOne({ id });
```
monitz87, line 91: We can get a performance boost (if we want) by calling this before awaiting `identity.authorizationsGiven` and saving it into a variable so that it runs in parallel. You would probably have to call `defusePromise` on it in case it errors out before you return it

So

```
/*
 * We're using `defusePromise` here because if the id corresponds to a sent auth, we don't
 *   care about the error, and if it doesn't, then we return the promise and the error will be
 *   handled by the caller
 */
const gettingReceivedAuth = defusePromise(super.getOne({ id }));

// ... logic, fetch the sent promise and stuff

return gettingReceivedAuth;
```

Just a suggestion, if you think it hurts readability too much ignore it. We can worry about performance later on

src/api/entities/Identity/__tests__/IdentityAuthorizations.ts
```diff
@@ -143,4 +143,109 @@ describe('IdentityAuthorizations class', () => {
...
+      const result = await identityAuthorization.getOne({ id });
+
+      expect(result.authId).toEqual(id);
+      expect(result.expiry).toBeNull();
+      expect(result.data).toEqual(data);
+      expect((result.target as Identity).did).toEqual(targetDid);
+      expect(result.issuer.did).toEqual(did);
+    });
+
+    it('should return the requested Authorization Request targeted by the signing Identity', async () => {
```
monitz87, line 198:
```suggestion
    it('should return the requested Authorization Request targeting the parent Identity', async () => {
```

src/api/entities/Identity/__tests__/IdentityAuthorizations.ts
```diff
@@ -143,4 +143,109 @@ describe('IdentityAuthorizations class', () => {
...
+    afterAll(() => {
+      sinon.restore();
+    });
+
+    beforeAll(() => {
+      sinon.stub(utilsConversionModule, 'signerValueToSignatory');
+      sinon.stub(utilsConversionModule, 'bigNumberToU64');
+    });
+
+    it('should return the requested Authorization Request issued by the signing Identity', async () => {
```
monitz87, line 157:
```suggestion
    it('should return the requested Authorization Request issued by the parent Identity', async () => {
```

src/testUtils/mocks/dataSources.ts
```diff
@@ -1445,7 +1445,7 @@ const createMockEnum = (enumValue?: string | Record<string, Codec | Codec[]>): E
     codec[`as${upperFirst(key)}`] = enumValue[key];
   }
 
-  return createMockCodec(codec, false) as Enum;
+  return createMockCodec(codec, !enumValue) as Enum;
```
monitz87, line 1448: Why this change? I think it makes sense, but what triggered it?

monitz87 (changes requested):

src/testUtils/mocks/dataSources.ts
prashantasdeveloper: This was required to test isEmpty condition for the signatory. 

It was always giving `false` and I wanted it to be `true` while mocking the targetSignatory returned by `authorizationsGiven`, so that it can be verify that the `getOne` from `authorizations` is being called i

src/api/entities/Identity/IdentityAuthorizations.ts
prashantasdeveloper: I liked the approach and I don't think it hurts readability. The comments make it crystal clear.

monitz87 (approved):

PR #727, opened 2022-04-11
fix: 🐛 Fetch missing `block.hash` from middleware queries

prashantasdeveloper:
Added `hash` property within `block` to be fetched from middleware queries.
(While standardizing input and output data from middleware, block hash
was added to output params wherever block number was returned.
But appently, the middleware queries were not updated to fetch the same.)

### DIFF
--- a/src/middleware/queries.ts
+++ b/src/middleware/queries.ts
@@ -137,6 +137,7 @@ export function eventByIndexedArgs(
         extrinsic_idx
         block {
           datetime
+          hash
         }
       }
     }
@@ -180,6 +181,7 @@ export function eventsByIndexedArgs(
         extrinsic_idx
         block {
           datetime
+          hash
         }
       }
     }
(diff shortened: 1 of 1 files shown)


src/middleware/queries.ts
```diff
@@ -212,6 +214,9 @@ export function transactionByHash(
         success
         spec_version_id
         extrinsic_hash
+        block {
+          hash
```
VictorVicente, line 218: Is it ok to not include datetime too here?

src/middleware/queries.ts
```diff
@@ -354,6 +359,9 @@ export function transactions(
           success
           spec_version_id
           extrinsic_hash
+          block {
+            hash
```
VictorVicente, line 363: Same as above

src/middleware/queries.ts
prashantasdeveloper: Yeah. `datetime` param is not used for this query

src/middleware/queries.ts
prashantasdeveloper: Yeah. `datetime` param is not used for this query

VictorVicente (approved):

PR #730, opened 2022-04-12
feat: 🎸 Add modify MultiSig method

polymath-eric:
Allows the creator of a MultiSig to add and remove signers


src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+      transaction: tx.multiSig.removeMultisigSignersViaCreator,
+      args: [rawAddress, rawRemovedSigners],
+    });
+  }
+}
+
+/**
+ * @hidden
+ */
+export const modifyMultiSigAccount = (): Procedure<ModifyMultiSigParams, void> =>
```
monitz87, line 93: Should have the same name as the file

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+  new Procedure(prepareModifyMultiSig, {
+    permissions: {
+      transactions: [
+        TxTags.multiSig.AddMultisigSignersViaCreator,
+        TxTags.multiSig.RemoveMultisigSignersViaCreator,
+      ],
```
monitz87, line 99: This needs to be dynamic. Use storage to share the diffs

src/testUtils/mocks/entities.ts
```diff
@@ -1418,6 +1426,68 @@ const MockCustomPermissionGroupClass = createMockEntityClass<CustomPermissionGro
...
+     */
+    public argsToOpts(...args: ConstructorParameters<typeof MultiSig>) {
+      return extractFromArgs(args, ['address']);
+    }
+
+    /**
+     * @hidden
+     */
+    public configure(opts: Required<MultiSigOptions>) {
+      this.uuid = 'account';
```
monitz87, line 1454:
```suggestion
      this.uuid = 'multiSig';
```

src/testUtils/mocks/entities.ts
```diff
@@ -1418,6 +1426,68 @@ const MockCustomPermissionGroupClass = createMockEntityClass<CustomPermissionGro
...
+    checkPermissions: {
+      result: true,
+    },
+    details: {
+      signers: [],
+      signaturesRequired: new BigNumber(0),
+    },
+    getCreator: getIdentityInstance(),
+  }),
+  ['MultiSig']
```
monitz87, line 1488:
```suggestion
  ['Account', 'MultiSig']
```

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+  }
+
+  if (removedSigners.length > 0) {
+    const rawRemovedSigners = removedSigners.map(s => signerToSignatory(s, context));
+    this.addTransaction({
+      transaction: tx.multiSig.removeMultisigSignersViaCreator,
+      args: [rawAddress, rawRemovedSigners],
+    });
+  }
```
monitz87, line 87: We should use a batch when both adding and removing. I suggest having an array of transactions with their arguments, and if it turns up empty at the end, we throw the error (which by the way should be a `NoDataChange` error). If it has two elements, we use a batch, and if it has only one, we use the regular `addTransaction`

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const rawAddress = stringToAccountId(multi.address, context);
+  const { signers: currentSigners, signaturesRequired } = await multi.details();
+
+  if (signaturesRequired.gt(signers.length)) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'The number of signatures required should not exceed the number of signers',
```
monitz87, line 60:
```suggestion
      message: 'The number of required signatures should not exceed the number of signers',
```

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/**
+ * @hidden
+ */
+export async function prepareModifyMultiSig(
+  this: Procedure<ModifyMultiSigParams, void>,
+  args: ModifyMultiSigParams
+): Promise<void> {
+  // adds or removes signers and can change number of signatures required.
```
monitz87, line 34:
```suggestion
  // adds or removes signers
```

Also, shouldn't these comments go at the top of the function?

monitz87 (changes requested):

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+  } = this;
+  const { signers, multiSig: multi } = args;
+
+  // ensure calling account is creator, other actions require a proposal from the multisig
+  const signingIdentity = await context.getSigningIdentity();
+  const creator = await multi.getCreator();
+  if (!creator.isEqual(signingIdentity)) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'A MultiSig can only be modified by its creator',
```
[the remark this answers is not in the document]
polymath-eric: Do you mean put this check in the branch this is based off of?

Right now we can't perform any action to change the multiSig unless this is the case though. This check makes sense to me.

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * @hidden
+ */
+export async function prepareModifyMultiSig(
+  this: Procedure<ModifyMultiSigParams, void>,
+  args: ModifyMultiSigParams
+): Promise<void> {
+  // adds or removes signers and can change number of signatures required.
+  // Should also provide a "as owner flag" to allow for a user to "override" other signatures
```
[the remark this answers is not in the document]
monitz87: My idea is to add a `ProcedureOpts` property, something like `multiSigProposal: boolean`. There is a chance that some procedures will be only callable as proposals (as well as the inverse), so we should add some validation there

src/api/procedures/modifyMultiSig.ts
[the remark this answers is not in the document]
monitz87: In the future, we will support calling this procedure as a proposal instead of just as the creator. When we do, we will need to remove this error. I'm worried we might forget to do so

src/api/entities/MultiSig.ts
```diff
@@ -92,4 +116,36 @@ export class MultiSig extends Account {
...
+      });
+    }
+
+    const did = identityIdToString(rawCreatorDid);
+
+    return new Identity({ did }, context);
+  }
+
+  /**
+   * Modify the signers for the MultiSig. The signingAccount must belong to the Identity of the creator of the MultiSig
```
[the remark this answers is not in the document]
polymath-eric: What do you mean? I think when we do add proposal support, this method would stay as is. If a user wants to make a proposal they would make it like they would for any other proposal.

This method is using the `viaCreator` versions. We might need to expose `addMultiSigSigner` in its own procedure.

src/api/procedures/modifyMultiSig.ts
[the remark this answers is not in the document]
polymath-eric: I am using the transactions `tx.multiSig.addMultisigSignersViaCreator` and `tx.multiSig.removeMultisigSignersViaCreator`.

We might need to expose the non ViaCreator variations, but either that should get its own method, or we will need to modify this one, in which this error will be found.

src/api/entities/MultiSig/MultiSig.ts
```diff
@@ -64,6 +91,33 @@ export class MultiSig extends Account {
...
+    const {
+      context: {
+        polymeshApi: {
+          query: { multiSig },
+        },
+      },
+      context,
+    } = this;
+
+    const rawAddress = addressToKey(this.address, context);
```
prashantasdeveloper, line 107: `this.address` can also be destructed above

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,154 @@
+import BigNumber from 'bignumber.js';
+
+import { PolymeshError, Procedure } from '~/internal';
+import { ErrorCode, MultiSig, Signer, TxTags } from '~/types';
+import { ProcedureAuthorization } from '~/types/internal';
+import { signerToSignatory, signerToString, stringToAccountId } from '~/utils/conversion';
+import { checkTxType } from '~/utils/internal';
+
+/**
+ * @hidden
```
prashantasdeveloper, line 10: This should be `procedure/types` and not be hidden. If you want only `multiSig` in this can be set to hidden

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,154 @@
...
+    context,
+  } = this;
+  const { signers, multiSig: multi } = args;
+
+  const [signingIdentity, creator] = await Promise.all([
+    context.getSigningIdentity(),
+    multi.getCreator(),
+  ]);
+
+  if (signersToAdd.length === 0 && signersToRemove.length === 0) {
```
prashantasdeveloper, line 66:
```suggestion
  if (!signersToAdd.length && !signersToRemove.length) {
```
(nit) Can leave as it is as well

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,154 @@
...
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'The given signers are equal to the current signers',
```
prashantasdeveloper, line 69: How about we specify in the message that atleast one of the signers should be added/removed? (OK in its current form as well)

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,154 @@
...
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'The number of required signatures should not exceed the number of signers',
+    });
+  }
+
+  const transactions = [];
+
+  if (signersToAdd.length > 0) {
+    const rawAddedSigners = signersToAdd.map(s => signerToSignatory(s, context));
```
prashantasdeveloper, line 92:
```suggestion
    const rawAddedSigners = signersToAdd.map(signer => signerToSignatory(signer, context));
```
I think we avoid using single character variable names. Can change to `signer` or `newSigner` or `signerToAdd`

src/api/procedures/modifyMultiSig.ts
```diff
@@ -0,0 +1,154 @@
...
+    transactions.push(
+      checkTxType({
+        transaction: tx.multiSig.addMultisigSignersViaCreator,
+        args: [rawAddress, rawAddedSigners],
+      })
+    );
+  }
+
+  if (signersToRemove.length > 0) {
+    const rawRemovedSigners = signersToRemove.map(s => signerToSignatory(s, context));
```
prashantasdeveloper, line 102: Same single char variable changes applicable here as well

prashantasdeveloper (commented):
Left some nit comments

prashantasdeveloper (approved):
LGTM

src/api/entities/MultiSig.ts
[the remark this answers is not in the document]
monitz87: What I mean is that ideally we would have a single endpoint for modifying the signers, and the user should be able to call it as the creator, or as a proposal. The procedure should figure out whether to use `addMultisigSignersViaCreator` (and/or remove) or to batch a bunch of `addMultisigSigner` (and/or remove)

src/api/entities/MultiSig/MultiSig.ts
```diff
@@ -97,45 +93,46 @@ export class MultiSig extends Account {
       context,
       address,
     } = this;
-    const rawAddress = stringToAccountId(address, context);
 
-    const rawProposals = await multiSig.proposalIds.entries(rawAddress);
+    const rawAddress = addressToKey(address, context);
```
monitz87, line 97: Why not `stringToAccountId`?

src/api/entities/MultiSig/MultiSig.ts
```diff
@@ -97,45 +93,46 @@ export class MultiSig extends Account {
...
+    const rawCreatorDid = await multiSig.multiSigToIdentity(rawAddress);
+    if (rawCreatorDid.isEmpty) {
```
monitz87, line 99:
```suggestion
    const rawCreatorDid = await multiSig.multiSigToIdentity(rawAddress);
    
    if (rawCreatorDid.isEmpty) {
```

src/api/entities/MultiSig/MultiSig.ts
```diff
@@ -97,45 +93,46 @@ export class MultiSig extends Account {
...
-      });
-    }
-
-    const did = identityIdToString(rawCreatorDid);
+    const rawProposals = await multiSig.proposalIds.entries(rawAddress);
 
-    return new Identity({ did }, context);
+    return rawProposals.map(([, rawId]) => {
+      const id = u64ToBigNumber(rawId.unwrap());
+      return new MultiSigProposal({ multiSigAddress: address, id: new BigNumber(id) }, context);
```
monitz87, line 130:
```suggestion
      const id = u64ToBigNumber(rawId.unwrap());
      
      return new MultiSigProposal({ multiSigAddress: address, id }, context);
```

src/api/procedures/modifyMultiSig.ts
```diff
@@ -1,101 +1,141 @@
...
-  const creator = await multi.getCreator();
-  if (!creator.isEqual(signingIdentity)) {
+  const [signingIdentity, creator] = await Promise.all([
+    context.getSigningIdentity(),
+    multi.getCreator(),
+  ]);
+
+  if (signersToAdd.length === 0 && signersToRemove.length === 0) {
     throw new PolymeshError({
       code: ErrorCode.ValidationError,
```
monitz87, line 54: This should be `NoDataChange`

src/api/procedures/modifyMultiSig.ts
```diff
@@ -1,101 +1,141 @@
...
-  // adds or removes signers and can change number of signatures required.
-  // Should also provide a "as owner flag" to allow for a user to "override" other signatures
   const {
     context: {
       polymeshApi: { tx },
     },
+    storage: { signersToAdd, signersToRemove, requiredSignatures },
     context,
   } = this;
   const { signers, multiSig: multi } = args;
```
monitz87, line 45: Why the rename to `multi`?

src/api/procedures/modifyMultiSig.ts
```diff
@@ -1,101 +1,141 @@
...
-function calculateSignerDelta(current: Signer[], target: Signer[]): [Signer[], Signer[]] {
+function calculateSignerDelta(
+  current: Signer[],
+  target: Signer[]
+): Pick<Storage, 'signersToAdd' | 'signersToRemove'> {
   const currentSet = new Set(current.map(signerToString));
   const targetSet = new Set(target.map(signerToString));
 
   const newSigners = new Set([...target].filter(s => !currentSet.has(signerToString(s))));
   const removedSigners = new Set([...current].filter(s => !targetSet.has(signerToString(s))));
```
monitz87, line 26: Single letter variables

src/api/procedures/modifyMultiSig.ts
```diff
@@ -1,101 +1,141 @@
...
-    });
+  if (signersToRemove.length > 0) {
+    const rawRemovedSigners = signersToRemove.map(signer => signerToSignatory(signer, context));
+    transactions.push(
+      checkTxType({
+        transaction: tx.multiSig.removeMultisigSignersViaCreator,
+        args: [rawAddress, rawRemovedSigners],
+      })
+    );
   }
```
monitz87, line 96: These two blocks look similar enough that they might be refactored into a function

src/internal.ts
```diff
@@ -141,9 +141,14 @@ export { quitSubsidy, QuitSubsidyParams } from '~/api/procedures/quitSubsidy';
 export { modifyAllowance, ModifyAllowanceParams } from '~/api/procedures/modifyAllowance';
 export { createMultiSigAccount } from '~/api/procedures/createMultiSig';
 export {
-  SetAllowanceParams,
-  DecreaseAllowanceParams,
-  IncreaseAllowanceParams,
-} from '~/api/procedures/modifyAllowance';
-export { createMultiSigAccount, CreateMultiSigParams } from '~/api/procedures/createMultiSig';
-export { modifyMultiSigAccount, ModifyMultiSigParams } from '~/api/procedures/modifyMultiSig';
+  modifyMultiSig as modifyMultiSigAccount,
```
monitz87, line 144: Why the rename?

monitz87 (changes requested):

src/api/procedures/modifyMultiSig.ts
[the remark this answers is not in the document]
polymath-eric: To add on. If making a proposal, then the proposal should be for `multiSig.addMultiSigSigner`, and not for this method.

Either we make a new method for the proposal version, or we make this one use a dynamic extrinsic depending on the circumstance.

PR #733, opened 2022-04-19
chore: 🤖 Add websocket-as-promised dependency

prashantasdeveloper:
Added dependency `websocket-as-promised` required to update README.


VictorVicente (approved):

polymath-eric:
I think I removed websocket as promised when trying to remove the Github security warnings. I didn't realize it was needed for the README step.

It might have been ok as dev dependency but since its already merged probably OK to leave as is.

PR #737, opened 2022-04-21
feat: 🎸 Always return `TickerReservation` from `assets.getTickerReservation`

prashantasdeveloper:
### Description

Regardless of the state, `assets.getReservation` now returns a `TickerReservation` instance.

### Breaking Changes

🧨 `assets.isTickerAvailable` has been removed. 
To check ticker availability, directly get the `TickerReservation` instance and use `details.status` to check the availability

### JIRA Link

DA-142

### Checklist

### DIFF
--- a/src/Assets.ts
+++ b/src/Assets.ts
@@ -11,14 +11,8 @@ import {
   ReserveTickerParams,
   TickerReservation,
 } from '~/internal';
-import {
-  ErrorCode,
-  ProcedureMethod,
-  SubCallback,
-  TickerReservationStatus,
-  UnsubCallback,
-} from '~/types';
-import { stringToIdentityId, stringToTicker, tickerToString } from '~/utils/conversion';
+import { ErrorCode, ProcedureMethod } from '~/types';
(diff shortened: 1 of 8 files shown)


src/Assets.ts
```diff
@@ -149,32 +114,9 @@ export class Assets {
    */
   public async getTickerReservation(args: { ticker: string }): Promise<TickerReservation> {
```
polymath-eric, line 115: Should we keep this async? looks like the await got removed.

polymath-eric (approved):
LGTM

src/utils/internal.ts
```diff
@@ -1033,3 +1034,31 @@ export function assertExpectedChainVersion(nodeUrl: string): Promise<void> {
     };
   });
 }
+
+/**
+ * @hidden
+ *
+ * Validates a ticker value
+ */
+export function assertTickerValid(ticker: string) {
```
monitz87, line 1043: Missing return type

VictorVicente (approved):

PR #740, opened 2022-04-27
feat: 🎸 Allow passing of `nonce` as part of `ProcedureOptions`

prashantasdeveloper:
### Description

Add `nonce` parameter to `ProcedureOptions`

### JIRA Link

DA-154

### Checklist

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1289,8 +1289,10 @@ export interface ProcedureOpts {
    * nonce value for signing the transaction
    *
    * An {@link api/entities/Account!Account} can directly fetch its current nonce by calling {@link api/entities/Account!Account.getCurrentNonce | account.getCurrentNonce}. More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
+   *
+   * @note the passed value can be either the nonce itself or a function that returns the nonce. This allows, for example, passing a closure that increases the returned value every time it's called, or a function that fetches the nonce from the chain or something else
    */
-  nonce?: BigNumber | (() => BigNumber);
+  nonce?: BigNumber | Promise<BigNumber> | (() => BigNumber | Promise<BigNumber>);
 }
 
 export interface ProcedureMethod<
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1290,7 +1290,7 @@ export interface ProcedureOpts {
    *
    * An {@link api/entities/Account!Account} can directly fetch its current nonce by calling {@link api/entities/Account!Account.getCurrentNonce | account.getCurrentNonce}. More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
    *
-   * @note the passed value can be either the nonce itself or a function that returns the nonce. This allows, for example, passing a closure that increases the returned value every time it's called, or a function that fetches the nonce from the chain or something else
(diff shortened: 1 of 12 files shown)


polymath-eric (approved):
LGTM

monitz87 (commented):
I just realized that this won't work properly until we eliminate Transaction Queues. Every transaction in the queue would have the same nonce with this implementation. If we were to take the passed value and increment it for every tx in the queue, users trying to run multiple procedures at once wouldn't know how much to increment it by unless they prepare each one in sequence and count the amount of transactions, which seems like very bad UX

Once we eliminate TQs it becomes much easier

src/types/index.ts
```diff
@@ -1284,6 +1284,13 @@ export interface ProcedureOpts {
    * Account or address of a signing key to replace the current one (for this procedure only)
    */
   signingAccount?: string | Account;
+
+  /**
+   * nonce value for signing the transaction
+   *
+   * More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
+   */
+  nonce?: BigNumber;
```
polymath-eric, line 1293: We need to expose a method to get the nonce of an `Account`. I think a method like `currentNonce` there would be reasonable. I don't think there is a way to get that now, which makes this hard to use.

Then for this field we can link too that method instead about getting the current nonce for the signingAccount. 


Also, what do you think about having this be `BigNumber | () => BigNumber`  or even `() => BigNumber`. By giving a function a user could call `currentNonce` and make a closure and be able to increment it. All we'd have to do is call it and use the result.

For example:
```ts
  let nonce = 0
  const incNonce = (n) => () => (nonce += n);
  const addOne = incNonce(1);

  console.log(addOne());  // 1
  console.log(addOne()); // 2
  console.log(addOne()); // 3
  console.log(nonce); // 3
```

monitz87: Agree on everything. I would leave the option of passing a plain number as well as a function

src/types/index.ts
```diff
@@ -1288,9 +1288,9 @@ export interface ProcedureOpts {
   /**
    * nonce value for signing the transaction
    *
-   * More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
+   * An {@link api/entities/Account!Account} can directly fetch its current nonce by calling {@link api/entities/Account!Account.getCurrentNonce | account.getCurrentNonce}. More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
```
monitz87, line 1291: I would add a section explaining how to use it with a function

monitz87 (changes requested):

polymath-eric (approved):
LGTM. 

Sonar Cloud lists two smells BTW

src/types/index.ts
```diff
@@ -1289,8 +1289,10 @@ export interface ProcedureOpts {
    * nonce value for signing the transaction
    *
    * An {@link api/entities/Account!Account} can directly fetch its current nonce by calling {@link api/entities/Account!Account.getCurrentNonce | account.getCurrentNonce}. More information can be found at: https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-take-the-pending-tx-pool-into-account-in-my-nonce
+   *
+   * @note the passed value can be either the nonce itself or a function that returns the nonce. This allows, for example, passing a closure that increases the returned value every time it's called, or a function that fetches the nonce from the chain or something else
```
monitz87, line 1293:
```suggestion
   * @note the passed value can be either the nonce itself or a function that returns the nonce. This allows, for example, passing a closure that increases the returned value every time it's called, or a function that fetches the nonce from the chain or a different source
```

monitz87 (approved):

PR #742, opened 2022-04-29
Da 179/upgrade 5.0.0

polymath-eric:
### Description

Upgrades the SDK for the 5.0.0 chain. The chain metatdata has made some changes. Notably the chain types are prefixed with `PolymeshPrimitive` or `PalletSomePallet`. While this makes the code more verbose, it does eliminate a lot of snake case in the Codecs which cleans up the eslint disable comments. Overall I think it reads cleaner, and the long names don't bother me that much besides the ones the stutter (e.g. Compliance Manager ones)

### Breaking Changes

I think the external API hasn't changed, but would need to double check to be sure. For the most part the breaking changes are with the interaction with polkadot.js

### JIRA Link

https://polymath.atlassian.net/browse/DA-260
https://polymath.atlassian.net/browse/DA-179

### Checklist


package.json
```diff
@@ -12,11 +12,11 @@
     "generate:middleware-types": "graphql-codegen -r dotenv/config",
     "fetch-definitions": "cross-env node scripts/fetchDefinitions.js",
     "update-readme": "cross-env node scripts/updateReadme.js",
-    "generate:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --package polymesh-types --input ./src/polkadot",
-    "generate:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package polymesh-types --endpoint ws://localhost:9944 --output ./src/polkadot --strict",
+    "generate:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --endpoint metadata.json --package polymesh-types --input ./src/polkadot/",
+    "generate:meta": "ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --endpoint metadata.json --package polymesh-types --output ./src/polkadot --strict",
```
polymath-eric, line 16: It seemed like one of these commands didn't like the ws endpoint with the chain.

I uploaded a metatdata.json to use to drive: https://drive.google.com/file/d/1137jNjQZYdSNOkXtk42LYv8yD-kOkcHP/view?usp=sharing

I used: `curl -H "Content-Type: application/json" -d '{"id":"1", "jsonrpc":"2.0", "method": "state_getMetadata", "params":[]}' http://localhost:9933 > metadata.json` to get it.

package.json
```diff
@@ -12,11 +12,11 @@
...
     "generate:tags": "cross-env node scripts/generateTxTags.js",
     "generate:post-process": "cross-env node scripts/postProcessTypes.js",
-    "test": "jest --detectOpenHandles --coverage",
+    "test": "jest --detectOpenHandles",
```
polymath-eric, line 19: Will add back in, but maybe we should have a `no-cov` variation. I find the coverage report annoying when debugging a test.

scripts/generateTxTags.js
```diff
@@ -27,7 +28,7 @@ websocket.onmessage = message => {
   const registry = new TypeRegistry();
 
   const metadata = new Metadata(registry, JSON.parse(message.data).result);
-  const modules = metadata.asLatest.modules;
+  const modules = metadata.asLatest.pallets;
```
polymath-eric, line 31: There are other changes we should investigate: https://polymath.atlassian.net/browse/DA-305

src/api/entities/Venue/index.ts
```diff
@@ -122,11 +122,11 @@ export class Venue extends Entity<UniqueIdentifiers, string> {
       settlement.details(venueId),
     ]);
 
-    const { creator, venue_type: type } = venueInfo.unwrap();
+    const { creator, venueType: type } = venueInfo.unwrap();
 
     return {
       owner: new Identity({ did: identityIdToString(creator) }, context),
-      description: venueDetailsToString(details),
+      description: bytesToString(details),
```
polymath-eric, line 129: Somehow these lost the special type we had and are now bytes.

Would be nice to have the custom types.

src/api/entities/common/namespaces/Authorizations.ts
```diff
@@ -120,4 +124,35 @@ export class Authorizations<Parent extends Signer> extends Namespace<Parent> {
         return new AuthorizationRequest(args, context);
       });
   }
+
+  /**
+   * @hidden
+   * Needed to handle snake case from rpc request
```
polymath-eric, line 130: This took me a long time to figure out what was happening. I was in a circle of fixing things while breaking other things for longer than I'd like to admit.

This works, but really we should figure out why RPC stuff didn't get updated. I put the word `basic` in for copy pasta for snake case stuff.

I am hoping if we figure out the generate script stuff it will eliminate the need for these (more in conversion).

src/api/procedures/__tests__/setTransferRestrictions.ts
```diff
@@ -1,7 +1,8 @@
 import { u64 } from '@polkadot/types';
```
polymath-eric, line 1: This is less than 100% coverage

src/api/procedures/addTransferRestriction.ts
```diff
@@ -89,47 +118,94 @@ export async function prepareAddTransferRestriction(
...
+    context
+  );
 
-  const transactions = [
-    checkTxType({
-      transaction: statistics.addTransferManager,
-      args: [rawTicker, rawTransferManager],
-    }),
-  ];
+  // BTreeSets need to be sorted
```
polymath-eric, line 126: `sort` will corece to string and sort by alphabetical order.

This should be stable, but there has to be a better way for this, since without it the error is `invalid signature` which is misleading if you don't know the cause.

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -66,32 +76,59 @@ export async function prepareSetTransferRestrictions(
...
+        condition = { type, value: (r as CountTransferRestrictionInput).count };
+      } else {
+        condition = { type, value: (r as PercentageTransferRestrictionInput).percentage };
+      }
+
+      const rawCondition = transferRestrictionToPolymeshTransferCondition(condition, context);
+
+      if (r.exemptedIdentities) {
+        r.exemptedIdentities.forEach(e => {
+          exemptions.push(e);
```
polymath-eric, line 123: This isn't 100% accurate, since exemptions are based on type, and don't reference a particular condition.

That said, I think with claim scopes this might make more sense so I am not sure if we should change it.

Screen Shot 2022-04-28 at 10 29 34 PM

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -66,32 +76,59 @@ export async function prepareSetTransferRestrictions(
...
+        });
+      }
+
+      return rawCondition;
+    })
+    .sort() as BTreeSetTransferCondition;
```
polymath-eric, line 129: Maybe the insides should be in storage? I don't like this cast so far away from the variable its affecting

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -100,41 +137,77 @@ export async function prepareSetTransferRestrictions(
...
+      transaction: statistics.setAssetTransferCompliance,
+      args: [{ Ticker: rawTicker }, conditions],
+    })
   );
 
-  this.addBatchTransaction({ transactions });
+  if (exemptions.length) {
+    const exemptedIds = await getExemptedIds(exemptions, context, ticker);
+    const exemptedScopeIds = exemptedIds.map(entityId => stringToIdentityId(entityId, context));
+    const btreeIds = scopeIdsToBtreeSetIdentityId(exemptedScopeIds, context);
```
polymath-eric, line 199: I sort inside here, so might be better to have other BTree methods use a similar pattern.

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -196,121 +275,37 @@ export async function prepareStorage(
...
-      );
-    });
     currentPercentageRestrictions.forEach(({ exemptedIds = [], percentage: value }) => {
       const restriction = { type: TransferRestrictionType.Percentage, value };
       currentRestrictions.push(restriction);
       currentExemptions.push(tuple(restriction, exemptedIds));
     });
   }
 
-  const toAddExemptions = await P.map(
```
polymath-eric, line 235: Semantics are now set on the extrinsics, so we don't need all this logic

src/base/Context.ts
```diff
@@ -1158,12 +1155,14 @@ export class Context {
...
   }
 
+  /* eslint-disable @typescript-eslint/no-explicit-any */
   /**
    *  @hidden
    *
    * Creates an instance of a type as registered in the polymeshApi instance
    */
-  public createType<K extends keyof InterfaceTypes>(type: K, params: unknown): InterfaceTypes[K] {
+  public createType<K extends keyof InterfaceTypes>(type: K, params: unknown): any {
```
polymath-eric, line 1164: They made a change that was supposed to make the createType API smarter, but now `InterfaceTypes[K]` exceeds TypeScript stack size, and might be infinite recursion even.

polymath-eric (commented):
I found this to be a pretty challenging upgrade. I had to fill in some holes in my knowledge with how polkadot schemas work.

Test coverage isn't perfect, and I think some conversion and mock functions need to be touched up in their params, but overall this seems to be working.

I'll pick this up when I am back Thursday, but if someone else picks it up and has questions feel free to Slack me.

package.json
monitz87: What was the specific problem with using the endpoint?

polymath-eric: It kept on thinking it was a file

`Error: Cannot find module '/Users/eric/polymesh-sdk/ws:/localhost:9944'`

polymath-eric: Thats with:

```
  "generate:defs": "ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --endpoint  metadata.json --package polymesh-types --input ./src/polkadot/",
    ```
    
    I had to check again since the error stripped a / it seems. the meta one seems fine though

File: src/utils/conversion.ts [hunk 13]
```diff
@@ -958,25 +994,85 @@ export function extrinsicPermissionsToTransactionPermissions(
...
-      const moduleName = stringLowerFirst(textToString(palletName));
+    pallets.forEach(({ palletName, dispatchableNames }) => {
+      const moduleName = stringLowerFirst(palletName.toString());
 
       if (dispatchableNames.isExcept) {
         const dispatchables = dispatchableNames.asExcept;
-        exceptions = [...exceptions, ...dispatchables.map(name => formatTxTag(name, moduleName))];
+        exceptions = [
+          ...exceptions,
+          ...dispatchables.map(name => formatTxTag(name.toString(), moduleName)),
```
monitz87, line 1009: bytesToString (also the next one)

src/utils/conversion.ts
```diff
@@ -2613,8 +2942,8 @@ export function venueTypeToMeshVenueType(type: VenueType, context: Context): Mes
 /**
  * @hidden
  */
-export function stringToVenueDetails(details: string, context: Context): VenueDetails {
-  return context.createType('VenueDetails', details);
+export function stringToVenueDetails(details: string, context: Context): Bytes {
```
monitz87, line 2945: Shouldn't this go in favor of `stringToBytes`?

src/utils/conversion.ts
```diff
@@ -3351,11 +3697,62 @@ export function distributionToDividendDistributionParams(
...
+  key2: PolymeshPrimitivesStatisticsStat2ndKey,
+  value: u128,
+  context: Context
+): PolymeshPrimitivesStatisticsStatUpdate {
+  return context.createType('PolymeshPrimitivesStatisticsStatUpdate', { key2, value });
+}
+
+type Stat = {
+  type: 'Count' | 'Balance';
+};
```
monitz87, line 3731: This should be in the internal types file. Also, magic strings

src/utils/conversion.ts
```diff
@@ -3351,11 +3697,62 @@ export function distributionToDividendDistributionParams(
...
-    // eslint-disable-next-line @typescript-eslint/naming-convention
-    local_id: bigNumberToU32(localId, context),
+    localId: bigNumberToU32(localId, context),
   });
 }
+
+/**
+ * @hidden
+ */
+export function primitiveStatisticsStatType(
```
monitz87, line 3711: There are a few new utils like this one that don't follow the naming convention

monitz87 (changes requested):
GJ overall, this was a massive endeavor

src/api/entities/Asset/Compliance/Requirements.ts
```diff
@@ -1,6 +1,10 @@
 import { QueryableStorageEntry } from '@polkadot/api/types';
 import { Vec } from '@polkadot/types/codec';
-import { AssetCompliance, AssetComplianceResult, TrustedIssuer } from 'polymesh-types/types';
+import {
+  PolymeshPrimitivesComplianceManagerAssetCompliance,
+  PolymeshPrimitivesConditionTrustedIssuer,
+} from '@polkadot/types/lookup';
+import { AssetComplianceResult } from 'polymesh-types/types';
```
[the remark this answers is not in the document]
polymath-eric: After some experimenting it seems like we should be able to take out the types, but we'll need to figure out CountryCode generate section since that needs types.

We'll also need to go through all the createType calls and make sure none are using the old types either.

src/api/entities/common/namespaces/Authorizations.ts
polymath-eric: The schema does play a role, but the actual types used are coming from `augment-api-rpc` and `augment-api-query`.

I think its weird the RPC one looks at `polymesh-types/polymesh` for its imports, the query one gets them from `@polkadot/types/lookup`.

There might be a reason for the divergence, but I'd think we can transform the rpc file like you are suggesting.

polymath-eric: In that file I changed `Authorization` to `PolymeshPrimitivesAuthorization` by hand and that seemed to work.

I think changing that file is the right approach. I am going to give that a shot

src/api/entities/Asset/Compliance/Requirements.ts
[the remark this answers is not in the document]
polymath-eric: I added in a script that deals with `Authorization` and that seems to work.

The `Result` ones seem to live in `src/polkadot/polymesh/types.ts`. I will take a look in the morning to see if they can get similar treatment.

src/api/procedures/setTransferRestrictions.ts
polymath-eric: https://polymath.atlassian.net/browse/DA-317

src/api/entities/Asset/Compliance/Requirements.ts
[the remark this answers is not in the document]
polymath-eric: Seems like the script isn't good after all. The types and tests are happy, but when I tried it out it seems like the snake case type is returned (undefined errors). Using snake case for the field resolves it, so for now I am going to switch back to having another method for handling RPC authorization stuff.

package.json
polymath-eric: Seems like they fixed it in a May 8th release. The note says `Deupe wss:// handling in polkadot-types-from-{chain, defs}`. I tried it out and web socket seems to work.

However, it seems to change some more types and has more breaking changes in it for us, even though its a minor version. I wan to get this merged in before making more changes.

monitz87: Created a task to track this

src/api/entities/Asset/Compliance/Requirements.ts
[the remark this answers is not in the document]
monitz87: The country code stuff should be able to be generated from the metadata. I'll create a task for that

src/api/procedures/addTransferRestriction.ts
```diff
@@ -89,47 +115,94 @@ export async function prepareAddTransferRestriction(
...
+        args: [{ Ticker: rawTicker }, currentStats],
+      })
+    );
+
+    // If the stat restriction is a Count the actual value needs to be set. This is due to the potentially slow transaction of counting all holders for the chain
+    if (type === TransferRestrictionType.Count) {
+      // if an asset has many investors this could be slow and instead should be fetched from SubQuery
+      // These should happen near the assets inception, so for now query the chain directly
+      const holders = await query.asset.balanceOf.entries(rawTicker);
+      const holderCount = new BigNumber(holders.length);
```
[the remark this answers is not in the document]
monitz87: I realized we're not taking PUIS into account here, which is a major problem. We need to figure out how to calculate unique investors when PUIS is enabled. I'll create a task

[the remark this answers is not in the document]
monitz87: In any case, we should encapsulate this logic into a function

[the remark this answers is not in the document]
monitz87: There's another problem. If the asset has the stat enabled but it hasn't been updated, `investorCount` will return a misleading value. We need to address this somehow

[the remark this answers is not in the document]
polymath-eric: Whoever makes the stat has to initialize it correctly. After its made the stat will get updated as part of the chain operation though. It shouldn't need constant updating.

There is a risk the stat gets misconfigured, but I am not sure what we can do about that if we call out to the chain.

I think a better approach would be to get this stuff into SubQuery and query that for the stats.

src/utils/conversion.ts
```diff
@@ -3756,3 +3688,14 @@ export function primitiveOpType(
...
+
+/**
+ * @hidden
+ */
+export function complianceRequirementsToBtreeSet(
+  conditions: PolymeshPrimitivesTransferComplianceTransferCondition[]
+): BTreeSetTransferCondition {
+  // The chain expects inputs to be sorted. Copy to avoid mutating input
+  const sorted = [...conditions].sort();
+  return sorted as BTreeSetTransferCondition;
```
monitz87, line 3700: Can't this be done via `createType`?

src/utils/conversion.ts
```diff
@@ -3726,23 +3652,29 @@ export function statUpdate(
...
  */
-export function meshStatToStat(rawStat: PolymeshPrimitivesStatisticsStatType): Stat {
-  return {
-    type: rawStat.op.type,
-  };
+export function statUpdatesToBtreeStatUpdate(
+  statUpdates: PolymeshPrimitivesStatisticsStatUpdate[]
+): BTreeSetStatUpdate {
+  const sorted = [...statUpdates];
+  return sorted as BTreeSetStatUpdate;
```
monitz87, line 3662: Same here

monitz87 (changes requested):

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -64,6 +64,76 @@ export interface Storage {
...
+      isSameCondition(transferCondition, r, type);
+    if (!someDifference) {
+      someDifference = !currentRestrictions.find(compareConditions);
+    }
+    let condition: TransferRestriction;
+    if (type === TransferRestrictionType.Count) {
+      condition = { type, value: (r as CountTransferRestrictionInput).count };
+    } else {
+      condition = { type, value: (r as PercentageTransferRestrictionInput).percentage };
+    }
```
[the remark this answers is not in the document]
polymath-eric: I see what you mean, but the `SetCountTransferRestrictionsParams` has restrictions as an Array, where as I have only a single value I am feeding in, so it doesn't quite work.

I was able to change this a bit to eliminate one of the assertions.

What I really want is for TS to let me do something like

`const value = inputCondition.count || inputCondition.percentage`, since I know one of them will be defined.

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -1,32 +1,32 @@
 import {
   BTreeSetStatType,
-  BTreeSetStatUpdate,
-  BTreeSetTransferCondition,
+  PolymeshPrimitivesTransferComplianceTransferCondition,
 } from '@polkadot/types/lookup';
 import BigNumber from 'bignumber.js';
 import { TransferCondition, TxTag, TxTags } from 'polymesh-types/types';
```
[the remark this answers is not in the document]
polymath-eric: I am not sure what you mean. Why not from here?

src/utils/conversion.ts
```diff
@@ -343,7 +334,7 @@ export function stringToInvestorZKProofData(proof: string, context: Context): In
  * @hidden
  */
 export function dateToMoment(date: Date, context: Context): Moment {
-  return context.createType('Moment', date.getTime());
+  return context.createType('Moment', date.getTime()) as Moment;
```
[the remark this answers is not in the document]
polymath-eric: The error is:

```
(parameter) context: Context
Property '__UIntType' is missing in type 'AbstractInt' but required in type 'u64'.
```

So seems like createType is giving a more general type as the result.

u64 seems to work, but will look into this for a bit.

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -64,6 +64,76 @@ export interface Storage {
...
+): [PolymeshPrimitivesTransferComplianceTransferCondition[], (string | Identity)[]] {
+  const exemptions: (string | Identity)[] = [];
+
+  let someDifference = restrictions.length !== currentRestrictions.length;
+  const conditions = restrictions.map(r => {
+    const compareConditions = (transferCondition: TransferCondition) =>
+      isSameCondition(transferCondition, r, type);
+    if (!someDifference) {
+      someDifference = !currentRestrictions.find(compareConditions);
+    }
```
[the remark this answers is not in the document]
monitz87: I incorrectly assumed they were in the same order

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -199,7 +231,7 @@ export async function prepareSetTransferRestrictions(
       checkTxType({
         transaction: statistics.setEntitiesExempt,
         feeMultiplier: new BigNumber(exemptions.length),
-        args: [true, { asset: { Ticker: rawTicker }, op }, btreeIds],
+        args: [true, { asset: tickerKey, op }, btreeIds],
```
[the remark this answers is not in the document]
monitz87: You're right. It's a struct. I'd create a conversion util for this

scripts/fetchMetadata.sh
```diff
@@ -0,0 +1 @@
+curl -H "Content-Type: application/json" -d '{"id":"1", "jsonrpc":"2.0", "method": "state_getMetadata", "params":[]}' http://localhost:9933 > metadata.json
```
monitz87, line 1:
```suggestion
curl -H "Content-Type: application/json" -d '{"id":"1", "jsonrpc":"2.0", "method": "state_getMetadata", "params":[]}' http://localhost:9933 > metadata.json

```

src/api/procedures/setTransferRestrictions.ts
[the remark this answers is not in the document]
monitz87: What you can do is 

```
let value: BigNumber;

if ('count' in inputCondition) {
  value = inputCondition.count;
} else {
  value = inputCondition.percentage;
}
```

You could stick this in a ternary as well but I'd rather not

monitz87 (approved):

PR #743, opened 2022-05-04
fix: 🐛 Add missing validations in `controllerTransfer` proc

prashantasdeveloper:
### Description

`controllerTransfer` procedure used to throw the following error when trying to transfer the Asset to a self Portfolio - 
```
asset.SenderSameAsReceiver:  Transfers to self are not allowed
```
This PR add the missing validation -> checks if the originPortfolio DID is same as the signingIdentity DID, it throws `ErrorCode.UnmetPrerequisite` with apt message

### JIRA Link

DA-304

### Checklist

### DIFF
--- a/src/api/procedures/controllerTransfer.ts
+++ b/src/api/procedures/controllerTransfer.ts
@@ -46,6 +46,15 @@ export async function prepareControllerTransfer(
 
   const originPortfolioId = portfolioLikeToPortfolioId(originPortfolio);
 
+  const { did } = await context.getSigningIdentity();
+
+  if (did === originPortfolioId.did) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Transfers to self are not allowed',
+    });
+  }
+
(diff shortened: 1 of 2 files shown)


src/api/procedures/controllerTransfer.ts
```diff
@@ -46,6 +46,15 @@ export async function prepareControllerTransfer(
 
   const originPortfolioId = portfolioLikeToPortfolioId(originPortfolio);
 
+  const { did } = await context.getSigningIdentity();
+
+  if (did === originPortfolioId.did) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Transfers to self are not allowed',
```
monitz87, line 54: I'd change this to "Controller transfers to self are not allowed" just to be super explicit. Also, to avoid fetching the signing identity twice, you should add a `prepareStorage` method and fetch it there

monitz87 (changes requested):

monitz87 (approved):

PR #745, opened 2022-05-05
feat: 🎸 Return gasFees, protocolFees in `getTransactionByHash`

prashantasdeveloper:
### Description

Added fee related params in extrinsic data returned by `network.getTransactionByHash`

### Breaking Changes

🧨 Return type of `network.getTransactionByHash` is changed from `Promise<ExtrinsicData | null>` to `Promise<ExtrinsicDataWithFees | null>`

### JIRA Link

DA-197

### Checklist

### DIFF
--- a/src/Network.ts
+++ b/src/Network.ts
@@ -5,7 +5,7 @@ import { eventByIndexedArgs, eventsByIndexedArgs, transactionByHash } from '~/mi
 import { EventIdEnum as EventId, ModuleIdEnum as ModuleId, Query } from '~/middleware/types';
 import {
   EventIdentifier,
-  ExtrinsicData,
+  ExtrinsicDataWithFees,
   NetworkProperties,
   ProcedureMethod,
   SubCallback,
@@ -15,9 +15,11 @@ import {
 import { Ensured } from '~/types/utils';
 import { TREASURY_MODULE_ADDRESS } from '~/utils/constants';
 import {
+  balanceToBigNumber,
   extrinsicIdentifierToTxTag,
   middlewareEventToEventIdentifier,
   moduleAddressToString,
+  stringToBlockHash,
   textToString,
   u32ToBigNumber,
 } from '~/utils/conversion';
@@ -228,8 +230,20 @@ export class Network {
    *
    * @note uses the middleware
    */
-  public async getTransactionByHash(opts: { txHash: string }): Promise<ExtrinsicData | null> {
-    const { context } = this;
+  public async getTransactionByHash(opts: {
+    txHash: string;
+  }): Promise<ExtrinsicDataWithFees | null> {
+    const {
+      context: {
+        polymeshApi: {
+          rpc: {
+            chain: { getBlock },
+            payment: { queryInfo },
+          },
+        },
+      },
+      context,
+    } = this;
 
     const { txHash: transactionHash } = opts;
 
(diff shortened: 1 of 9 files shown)


src/types/index.ts
```diff
@@ -330,6 +330,11 @@ export interface ExtrinsicData {
   extrinsicHash: string;
 }
 
+export interface ExtrinsicDataWithFees extends ExtrinsicData {
```
polymath-eric, line 333: Declaring the type here is OK, 

ideally we'd be able to provide the fees for all Extrinsic, but it seems like `Account.getTransactionHistory` populates `ExtrinsicData` from middleware, which doesn't have it.

It would be nice if SubQuery would have these fees, and then we could get away with only `ExtrinsicData`

polymath-eric (approved):
LGTM

src/types/index.ts
prashantasdeveloper: > ideally we'd be able to provide the fees for all Extrinsic, but it seems like Account.getTransactionHistory populates ExtrinsicData from middleware, which doesn't have it.

Exactly my point. Lets see what @monitz87 has to say here

monitz87: Can't they be included in subquery?

For now, I think extending `getTransactionHistory` to fetch fees makes sense so that we can return the same interface from both methods. For that one, you can use `entriesAt`, which returns entries for a query at a specific block

monitz87: If we go this route, I guess `getProtocolFees` should be extended to fetch for a set of tags using `entriesAt`. You would get all existing protocol fees, which shouldn't be too many, and filter them using the passed tags (any tags not present are considered to have 0 fees)

monitz87 (changes requested):

monitz87 (changes requested):

src/Network.ts
```diff
@@ -258,19 +272,36 @@ export class Network {
...
+
+      const rawBlockHash = stringToBlockHash(blockHash, context);
+
+      const {
+        block: { extrinsics },
+      } = await getBlock(rawBlockHash);
+
+      const [{ partialFee }, protocolFees] = await Promise.all([
+        queryInfo(extrinsics[extrinsicIdx].toHex(), rawBlockHash),
+        this.getProtocolFees({ tag: txTag }),
```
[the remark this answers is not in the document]
prashantasdeveloper: How do we get calculate fees for a particular block?  Currently, it is calculated as `baseFees * coefficient` which are both just fetched directly from protocolFee

src/testUtils/mocks/dataSources.ts
```diff
@@ -1468,20 +1468,22 @@ function isOption<T extends Codec>(codec: any): codec is Option<T> {
...
 }
 
+type MockCodec<C extends Codec> = C & { eq: sinon.SinonStub };
+
 /**
  * @hidden
  */
-const createMockCodec = (codec: unknown, isEmpty: boolean): Codec => {
-  const clone = cloneDeep(codec) as Mutable<Codec>;
+const createMockCodec = (codec: unknown, isEmpty: boolean): MockCodec<Codec> => {
```
monitz87, line 1476: You forgot to actually add the`eq` stub. Also, remove it from `createMockStringCodec` because it would be redundant (I suspect your tests passed because it was there)

monitz87 (changes requested):

PR #747, opened 2022-05-16
feat: 🎸 DA-311 Add support for middleware v2 with claims POC

prashantasdeveloper:
### Description

* Add support for middleware V2
* POC of claims pagination

### JIRA Link

DA-311

### Checklist

### DIFF
--- a/src/Claims.ts
+++ b/src/Claims.ts
@@ -1,5 +1,5 @@
 import BigNumber from 'bignumber.js';
-import { filter, isEqual, uniqBy, uniqWith } from 'lodash';
+import { filter, flatten, isEqual, uniqBy, uniqWith } from 'lodash';
 
 import {
   addInvestorUniquenessClaim,
@@ -10,7 +10,9 @@ import {
   ModifyClaimsParams,
 } from '~/internal';
 import { didsWithClaims, issuerDidsWithClaimsByTarget } from '~/middleware/queries';
+import { claims as claimsData, claimTargets } from '~/middleware/queriesV2';
 import { ClaimTypeEnum, Query } from '~/middleware/types';
+import { Query as QueryV2 } from '~/middleware/types-v2';
 import {
   CddClaim,
   ClaimData,
@@ -26,10 +28,12 @@ import {
 } from '~/types';
 import { ClaimOperation } from '~/types/internal';
 import { Ensured } from '~/types/utils';
+import { DEFAULT_GQL_PAGE_SIZE } from '~/utils/constants';
 import {
   scopeToMiddlewareScope,
   signerToString,
   toIdentityWithClaimsArray,
+  toIdentityWithClaimsArrayV2,
 } from '~/utils/conversion';
 import { calculateNextKey, createProcedureMethod, getDid, removePadding } from '~/utils/internal';
 
@@ -221,6 +225,103 @@ export class Claims {
 
     const count = new BigNumber(totalCount);
     const data = toIdentityWithClaimsArray(didsWithClaimsList, context);
+    const next = calculateNextKey(count, size, start);
+
+    return {
+      data,
+      next,
+      count,
+    };
+  }
+
+  /**
+   * Retrieve a list of Identities with claims associated to them. Can be filtered using parameters
+   *
+   * @param opts.targets - Identities (or Identity IDs) for which to fetch claims (targets). Defaults to all targets
+   * @param opts.trustedClaimIssuers - Identity IDs of claim issuers. Defaults to all claim issuers
+   * @param opts.scope - scope of the claims to fetch. Defaults to any scope
+   * @param opts.claimTypes - types of the claims to fetch. Defaults to any type
+   * @param opts.includeExpired - whether to include expired claims. Defaults to true
+   * @param opts.size - page size
+   * @param opts.start - page offset
+   *
+   * @note supports pagination
+   * @note uses the middleware V2
+   */
+  public async getIdentitiesWithClaimsV2(
+    opts: {
+      targets?: (string | Identity)[];
+      trustedClaimIssuers?: (string | Identity)[];
+      scope?: Scope;
+      claimTypes?: Exclude<ClaimType, ClaimType.InvestorUniquenessV2>[];
+      includeExpired?: boolean;
+      size?: BigNumber;
+      start?: BigNumber;
+    } = {}
+  ): Promise<ResultSet<IdentityWithClaims>> {
+    const { context } = this;
+
+    const {
+      targets,
+      trustedClaimIssuers,
+      scope,
+      claimTypes,
+      includeExpired = true,
+      size = new BigNumber(DEFAULT_GQL_PAGE_SIZE),
+      start = new BigNumber(0),
+    } = opts;
+
+    let targetIssuers;
+
+    const filters = {
+      scope: scope ? scopeToMiddlewareScope(scope) : undefined,
+      trustedClaimIssuers: trustedClaimIssuers?.map(trustedClaimIssuer =>
+        signerToString(trustedClaimIssuer)
+      ),
+      claimTypes: claimTypes?.map(ct => ClaimTypeEnum[ct]),
+      includeExpired,
+    };
+
+    if (!targets) {
+      const targetResults = await context.queryMiddlewareV2<Ensured<QueryV2, 'claims'>>(
+        claimTargets({
+          ...filters,
+        })
+      );
+
+      const {
+        data: { claims },
+      } = targetResults;
+
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const groupedTargets = claims!.groupedAggregates!;
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      targetIssuers = flatten(groupedTargets.map(groupedAggregate => groupedAggregate.keys!));
+    } else {
+      targetIssuers = targets.map(target => signerToString(target));
+    }
+
+    // note: pagination count is based on the target issuers and not the claims count
+    const count = new BigNumber(targetIssuers.length);
+
+    targetIssuers = targetIssuers.sort().slice(start.toNumber(), size.plus(start).toNumber());
+
+    const result = await context.queryMiddlewareV2<Ensured<QueryV2, 'claims'>>(
+      claimsData({
+        dids: targetIssuers,
+        ...filters,
+      })
+    );
+
+    const {
+      data: { claims },
+    } = result;
+
+    const data = toIdentityWithClaimsArrayV2(
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      claims!.nodes!.map(node => node!),
+      context
+    );
     const next = calculateNextKey(count, size, start);
 
     return {
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -27,6 +27,7 @@ import {
 import { Account, Asset, DividendDistribution, Identity, PolymeshError, Subsidy } from '~/internal';
 import { didsWithClaims, heartbeat } from '~/middleware/queries';
 import { ClaimTypeEnum, Query } from '~/middleware/types';
+import { Query as QueryV2 } from '~/middleware/types-v2';
 import {
   AccountBalance,
   ArrayTransactionArgument,
@@ -76,6 +77,7 @@ import { assertAddressValid, calculateNextKey, createClaim } from '~/utils/inter
 interface ConstructorParams {
   polymeshApi: ApiPromise;
   middlewareApi: ApolloClient<NormalizedCacheObject> | null;
+  middlewareApiV2: ApolloClient<NormalizedCacheObject> | null;
   ss58Format: BigNumber;
   signingManager?: SigningManager;
 }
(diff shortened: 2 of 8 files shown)


src/Claims.ts
```diff
@@ -230,6 +234,103 @@ export class Claims {
...
+    const filters = {
+      scope: scope ? scopeToMiddlewareScope(scope) : undefined,
+      trustedClaimIssuers: trustedClaimIssuers?.map(trustedClaimIssuer =>
+        signerToString(trustedClaimIssuer)
+      ),
+      claimTypes: claimTypes?.map(ct => ClaimTypeEnum[ct]),
+      includeExpired,
+    };
+
+    if (!targets) {
```
prashantasdeveloper, line 285: We fetch the targets first and once we have the list of targets, we fetch all the claims associated with those targets

src/Claims.ts
```diff
@@ -230,6 +234,103 @@ export class Claims {
...
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      targetIssuers = flatten(groupedTargets.map(groupedAggregate => groupedAggregate.keys!));
+    } else {
+      targetIssuers = targets.map(target => signerToString(target));
+    }
+
+    // note: pagination count is based on the target issuers and not the claims count
+    const count = new BigNumber(targetIssuers.length);
+
+    targetIssuers = targetIssuers.sort().slice(start.toNumber(), size.plus(start).toNumber());
```
prashantasdeveloper, line 307: Tooling gql does pagination based on sorted target issuers. Hence the explicit `sort()` function (as graphql doesn't sort the final data)

src/middleware/queriesV2.ts
```diff
@@ -0,0 +1,117 @@
...
+function createClaimsFilters(variables: QueryDidsWithClaimsArgs): {
+  args: string[];
+  filters: string[];
+} {
+  const args = [];
+  const filters = [];
+  const { dids, claimTypes, trustedClaimIssuers, scope, includeExpired } = variables;
+  if (dids) {
+    args.push('$dids: [String!]');
+    filters.push('targetId: { in: $dids }');
```
prashantasdeveloper, line 18: In previous middleware POC, @polymath-eric you pointed out that Apollo client would handle the part but apparently the `filter` can neither be empty not can have empty objects. Hence, we need to handle this separately

src/middleware/queriesV2.ts
```diff
@@ -0,0 +1,117 @@
...
+      $count: Int,
+      $skip: Int
+    ) {
+      claims(
+        ${maybeFilters}
+        orderBy: [TARGET_ID_ASC],
+        first: $count,
+        offset: $skip
+      ) {
+        groupedAggregates(groupBy: [TARGET_ID], having: {}) {
```
prashantasdeveloper, line 65: This provides us all the unique `targetId`s. Once we have them, we use the below query to fetch all  claims associated with them

prashantasdeveloper (commented):
@polymath-eric Let me know your thoughts on this pagination logic. PS. ignore the missing tests. Right now this was just a small POC on one of the queries. Tidying up is also left

src/middleware/queriesV2.ts
polymath-eric: This makes sense to me, although double query isn't ideal its probably OK.

I had to change `TARGET_ID` to `TARGET_DID`. We are probably pointing to different SubQueries though.

Also calling `getIdentitiesWithClaimsV2` gets me a 400 when trying out in sandbox though.

src/Claims.ts
monitz87: This belongs in a comment

src/utils/conversion.ts
```diff
@@ -2691,6 +2693,28 @@ export function toIdentityWithClaimsArray(
...
+/**
+ * @hidden
+ */
+export function toIdentityWithClaimsArrayV2(
+  data: MiddlewareV2Claim[],
+  context: Context
+): IdentityWithClaims[] {
+  const groupedData = groupBy(data, 'targetId');
+
+  return Object.keys(groupedData).map(key => ({
```
monitz87, line 2705: You can just use lodash `map` here and get the keys from the second argument of the callback

monitz87 (changes requested):

src/middleware/queriesV2.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+      }
+    }
+  `;
+
+  console.log(query);
+  return {
+    query: gql`
+      ${query}
+    `,
```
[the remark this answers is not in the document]
prashantasdeveloper: This got left over from testing as I was logging the value to verify something. Fixed as suggested

polymath-eric (commented):
LGTM

Few things left to make SonarCloud happy though.

src/utils/conversion.ts
```diff
@@ -2702,14 +2702,14 @@ export function toIdentityWithClaimsArrayV2(
...
-    claims: groupedData[key].map(
+  return map(groupedData, (claims, did) => ({
+    identity: new Identity({ did }, context),
+    claims: claims.map(
       ({ targetId, issuerId, issuanceDate, expiry, type, jurisdiction, scope, cddId }) => ({
         target: new Identity({ did: targetId }, context),
         issuer: new Identity({ did: issuerId }, context),
         issuedAt: new Date(parseFloat(issuanceDate)),
-        expiry: expiry ? new Date(expiry) : null,
+        expiry: expiry ? new Date(parseFloat(expiry)) : null,
```
monitz87, line 2712: Why parseFloat?

src/Claims.ts
```diff
@@ -412,4 +556,96 @@ export class Claims {
...
+            claims: { groupedAggregates: groupedIssuers },
+          },
+        } = await context.queryMiddlewareV2<EnsuredV2<QueryV2, 'claims'>>(
+          claimsGroupingQuery(filters, ClaimsOrderBy.IssuerIdAsc, ClaimsGroupBy.IssuerId)
+        );
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        claimIssuers = flatten(groupedIssuers!.map(groupedAggregate => groupedAggregate.keys!));
+      } else {
+        claimIssuers = trustedClaimIssuers.map(issuer => signerToString(issuer));
```
polymath-eric, line 614: (nit) can just pass `signerToString` no need to make an anonymous function.

src/Claims.ts
```diff
@@ -412,4 +556,96 @@ export class Claims {
...
+      } = await context.queryMiddlewareV2<EnsuredV2<QueryV2, 'claims'>>(
+        claimsQuery({
+          trustedClaimIssuers: claimIssuers,
+          ...filters,
+        })
+      );
+
+      const data = toIdentityWithClaimsArrayV2(
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        nodes.map(node => node!),
```
polymath-eric, line 636:
```suggestion
  nodes.filter(isNotNull)
```
(see above)

src/api/entities/Identity/AssetPermissions.ts
```diff
@@ -350,6 +353,32 @@ export class AssetPermissions extends Namespace<Identity> {
     return optionize(middlewareEventToEventIdentifier)(event);
   }
 
+  /**
+   * Retrieve the identifier data (block number, date and event index) of the event that was emitted when this Identity was enabled/added as
+   *   an Agent with permissions over a specific Asset
+   *
+   * @note uses the middlewareV2
+   * @note there is a possibility that the data is not ready by the time it is requested. In that case, `null` is returned
```
polymath-eric, line 361: I don't fully understand this comment, I get there is eventual consistency between chain and middleware, but wouldn't the case of the Asset simply not existing be hard to distinguish between it will eventually be made.

src/api/entities/Identity/index.ts
```diff
@@ -345,6 +347,52 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+        size,
+        start,
+        order
+      )
+    );
+
+    const count = new BigNumber(totalCount);
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const data = nodes.map(node => new Asset({ ticker: node!.assetId }, context));
```
polymath-eric, line 385: typeGuard filter

src/api/entities/Instruction/index.ts
```diff
@@ -382,6 +385,44 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
...
+      EventIdEnum.InstructionExecuted
+    );
+    if (eventIdentifier) {
+      return {
+        status: InstructionStatus.Executed,
+        eventIdentifier,
+      };
+    }
+
+    eventIdentifier = await this.getInstructionEventFromMiddlewareV2(EventIdEnum.InstructionFailed);
```
polymath-eric, line 412: should this be done in parallel with the `InstructionExecuted` case?

With GraphQL it can be one query, but a `Promise.all` should be OK as well.

src/api/entities/Portfolio/index.ts
```diff
@@ -390,6 +393,143 @@ export abstract class Portfolio extends Entity<UniqueIdentifiers, HumanReadable>
...
+        address: accountAddress,
+      } = node!;
+
+      const { blockId, hash } = createdBlock!;
+
+      data.push({
+        blockNumber: new BigNumber(blockId),
+        blockHash: hash,
+        status: SettlementResultEnum.Executed,
+        accounts: [new Account({ address: keyToAddress('0x' + accountAddress, context) }, context)],
```
polymath-eric, line 516: would be nice to put in a check in `keyToAddress` for the `0x` prefix and add it if needed instead of making the caller know about that detail.

src/base/Context.ts
```diff
@@ -1015,6 +1034,59 @@ export class Context {
...
+        },
+        size,
+        start
+      )
+    );
+
+    const count = new BigNumber(totalCount);
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const data = claimsList.map(claim => middlewareV2ClaimToClaimData(claim!, this));
```
polymath-eric, line 1079: typeGuard filter

src/Network.ts
```diff
@@ -224,6 +277,62 @@ export class Network {
...
+          eventArg1,
+          eventArg2,
+        },
+        size,
+        start
+      )
+    );
+
+    if (events.length) {
+      return events.map(event => {
```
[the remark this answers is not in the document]
polymath-eric: Makes sense, I think the block stuff is safe since it should always be there.

polymath-eric (approved):
LGTM

src/Claims.ts
```diff
@@ -230,6 +264,104 @@ export class Claims {
...
+    } = await context.queryMiddlewareV2<EnsuredV2<QueryV2, 'claims'>>(
+      claimsQuery({
+        dids: targetIssuers,
+        ...filters,
+      })
+    );
+
+    const data = toIdentityWithClaimsArrayV2(
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      nodes.map(node => node!),
```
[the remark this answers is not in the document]
monitz87: The problem is that the code is misleading. We were using an assertion because we know those values cannot be null, and the code is very explicit about that. Someone reading the code with `filter` will assume that the values *can* be null and we're ignoring them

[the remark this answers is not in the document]
monitz87: The "correct" way to handle this would be to remove the `Maybe`s from properties we know are always defined

src/Network.ts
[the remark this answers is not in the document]
monitz87: https://github.com/dotansimha/graphql-code-generator/issues/3919#issuecomment-618431761

src/Claims.ts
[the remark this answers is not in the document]
polymath-eric: If the type says Maybe, then I will think its nullable, and not because of the filter. My concern is something changing and TS will say its OK and we get a runtime exception.

If we can get rid of the `Maybe` where possible I agree that would be better though.

src/Network.ts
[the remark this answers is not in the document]
monitz87: @prashantasdeveloper could you create a task to address this on the subquery side? In the meantime, should we fall back to using assertions instead of isNull or should we leave it for now and add a task (blocked by the subquery one) to re-generate the types and remove the typeguards?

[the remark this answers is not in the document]
prashantasdeveloper: We mainly used the typeguard to remove the non-null elements which we already knew were falsely accused to be null. I tweaked the definition of `Maybe` using the codegen options and we got rid of the unnecessary non-null assertion and typeguard for cases such as filtering `nodes` from the response.

What we are left with are the optional fields which are generated even for non-null one-to-one mappings which I think we can't get rid of (For this I tried using the `optionalFields` param within the codegen options but it then also marks primitive fields also as non-optional). 

I'll add a separate task to make sure all existing primitive fields are correctly marked for in schema for non-null state (!).

polymath-eric (approved):
LGTM

Having `Maybe<T> = T` seems a bit weird to me, like why not just T, but its an improvement over the assertions or guards in the code.

monitz87:
> LGTM
> 
> Having `Maybe<T> = T` seems a bit weird to me, like why not just T, but its an improvement over the assertions or guards in the code.

It's not ideal but it's temporary

monitz87 (approved):

PR #753, opened 2022-05-26
Final 5.0.0 upgrade

polymath-eric:
### Description
Adds SDK support for the chain 5.0.0-rc1. Notable changes 

### JIRA Link

[DA-179](https://polymath.atlassian.net/browse/DA-179)

### Checklist


src/api/entities/Account.ts
```diff
@@ -324,20 +324,37 @@ export class Account extends Entity<UniqueIdentifiers, string> {
...
+    if (keyRecord.isPrimaryKey) {
+      did = identityIdToString(keyRecord.asPrimaryKey);
+    } else if (keyRecord.isSecondaryKey) {
+      did = identityIdToString(keyRecord.asSecondaryKey[0]);
+    } else if (keyRecord.isMultiSigSignerKey) {
+      const multiSigAddress = keyRecord.asMultiSigSignerKey;
+      const rawMultiSigDid = await multiSig.multiSigToIdentity(multiSigAddress);
+      did = identityIdToString(rawMultiSigDid);
+    } else {
+      throw new PolymeshError({
```
polymath-eric, line 352: What I really want is exhaustive enum matching.

It might be possible with `never`, but I think this is simpler. Seems like Sonar cloud doesn't like it though, so maybe the multiSig case could be assumed in the `else`.

src/api/entities/Identity/__tests__/index.ts
```diff
@@ -1205,23 +1230,27 @@ describe('Identity class', () => {
...
     let account: Account;
     let fakeResult: PermissionedAccount[];
 
-    let signatoryToAccountStub: sinon.SinonStub<[Signatory, Context], Account>;
-    let didRecordsStub: sinon.SinonStub;
-    let rawDidRecord: PolymeshPrimitivesIdentity;
+    let rawPrimaryKeyRecord: PolymeshPrimitivesSecondaryKeyKeyRecord;
+    let rawSecondaryKeyRecord: PolymeshPrimitivesSecondaryKeyKeyRecord;
+    let rawMultiSigKeyRecord: PolymeshPrimitivesSecondaryKeyKeyRecord;
+    let rawDidRecord: { args: [PolymeshPrimitivesIdentityId, string] };
```
polymath-eric, line 1240: I saw there was a pattern for this with tuple. Will look into it more in the morning

src/api/entities/Identity/index.ts
```diff
@@ -691,21 +705,45 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+              permissions: meshPermissionsToPermissions(permissions, context),
+            });
+          }
+        }
+
+        return answer;
+      }, [] as PermissionedAccount[]);
     };
 
+    // TODO should use pagination
```
polymath-eric, line 730: https://polymath.atlassian.net/browse/DA-341

I think its OK to merge in to alpha without pagination, but it should be added in since 200 is no longer the limit before beta.

src/api/entities/Account.ts
```diff
@@ -324,20 +324,31 @@ export class Account extends Entity<UniqueIdentifiers, string> {
...
       },
       context,
       address,
     } = this;
 
-    const identityId = await identity.keyToIdentityIds(stringToAccountId(address, context));
+    const optKeyRecord = await identity.keyRecords(stringToAccountId(address, context));
 
-    if (identityId.isEmpty) {
+    if (optKeyRecord.isEmpty) {
```
monitz87, line 336: Use isNone here, it's more correct

src/api/entities/Identity/index.ts
```diff
@@ -279,9 +283,19 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
       context,
     } = this;
 
-    const assembleResult = ({ primaryKey }: PolymeshPrimitivesIdentity): PermissionedAccount => {
+    const assembleResult = ({
+      primaryKey,
+    }: PolymeshPrimitivesIdentityDidRecord): PermissionedAccount => {
+      if (primaryKey.isNone) {
+        throw new PolymeshError({
```
monitz87, line 290: I left a [question](https://polymathnetwork.slack.com/archives/C01RLA867C5/p1653668029041419) in the mesh questions channel because it seems weird to me that this value is an `Option`

src/api/entities/Identity/index.ts
```diff
@@ -671,6 +685,7 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
    * Get the list of secondary Accounts related to the Identity
    *
    * @note can be subscribed to
+   * @note This method currently lacks pagination and maybe slow for identities with many thousands of keys
```
monitz87, line 688:
```suggestion
   * @note This method currently lacks pagination and may be slow for identities with many thousands of keys
```

src/api/entities/Identity/index.ts
```diff
@@ -691,21 +706,44 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
-      return secondaryAccounts.map(({ signer: rawSigner, permissions }) => ({
-        account: signatoryToAccount(rawSigner, context),
-        permissions: meshPermissionsToPermissions(permissions, context),
-      }));
+    const assembleResult = (
+      rawSecondaryKeyKeyRecord: Option<PolymeshPrimitivesSecondaryKeyKeyRecord>[],
+      accountIds: AccountId32[]
+    ): PermissionedAccount[] => {
+      return rawSecondaryKeyKeyRecord.reduce((secondaryKeys, optKeyRecord, index) => {
+        if (optKeyRecord.isSome) {
```
monitz87, line 714: This should always be Some, since the Identity is validated for existence when being fetched

monitz87: We should just unwrap an add a comment explaining why we know

src/api/entities/Identity/index.ts
```diff
@@ -691,21 +706,44 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+            const [, permissions] = keyRecord.asSecondaryKey;
+            secondaryKeys.push({
+              account,
+              permissions: meshPermissionsToPermissions(permissions, context),
+            });
+          }
+        }
+
+        return secondaryKeys;
+      }, [] as PermissionedAccount[]);
```
monitz87, line 728: I think you can pass this as a type parameter to `reduce` instead of having to use an assertion

monitz87 (changes requested):

monitz87 (approved):

PR #759, opened 2022-06-01
chore: 🤖 Pin transitive polkadot deps

polymath-eric:
Pinning transitive polkadot dependencies in case they have breaking
changes

### Description

Pins transitive dependencies for polkadot in the case of a dependency of the dependency not respecting semver.

If we upgrade polkadot to a new version with incompatible deps we will need to update the resolutions. It will give warnings when running yarn when that happens.

For util-crypto I left a few as they were:

```
"@polkadot/wasm-crypto" "^6.0.1"
 "@polkadot/x-randomvalues" "9.0.1"
```

My reasoning being that the risk profile in crypto deps are more likely to be critical bug fixes than user facing breaking api changes.

### JIRA Link

Disovered as part of [DA-348](https://polymath.atlassian.net/browse/DA-348)

### Checklist


polymath-eric:
On second thought I pushed a feat commit.

I am thinking this could clear up the version mismatch warnings for consumers.

monitz87 (approved):

PR #761, opened 2022-06-01
fix: 🐛 Validate primary key on `modifySignerPermissions`

prashantasdeveloper:
### Description

Added validation that the signing Account should be the primary Account of the Identity in `modifySignerPermissions` procedure

### Breaking Changes

🧨 `accountManagement.modifyPermissions` and `accountManagement.revokePermissions`
 now throw an error if the signing Account is not the primary Account of the Identity
 whose secondary Accounts' permissions are modified/revoked

### JIRA Link

DA-308

### Checklist

### DIFF
--- a/src/AccountManagement.ts
+++ b/src/AccountManagement.ts
@@ -111,14 +111,14 @@ export class AccountManagement {
   /**
    * Revoke all permissions of a list of secondary Accounts associated with the signing Identity
    *
-   * @throws if signing Account is not the primary Account of the Identity whose secondary Accounts' permissions are being revoked
+   * @throws if the signing Account is not the primary Account of the Identity whose secondary Account permissions are being revoked
    */
   public revokePermissions: ProcedureMethod<{ secondaryAccounts: Account[] }, void>;
 
   /**
    * Modify all permissions of a list of secondary Accounts associated with the signing Identity
    *
-   * @throws if signing Account is not the primary Account of the Identity whose secondary Accounts' permissions are being modified
+   * @throws if the signing Account is not the primary Account of the Identity whose secondary Account permissions are being modified
    */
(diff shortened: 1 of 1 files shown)


VictorVicente:
> Sonar says the following in the procedure methods
> 
> > Remove the unused function parameter "this" or rename it to "_this" to make intention explicit.
> 
> I think this is something which @polymath-eric also faced once. Probably to be marked as false positive. @monitz87 can you help here?

Solved!

src/AccountManagement.ts
```diff
@@ -108,11 +110,15 @@ export class AccountManagement {
 
   /**
    * Revoke all permissions of a list of secondary Accounts associated with the signing Identity
+   *
+   * @throws if signing Account is not the primary Account of the Identity whose secondary Accounts' permissions are being revoked
```
monitz87, line 114:
```suggestion
   * @throws if the signing Account is not the primary Account of the Identity whose secondary Account permissions are being revoked
```

src/AccountManagement.ts
```diff
@@ -108,11 +110,15 @@ export class AccountManagement {
...
    */
   public revokePermissions: ProcedureMethod<{ secondaryAccounts: Account[] }, void>;
 
   /**
    * Modify all permissions of a list of secondary Accounts associated with the signing Identity
+   *
+   * @throws if signing Account is not the primary Account of the Identity whose secondary Accounts' permissions are being modified
```
monitz87, line 121:
```suggestion
   * @throws if the signing Account is not the primary Account of the Identity whose secondary Account permissions are being modified
```

monitz87 (changes requested):

VictorVicente (approved):

PR #762, opened 2022-06-01
fix: 🐛 Settlement not being modifable before end block

polymath-eric:
Use BigNumber gte function over >=, as the inline operator produces
unexpected results, preventing settlements from being affirmed before
the instructions end block

### Description
Using >= with big number seems to ignore significant digits. This prevented some instructions from being modified when they had an endBlock specified at times. Depends on the latest block value, so its a flaky error case.

```js
const one = new BigNumber(5680);
  const two = new BigNumber(10000000);
  console.log('one bigger? ', one >= two);
  console.log('two bigger? ', two >= one);
```
### JIRA Link

const one = new BigNumber(5680);
  const two = new BigNumber(10000000);
  console.log('one bigger? ', one >= two);
  console.log('two bigger? ', two >= one);

### Checklist


prashantasdeveloper (approved):

VictorVicente (approved):
Nice catch!

PR #768, opened 2022-06-02
fix: 🐛 Handle NoData claimType in conversion util

polymath-eric:
### Description

`sdk.claims.getTargetingClaims()` was throwing an error due to iterating on ClaimType and passing to createType.

We need to keep NoData around, since Middleware uses it, but in createType, we can only pass on or the other.

They are equivalent in function so this should be OK, although its a bit messy.

### Checklist

### DIFF
--- a/src/utils/__tests__/conversion.ts
+++ b/src/utils/__tests__/conversion.ts
@@ -3593,7 +3593,7 @@ describe('meshClaimTypeToClaimType and claimTypeToMeshClaimType', () => {
   });
 
   describe('claimTypeToMeshClaimType', () => {
-    it('should convert a ClaimType to a polkadot Claim', () => {
+    it('should convert a ClaimType to a polkadot ClaimType', () => {
       const context = dsMockUtils.getContextInstance();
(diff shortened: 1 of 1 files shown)


prashantasdeveloper (approved):
LGTM. If you could add a small test for this that would be great

src/utils/__tests__/conversion.ts
```diff
@@ -3523,73 +3524,97 @@ describe('meshClaimTypeToClaimType and claimTypeToMeshClaimType', () => {
...
       claimType = dsMockUtils.createMockClaimType(fakeResult);
 
-      result = claimTypeToClaimType(claimType);
+      result = meshClaimTypeToClaimType(claimType);
       expect(result).toEqual(fakeResult);
     });
   });
+
+  describe('claimTypeToMeshClaimType', () => {
+    it('should convert a ClaimType to a polkadot Claim', () => {
```
monitz87, line 3596:
```suggestion
    it('should convert a ClaimType to a polkadot ClaimType', () => {
```

monitz87 (changes requested):

VictorVicente (approved):

PR #770, opened 2022-06-03
Use createGroupAndAddAuth when inviting external agents

monitz87:
### Description

- Replace the use of `createGroup` in the `inviteExternalAgent` procedure with `createGroupAndAddAuth`

### JIRA Link

https://polymath.atlassian.net/browse/DA-353

### Checklist


src/utils/__tests__/conversion.ts
```diff
@@ -3236,7 +3236,7 @@ describe('claimToMeshClaim and meshClaimToClaim', () => {
       let value: Claim = {
         type: ClaimType.Jurisdiction,
         code: CountryCode.Cl,
-        scope: { type: ScopeType.Identity, value: 'someTickerDid' },
+        scope: { type: ScopeType.Identity, value: 'SOME_TICKERDid' },
```
polymath-eric, line 3239: (nit) mixed case looks weird to me

```suggestion
        scope: { type: ScopeType.Identity, value: 'SOME_TICKER_DID' },
```

src/api/procedures/inviteExternalAgent.ts
```diff
@@ -114,16 +130,22 @@ export async function prepareInviteExternalAgent(
...
-    })) as PostTransactionValue<CustomPermissionGroup>;
-    postTransactionAuthorization = createGroupResult.transform(createBecomeAgentData);
-
-    rawAuthorizationData = postTransactionAuthorization.transform(authorization =>
-      authorizationToAuthorizationData(authorization, context)
-    );
+    const rawTicker = stringToTicker(ticker, context);
+    const { transactions } = permissionsLikeToPermissions(permissions, context);
+
+    await assertGroupDoesNotExist(asset, transactions);
```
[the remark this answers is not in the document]
monitz87: Now that I think about it, it does make sense. I guess I thought that if users were passing permissions explicitly, they were expecting a group to be created, but we're actually "hiding" that from them and it shouldn't make much of a difference. Nice catch

src/utils/__tests__/conversion.ts
monitz87: My bad, I did a mass replace of `someTicker` to `SOME_TICKER`

VictorVicente (approved):

src/utils/__tests__/conversion.ts
polymath-eric: There's a few more spots like this still

PR #775, opened 2022-06-06
feat: 🎸 support passing existing group permissions

monitz87:
### Description

When adding an external agent, we used to throw an error if the passed
permissions corresponded to an existing group. Now that group is fetched
and used to create the authorization

### Breaking Changes

no longer throw an error when adding an external agent with explicit
permissions if there is already an existing permission group with those
permissions

### Checklist


polymath-eric (approved):
LGTM, this should save some confusion for whoever is using this.

One unused import to remove according to SonarCloud btw.

prashantasdeveloper (approved):

VictorVicente (approved):

PR #777, opened 2022-06-09
feat: 🎸 Publish to polymesh association npm

polymath-eric:
Publish to the new association npm organization. The previous
@polymathnetwork organization remains for now

### Description

Publishes the package under the new association. Uses `yarn publish` to publish to the previous account for now.

### JIRA Link

[DA-360](https://polymath.atlassian.net/browse/DA-360)

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,12 +62,5 @@ jobs:
           NPM_TOKEN: ${{ secrets.ASSOCIATION_NPM_TOKEN }}
         run: |
           cd dist
-          yarn semantic-release
-      # Publish to the old npm repository for now
-      - name: legacy-release
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-        run: |
-          jq '.name = $newVal' --arg newVal '@polymathnetwork/polymesh-sdk' < package.json > tmp.json && mv tmp.json package.json
           yarn --frozen-lockfile
-          yarn publish
+          yarn semantic-release


.github/workflows/main.yml
```diff
@@ -59,7 +59,15 @@ jobs:
...
           cd dist
           yarn semantic-release
+      # Publish to the old npm repository for now
+      - name: legacy-release
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          jq '.name = $newVal' --arg newVal '@polymathnetwork/polymesh-sdk' < package.json > tmp.json && mv tmp.json package.json
+          yarn --frozen-lockfile
+          yarn publish
```
polymath-eric, line 73: @monitz87 My understanding is this will be the same as semantic release, except for not pushing stuff back to git like tags + notes.

I originally tried to run semantic release here, but it would give an error about tag existing. When I removed the tag for second run it would say no changes. Seems like using basic publish is the way to go.

I looked at a [plugin](https://github.com/amanda-mitchell/semantic-release-npm-multiple) for semantic release for multi releases, but it was geared for multi registry, e.g. npm + github, so didn't quite work this use case. It would need the package name overridden for one, but otherwise would work.

monitz87: Won't this publish a new release regardless of the type of commits? The main allure of semantic release is that it calculates the version based on the types of commits that exist betweeen the last tag and the current one, and doesn't publish unless there is a feature, bugfix or breaking change. This seems like it will publish on any push

polymath-eric: Looks like semantic release gates the release from a PR, so I'll need to add a branch check in this step as well.

monitz87: You can configure github actions to store variables and reuse them accross steps. Assuming you can capture the relevant output from semantic release, you could inform the next step whether to run or not depending on whether an actual release was made. Also I'm not entirely sure how semantic release injects the version into the package.json before publishing. You'd have to make sure that the next step has access to the same package.json, or otherwise capture the version from the semantic release step and rewrite the package.json yourself before publishing

Seems like more trouble than it is worth

monitz87 (approved):

PR #782, opened 2022-06-14
feat: 🎸 Add enableStat to transferRestriction namespace

polymath-eric:
### Description

Adds in `enableStat` method to count and percentage transfer restriction namespace.

### JIRA Link

[DA-326](https://polymath.atlassian.net/browse/DA-326)

### Checklist


src/api/entities/Asset/TransferRestrictions/Percentage.ts
```diff
@@ -46,6 +47,11 @@ export class Percentage extends TransferRestrictionBase<TransferRestrictionType.
    */
   public declare removeRestrictions: NoArgsProcedureMethod<BigNumber>;
 
+  /**
+   * Enables investor ownership percentage statistic for the Asset, which is required before creating restrictions
+   */
+  public declare enableStat: ProcedureMethod<Omit<AddPercentStatParams, 'type'>, void>;
```
polymath-eric, line 53: Ideally this would be a NoArgsProcedure, but I haven't figured out how to make that work when Count needs an arg.

Right now this requires the user invoke the function with an empty object, like:
`asset.transferRestrictions.percentage.enableStat({});`

I based this off TransferRestriction, but in that example they both need args.

polymath-eric: Actually seems like a NoArgsProcedure is working after all. Not sure what was different when I tried it yesterday.

src/api/procedures/__tests__/removeAssetStat.ts
```diff
@@ -0,0 +1,283 @@
...
+  });
+
+  it('should add an setAssetStats transaction to the queue', async () => {
+    args = {
+      type: StatType.Balance,
+      ticker,
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const currentStats = { toArray: () => [], indexOf: () => 0 } as any;
```
polymath-eric, line 129: Looking at making this into a proper mock

src/api/procedures/addTransferRestriction.ts
```diff
@@ -250,10 +203,15 @@ export async function prepareStorage(
     return cmpStat === type;
```
prashantasdeveloper, line 203: Same here. We need to check the claimType and issuer as well

src/api/procedures/removeAssetStat.ts
```diff
@@ -0,0 +1,144 @@
...
+      (type === StatType.Count && r.isMaxInvestorCount) ||
+      (type === StatType.Balance && r.isMaxInvestorOwnership)
+    ) {
+      throw new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message:
+          'This statistics cannot be removed because a TransferRequirement is currently using it',
+      });
+    }
+  });
```
prashantasdeveloper, line 133: For `ClaimCount` and `ClaimOwnership` compliance types, you need to consider the claimTypes as well. Also for `MaxInvestorCount` and `MaxInvestorOwnership` you need to check if claimType is `null`

prashantasdeveloper (changes requested):

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+    issuer: { did: issuerDid },
+    claimType,
+  } = claimIssuer;
+  if (type === StatType.ScopedCount && transferCondition.isClaimCount) {
+    const [rawClaim, issuer] = transferCondition.asClaimCount;
+    const restrictionIssuerDid = identityIdToString(issuer);
+    const claim = statsClaimToStatClaimUserType(rawClaim);
+    if (restrictionIssuerDid === issuerDid && claim.type === claimType) {
+      return true;
+    }
```
monitz87, line 1280: These three things are repeated in both blocks

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+  }
+  return a.eq(b);
+}
+
+/**
+ * @hidden
+ */
+export function compareTransferRestrictionToInput(
+  transferRestriction: PolymeshPrimitivesTransferComplianceTransferCondition,
+  value: BigNumber | ClaimCountRestrictionValue,
```
monitz87, line 1326: Shouldn't this also be `ClaimPercentageRestrictionValue`?

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+    type === TransferRestrictionType.Percentage
+  ) {
+    const currentOwnership = permillToBigNumber(transferRestriction.asMaxInvestorOwnership);
+    return currentOwnership.eq(value as BigNumber);
+  } else if (transferRestriction.isClaimCount && type === TransferRestrictionType.ClaimCount) {
+    const [statClaim, rawIssuerId, rawMin, maybeMax] = transferRestriction.asClaimCount;
+    const issuerDid = identityIdToString(rawIssuerId);
+    const min = u64ToBigNumber(rawMin);
+    const max = maybeMax.isSome ? u64ToBigNumber(maybeMax.unwrap()) : undefined;
+    const castedValue = value as ClaimCountRestrictionValue;
```
monitz87, line 1343: Why are these assertions necessary?

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+      castedValue.min.eq(min) &&
+      compareOptionalBigNumbers(max, castedValue.max) &&
+      castedValue.claim.type === getClaimType(statClaim) &&
+      issuerDid === castedValue.issuer.did
+    );
+  } else if (
+    transferRestriction.isClaimOwnership &&
+    type === TransferRestrictionType.ClaimPercentage
+  ) {
+    const castedValue = value as ClaimCountRestrictionValue;
```
monitz87, line 1355: This should be `ClaimPercentageRestrictionValue` (assuming the casts are necessary)

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+    const [statClaim, rawIssuerId, rawMin, rawMax] = transferRestriction.asClaimOwnership;
+    const issuerDid = identityIdToString(rawIssuerId);
+
+    const min = permillToBigNumber(rawMin);
+    const max = permillToBigNumber(rawMax);
+
+    return !!(
+      castedValue.min.eq(min) &&
+      compareOptionalBigNumbers(castedValue.max, max) &&
```
monitz87, line 1364: These aren't optional

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+      castedValue.claim.type === getClaimType(statClaim) &&
+      issuerDid === castedValue.issuer.did
+    );
+  }
```
monitz87, line 1368: There's a lot of repeated code between the last two blocks, you should be able to reuse some of it, or at least destructure `castedValue` to avoid repeating it so much

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+    const currentOwnership = permillToBigNumber(transferRestriction.asMaxInvestorOwnership);
+    return currentOwnership.eq(value as BigNumber);
+  } else if (transferRestriction.isClaimCount && type === TransferRestrictionType.ClaimCount) {
+    const [statClaim, rawIssuerId, rawMin, maybeMax] = transferRestriction.asClaimCount;
+    const issuerDid = identityIdToString(rawIssuerId);
+    const min = u64ToBigNumber(rawMin);
+    const max = maybeMax.isSome ? u64ToBigNumber(maybeMax.unwrap()) : undefined;
+    const castedValue = value as ClaimCountRestrictionValue;
+
+    return !!(
```
monitz87, line 1345: The !! shouldn't be necessary

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+    return transferRestrictionType === TransferRestrictionType.Percentage;
+  } else if (opType === StatisticsOpType.ClaimCount) {
+    return transferRestrictionType === TransferRestrictionType.ClaimCount;
+  } else {
+    return transferRestrictionType === TransferRestrictionType.ClaimPercentage;
+  }
+}
+
+/**
+ *@ hidden
```
monitz87, line 1393:
```suggestion
 * @hidden
```

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+  const { type, claimIssuer } = args;
+
+  let rawOp;
+  if (type === TransferRestrictionType.Count || type === TransferRestrictionType.ClaimCount) {
+    rawOp = statisticsOpTypeToStatOpType(StatisticsOpType.Count, context);
+  } else {
+    rawOp = statisticsOpTypeToStatOpType(StatisticsOpType.Balance, context);
+  }
+
+  const rawIssuer = claimIssuer ? claimIssuerToMeshClaimIssuer(claimIssuer, context) : undefined;
```
monitz87, line 1412: Can't you use optionize here?

monitz87: (I'm assuming `claimIssuer` is an `Option`)

src/utils/internal.ts
```diff
@@ -1174,3 +1197,237 @@ export function assertTickerValid(ticker: string): void {
...
+  currentStats: BTreeSet<PolymeshPrimitivesStatisticsStatType>,
+  neededStat: PolymeshPrimitivesStatisticsStatType
+): void {
+  const needStat = ![...currentStats].find(s => neededStat.eq(s));
+
+  if (needStat) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message:
+        'The appropriate stat type for this restriction is not set for the Asset. Try calling enableStat in the namespace first',
```
monitz87:
```suggestion
        "The stat type required for this restriction is not enabled for the Asset. Try calling `enableStat` in the restriction's namespace first",
```

monitz87 (changes requested):
I would open the discussion on whether to rename `enableStat` to just `enable`. We're trying to hide the whole statistics thing from the user

src/utils/conversion.ts
```diff
@@ -3771,16 +3771,72 @@ export function createStat2ndKey(
...
+      if (!condition.asClaimCount[0].isJurisdiction) {
+        return undefined;
+      } else {
+        return condition.asClaimCount[0].asJurisdiction;
+      }
+    } else if (condition.isClaimOwnership) {
+      if (!condition.asClaimOwnership[0].isJurisdiction) {
+        return undefined;
+      }
+      return condition.asClaimOwnership[0].asJurisdiction;
```
monitz87, line 3796: Use destructuring here

src/utils/conversion.ts
```diff
@@ -3771,16 +3771,72 @@ export function createStat2ndKey(
...
+      return -1;
+    }
+    if (bClaim.isNone) {
+      return 1;
+    }
+    const aCode = meshCountryCodeToCountryCode(aClaim.unwrap());
+    const bCode = meshCountryCodeToCountryCode(bClaim.unwrap());
+
+    const countryOrder = Object.values(CountryCode);
+    return countryOrder.indexOf(aCode) - countryOrder.indexOf(bCode);
```
monitz87, line 3825: You can just use `aCode.localeCompare(bCode)`

monitz87 (changes requested):

src/api/entities/Asset/TransferRestrictions/Percentage.ts
```diff
@@ -20,9 +21,12 @@ export class Percentage extends TransferRestrictionBase<TransferRestrictionType.
   protected type = TransferRestrictionType.Percentage as const;
 
   /**
-   * Add a Percentage Transfer Restriction to this Asset
+   * Add a Percentage Transfer Restriction to this Asset. This limits the total percentage of the float
```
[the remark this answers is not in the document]
polymath-eric: https://www.investopedia.com/ask/answers/what-is-companys-float/

Lets us supply though. I think I misused the term also

src/api/entities/Asset/TransferRestrictions/ClaimCount.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  public declare removeRestrictions: NoArgsProcedureMethod<BigNumber>;
+
+  /**
+   * Enables investor count statistic for the Asset scope by a claim, which is required before creating restrictions
+   *
+   * The number of investors need to be passed in, as the chain will not calculate the initial value for that stat,
+   * it only increments / decrements the counters as transactions happen.
+   * For `Affiliate` and `Accredited` scoped stats the number of investor who are and are not
+   * need to be given. For `Jurisdiction` the amount of holders for each CountryCode need to be given.
```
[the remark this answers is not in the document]
polymath-eric: It needs the amount that are, as well as the amount that aren't. So accredited + non accredited.

Reworded it a bit, hopefully thats better

src/utils/internal.ts
polymath-eric: It is, but its actually Option<ITuple>, which made it trickier than dealing with just a normal tuple and letting polkadot.js deal with it.

We could look into making an ITuple conversion util (or maybe I am missing it), but for now I think this is easier.

for reference raw type is: 
```
Option<
      ITuple<[PolymeshPrimitivesIdentityClaimClaimType, PolymeshPrimitivesIdentityId]>
 >;
```
and I get `[PolymeshPrimitivesIdentityClaimClaimType, PolymeshPrimitivesIdentityId]` doesn't implement Codec error

src/utils/conversion.ts
```diff
@@ -2870,26 +2873,33 @@ export function transferRestrictionToPolymeshTransferCondition(
   } else if (type === TransferRestrictionType.Percentage) {
     restrictionType = 'MaxInvestorOwnership';
     restrictionValue = percentageToPermill(value as BigNumber, context);
-  } else {
-    const castedValue = value as ClaimCountRestrictionValue;
+  } else if (
+    type === TransferRestrictionType.ClaimCount ||
+    type === TransferRestrictionType.ClaimPercentage
+  ) {
```
monitz87, line 2932: Can't you destructure `min`, `max`, `claim` and `issuer` here at the beginning?

src/utils/conversion.ts
```diff
@@ -3783,17 +3784,18 @@ export function sortTransferRestrictionByClaimValue(
   const getJurisdictionValue = (
     condition: PolymeshPrimitivesTransferComplianceTransferCondition
   ): Option<MeshCountryCode> | undefined => {
-    if (condition.isClaimCount) {
-      if (!condition.asClaimCount[0].isJurisdiction) {
+    const { isClaimCount, asClaimCount, isClaimOwnership, asClaimOwnership } = condition;
```
monitz87, line 3855: You can't destructure the `asX` members right away, they are getters that throw if the `isX` part is false. You have to destructure them inside the corresponding if

src/utils/internal.ts
```diff
@@ -1271,21 +1271,20 @@ export function compareTransferRestrictionToStat(
...
-    }
-  }
 
-  if (type === StatType.ScopedBalance && transferCondition.isClaimOwnership) {
-    const [rawClaim, issuer] = transferCondition.asClaimOwnership;
-    const restrictionIssuerDid = identityIdToString(issuer);
-    const claim = statsClaimToStatClaimUserType(rawClaim);
-    if (restrictionIssuerDid === issuerDid && claim.type === claimType) {
-      return true;
+  if (type === StatType.ScopedCount || type === StatType.ScopedBalance) {
```
monitz87: What happens if `type === StatType.ScopedCount` but `transferCondition.isClaimOwnership`?

What about

```
let rawClaim, issuer;

if (type === StatType.ScopedCount && transferCondition.isClaimCount) {
  [rawClaim, issuer] = transferCondition.asClaimCount;
}

if (type === StatType.ScopedBalance && transferCondition.isClaimOwnership) {
  [rawClaim, issuer] = transferCondition.asClaimOwnership;
}

if (rawClaim && issuer) {
  // ...
}

return false;
```

monitz87 (changes requested):

src/api/procedures/addTransferRestriction.ts
```diff
@@ -90,14 +90,8 @@ export async function prepareAddTransferRestriction(
...
   const maxConditions = u32ToBigNumber(consts.statistics.maxTransferConditionsPerAsset);
-  console.log(
-    'restrictions size',
-    currentRestrictions,
-    currentRestrictions.size,
-    [...currentRestrictions].length
-  );
 
-  const restrictionAmount = new BigNumber(currentRestrictions.size);
+  const restrictionAmount = new BigNumber([...currentRestrictions].length);
```
monitz87, line 94: Why this change?

src/utils/__tests__/conversion.ts
```diff
@@ -7852,7 +7851,9 @@ describe('sortByClaimType', () => {
     const countStat = dsMockUtils.createMockStatisticsStatType({ op });
 
     let result = sortStatsByClaimType([jurisdictionStat, accreditedStat, affiliateStat, countStat]);
+    console.log(JSON.stringify(result));
```
monitz87: This should go

src/utils/__tests__/conversion.ts
```diff
@@ -7852,7 +7851,9 @@ describe('sortByClaimType', () => {
...
     expect(result).toEqual([accreditedStat, affiliateStat, jurisdictionStat, countStat]);
+    // expect(result).toEqual([jurisdictionStat, accreditedStat, affiliateStat, countStat]);
```
monitz87: Either delete or uncomment

src/utils/internal.ts
```diff
@@ -1257,6 +1257,7 @@ export function compareTransferRestrictionToStat(
   type: StatType,
   claimIssuer?: StatClaimIssuer
 ): boolean {
+  console.log('comparing internally', type, transferCondition);
```
monitz87: This should go

src/api/procedures/__tests__/removeAssetStat.ts
```diff
@@ -274,6 +283,7 @@ describe('removeAssetStat procedure', () => {
       },
     };
     await expect(prepareRemoveAssetStat.call(proc, args)).rejects.toThrowError(expectedError);
+    console.log('being used done');
```
monitz87: This should go

monitz87 (changes requested):

monitz87 (changes requested):

src/types/internal.ts
```diff
@@ -390,7 +390,7 @@ export enum StatType {
    */
   ScopedCount = 'ScopedCount',
   /**
-   * Keeps track of the amount of supply investors hold who have a certain claim
+   * Keeps track of the percentage of total supply held by investors hold who have a certain claim
```
monitz87, line 393:
```suggestion
   * Keeps track of the percentage of total supply held by investors who have a certain claim
```

src/utils/conversion.ts
```diff
@@ -2932,21 +2932,22 @@ export function transferRestrictionToPolymeshTransferCondition(
...
-      rawMin = bigNumberToU64(value.min, context);
-      rawMax = optionize(bigNumberToU64)(value.max, context);
+      rawMin = bigNumberToU64(min, context);
+      rawMax = optionize(bigNumberToU64)(max, context);
     } else {
       // i.e. TransferRestrictionType.ClaimPercentage
       restrictionType = 'ClaimOwnership';
-      rawMin = percentageToPermill(value.min, context);
+      rawMin = percentageToPermill(min, context);
       rawMax = percentageToPermill(value.max, context);
```
monitz87, line 2944:
```suggestion
      rawMax = percentageToPermill(max, context);
```

monitz87 (changes requested):

src/utils/conversion.ts
polymath-eric: max got destructured earlier, so its type is `BigNumber | undefined` (optional for count).

I did this to satisfy the type checker without needing an assertion, or have another line for the assignment.

monitz87: Huh, thought the type checker was smarter. Fair enough

src/api/procedures/types.ts
```diff
@@ -75,6 +79,85 @@ export type GetTransferRestrictionReturnType<T> = ActiveTransferRestrictions<
...
+
+export type AddBalanceStatParams = {
+  type: StatType.Percentage;
+};
+
+export type AddClaimCountStatParams = ClaimCountStatInput & {
+  type: StatType.ScopedCount;
+};
+
+export type AddClaimPercentageStatParams = StatClaimIssuer & {
```
monitz87, line 101: `AddBalanceStatParams` and `AddClaimPercentageStatParams` have inconsistent naming

src/api/procedures/createAsset.ts
```diff
@@ -224,6 +230,18 @@ export async function prepareCreateAsset(
     );
   }
 
+  if (initialStatistics?.length) {
```
monitz87, line 233: Even though currently there are no protocol fees for setting active stats, there might be in the future. You should use the `addManualFees` "trick" we're using for all other transactions in this procedure (remember to change the comment to reflect this case)

monitz87: Looking back, calling `addManualFees` once per every transaction we add is cumbersome. I think it makes more sense to store all the tags in an array and just make a single call before the `addBatchTransaction` call

monitz87 (changes requested):

src/api/procedures/createAsset.ts
polymath-eric: We are storing the transactions in an array, so with a lookup table, we can map from transactions to tags.

polymath-eric: After trying it out, I think we should keep a separate array like you suggested, 

Mapping from `SubmittableExtrinsic` types is vague and harder to reason about. Using `toString` could work, but the extra variable is cleaner imo.

prashantasdeveloper (approved):

monitz87 (approved):

PR #783, opened 2022-06-24
feat: 🎸 upgrade polkadot.js to 8.9.1

polymath-eric:
### Description

Upgrade polkadot.js to get access to BTreeSet typegen supoort

### JIRA Link

[DA-368](https://polymath.atlassian.net/browse/DA-368)

### Checklist


prashantasdeveloper (approved):
LGTM

src/testUtils/mocks/dataSources.ts
```diff
@@ -1614,6 +1615,24 @@ export const createMockEcdsaSignature = (
...
+ */
+export const createMockBTreeSet = <T extends Codec>(
+  items: BTreeSet<T> | unknown[]
+): MockCodec<BTreeSet<T>> => {
+  if (isCodec<BTreeSet>(items)) {
+    return items as MockCodec<BTreeSet<T>>;
+  }
+  const res = createMockCodec(items, !items) as unknown as BTreeSet;
+  res.has = sinon.stub().returns(false);
+  (res as any).size = items.length;
```
monitz87, line 1629: Why is this assertion necessary?

src/testUtils/mocks/dataSources.ts
```diff
@@ -1614,6 +1615,24 @@ export const createMockEcdsaSignature = (
...
+  items.forEach(i => {
+    (res.has as sinon.SinonStub).withArgs(i).returns(true);
```
monitz87, line 1631: If you store the stub in a variable before assigning it to the result you can use that and avoid having to use this type assertion

monitz87 (changes requested):

src/testUtils/mocks/dataSources.ts
polymath-eric: I think size is a special readonly property of a BTreeSet, so I used this to set it anyway. I forget the exact reason but will look again in the morning

VictorVicente (approved):

PR #786, opened 2022-07-12
Alpha into beta

polymath-eric:
### Description

Merge the latest alpha changes into beta


polymath-eric:
There are 12 code smells, but its for deprecated calls and changing the autogen files from polkadot.js

I think we should ignore the smells for this PR. We might want to change the deprecated stuff in a later PR, but I figure we can put it off until the deprecated becomes breaking.

VictorVicente (approved):

PR #788, opened 2022-07-13
Da 325/create asset with stats

polymath-eric:
### Description

Adds ability to create an Asset with statistics

### Breaking Changes

N/A

### JIRA Link

[DA-325](https://polymath.atlassian.net/browse/DA-325)


src/api/procedures/types.ts
```diff
@@ -236,6 +236,18 @@ export interface CreateAssetParams {
...
+
+  /**
+   * (optional) type of statistics that should be enabled for the Asset
+   *
+   * Enabling statistics allows for TransferRestrictions to be made. For example the SEC requires registration for a company that
+   * has either more than 2000 investors, or more than 500 non accredited investors. To prevent crossing this limit two restrictions are
+   * needed, a `Count` of 2000, and a `ScopedCount` of non accredited with a maximum of 500. [source](https://www.sec.gov/info/smallbus/secg/jobs-act-section-12g-small-business-compliance-guide.htm)
+   *
+   * These restrictions require a `Count` and `ScopedCount` statistic to be created. Although they an be created after the Asset is made, it is recommended to create statistics
+   * before the Asset is circulated. Count statistics made after Asset creation need their initial value set, so it is simpler to create them before investors hold the Asset.
```
prashantasdeveloper, line 248: nit-> We can also add link to `batchUpdateAssetStats` to make it clear how to set an asset stat value if it has not been created at the time of asset creation

prashantasdeveloper (approved):
LGTM

monitz87 (changes requested):

src/api/procedures/types.ts
polymath-eric: Thats the raw chain method, which I don't think we should link from the SDK. I think linking to `enableStat` would work though, it takes in initial values as input when needed.

src/types/index.ts
```diff
@@ -339,11 +339,47 @@ export type StatClaimInput =
...
+      type: StatType.Balance;
+    }
+  | {
+      type: StatType.ScopedCount;
+      claimIssuer: StatClaimIssuer;
+    }
+  | {
+      type: StatType.ScopedBalance;
+      claimIssuer: StatClaimIssuer;
+    };
```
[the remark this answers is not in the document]
polymath-eric: I went with the union type. If there is a new stat type I think we'd need to add it in. If it happens to use claimIssuer like this, then it should be just as easy to add in.

src/utils/conversion.ts
```diff
@@ -544,7 +544,7 @@ export function permillToBigNumber(value: Permill): BigNumber {
 /**
  *  @hidden
  */
-export function meshClaimToStatClaimInput(claim: StatClaim): StatClaimInput {
+export function meshClaimToStatClaimInput(claim: StatClaim): InputStatClaim {
```
monitz87, line 547: ToInputStatClaim

monitz87 (changes requested):
Apparently something went wrong with a find and replace, there's a lot of lines that aren't valid typescript

monitz87 (approved):

PR #793, opened 2022-07-18
Da 341/paginate secondary accounts

polymath-eric:
### Description

Adds pagination to getSecondaryAccounts()

Introduced getSecondaryKeyPermission() method so other SDK methods have a more efficient method rather than checking all secondary accounts

### Breaking Changes

N/A — (not passing pagination args defaults to all entries in requesPaginated)

### JIRA Link

[DA-341](https://polymath.atlassian.net/browse/DA-341)

### Checklist


src/api/entities/Identity/index.ts
```diff
@@ -683,17 +685,22 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
 
   // eslint-disable-next-line require-jsdoc
-  public async getSecondaryAccounts(
-    callback?: SubCallback<PermissionedAccount[]>
-  ): Promise<PermissionedAccount[] | UnsubCallback> {
+  public async getSecondaryAccounts(args: {
+    callback?: SubCallback<PermissionedAccount[]>;
+    opts?: PaginationOptions;
+    fetchAll?: boolean;
+  }): Promise<ResultSet<PermissionedAccount> | UnsubCallback> {
```
monitz87, line 703: For consistency, `callback` should be a separate parameter. Use `AccountManagement.getAccountBalance` for reference. Also, I assume pagination options and `fetchAll` are not meant to be passed together, so there should be two separate signatures where each of them is defined and the other isn't. This means a lot of possible combinations when mixed with the optional callback, so I suggest defining a separate type for the first argument

Also, you can type `fetchAll` as `true` (you should also document it or at least use the exact property name in the note)

src/api/entities/Identity/index.ts
```diff
@@ -742,13 +759,61 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
 
-    return assembleResult(rawPermissions, identityKeys);
+    return {
+      data: assembleResult(rawPermissions, identityKeys),
+      next,
+    };
+  }
+
+  /**
+   * Fetches account permissions for the given secondary Accounts
```
monitz87, line 769:
```suggestion
   * Fetches Account permissions for the given secondary Accounts
```

src/api/entities/Identity/index.ts
```diff
@@ -742,13 +759,61 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+      const record = optKeyRecord.unwrap();
+      if (record.isSecondaryKey) {
+        const [rawIdentityId, rawPermissions] = record.asSecondaryKey;
+
+        if (identityIdToString(rawIdentityId) !== did) {
+          throw new PolymeshError({
+            code: ErrorCode.General,
+            message: 'Given an account that belongs to another Identity',
+          });
+        }
```
monitz87, line 801: Remember to remove this when you transform the method into a util

src/api/entities/Identity/index.ts
```diff
@@ -742,13 +759,61 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+      context: {
+        polymeshApi: {
+          query: { identity },
+        },
+      },
+    } = this;
+
+    const { accounts } = args;
+
+    const identityKeys = accounts.map(a => stringToAccountId(a.address, context));
```
monitz87, line 788:
```suggestion
    const identityKeys = accounts.map(({ address }) => stringToAccountId(address, context));
```

src/api/entities/Identity/index.ts
```diff
@@ -742,13 +759,61 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+    const results = await identity.keyRecords.multi(identityKeys);
```
monitz87, line 789:
```suggestion
    const results = await identity.keyRecords.multi(identityKeys);
    
```

src/api/procedures/leaveIdentity.ts
```diff
@@ -22,10 +22,11 @@ export async function prepareLeaveIdentity(this: Procedure<void, void>): Promise
     });
   }
 
-  const secondaryAccounts = await signingIdentity.getSecondaryAccounts();
-  const isSecondaryAccount = secondaryAccounts.find(({ account }) =>
-    account.isEqual(signingAccount)
-  );
+  const accountPermission = await signingIdentity.getSecondaryAccountPermissions({
```
monitz87, line 25: You can just destructure the first element of this and check if it is defined

src/api/entities/Identity/index.ts
```diff
@@ -742,13 +759,61 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+    return {
+      data: assembleResult(rawPermissions, identityKeys),
+      next,
+    };
+  }
+
+  /**
+   * Fetches account permissions for the given secondary Accounts
+   *
+   * @throws if given an Account that belongs to a different Identity. Unlinked Accounts are ignored
```
monitz87, line 771: Remember to remove this when you transform the method into a util

monitz87 (changes requested):

monitz87:
Also, there are breaking channges (`getPermissionedAccounts` has a different return type now, and a different signature as well)

src/api/entities/Identity/index.ts
polymath-eric: Good idea about having getPermisionedAccounts calling this. I made it support a callback, it doesn't take pagination params though since its a wrapper around the `.multi` call. `getSecondaryAccounts` will still handle pagination part.

src/api/procedures/modifySignerPermissions.ts
```diff
@@ -34,7 +35,10 @@ export async function prepareModifySignerPermissions(
 
   const { secondaryAccounts } = args;
   const accounts = secondaryAccounts.map(({ account }) => account);
-  const existingSecondaryAccounts = await identity.getSecondaryAccountPermissions({ accounts });
+  const existingSecondaryAccounts = await getSecondaryAccountPermissions(
+    { accounts, identity },
+    context
+  );
   assertSecondaryAccounts(accounts, existingSecondaryAccounts);
```
monitz87, line 42: I think this function can be simplified to just checking if the lengths of both arrays are the same

src/utils/internal.ts
```diff
@@ -1197,3 +1207,73 @@ export function neededStatTypeForRestrictionInput(
...
+
+/**
+ * @hidden
+ *
+ * Fetches Account permissions for the given secondary Accounts
+ *
+ * @note non secondary Accounts will be skipped, so there maybe less PermissionedAccounts returned than Accounts given
+ *
+ * @param args.accounts a list of accounts to fetch permissions for
+ * @param args.did an optional did to filter out keys that do not belong to an Identity
```
monitz87, line 1219:
```suggestion
 * @param args.accounts - a list of Accounts to fetch permissions for
 * @param args.Identity - optional. If passed, Accounts that are not part of the given Identity will be filtered out
```

src/utils/internal.ts
```diff
@@ -1197,3 +1207,73 @@ export function neededStatTypeForRestrictionInput(
...
+  const identityKeys = accounts.map(({ address }) => stringToAccountId(address, context));
+
+  if (callback) {
+    return identityQuery.keyRecords.multi(identityKeys, result => {
+      return callback(assembleResult(result));
+    });
+  }
+
+  const rawResults = await identityQuery.keyRecords.multi(identityKeys);
+  return assembleResult(rawResults);
```
monitz87, line 1278:
```suggestion
      const record = optKeyRecord.unwrap();
      
      if (record.isSecondaryKey) {
        const [rawIdentityId, rawPermissions] = record.asSecondaryKey;
        
        if (identity && identityIdToString(rawIdentityId) !== identity.did) {
          return result;
        }

        result.push({
          account,
          permissions: meshPermissionsToPermissions(rawPermissions, context),
        });
      }

      return result;
    }, []);
  };
  
  const identityKeys = accounts.map(({ address }) => stringToAccountId(address, context));

  if (callback) {
    return identityQuery.keyRecords.multi(identityKeys, result => {
      return callback(assembleResult(result));
    });
  }

  const rawResults = await identityQuery.keyRecords.multi(identityKeys);
  
  return assembleResult(rawResults);
```

Just nitpicking

monitz87 (changes requested):

src/api/procedures/modifySignerPermissions.ts
polymath-eric: While that would work and give us consistent behavior with what we have, it seems like it would be a weird function to have. If it was only checking length then it should be called something like `assertSameLength`. Which seems too simple to me.

What about dropping it entirely and doing the check in line in the procedures instead?

polymath-eric: What about the `data` field in the error? I think we need this function if we want to tell the user which one is missing, where the length compare isn't going to do that. For now I am leaning towards leaving it in.

I think there is a simpler way to write the comparison inside the function though.

polymath-eric: With a functional style I have:

```typescript
  const addresses = accounts.map(({ address }) => address);
  const secondaryAddresses = secondaryAccounts.map(({ account: { address } }) => address);
  const notInTheList = addresses.filter(address => !secondaryAddresses.includes(address));
```

monitz87 (approved):

PR #794, opened 2022-07-19
chore: 🤖 Add support for Middleware V2 Enum values

prashantasdeveloper:
### Description

* Updated Middleware V2 enum values
* Updated GQL queries to support respective Enum fields
* Updated attribute types in all V2 functions to point to Middleware V2 Enums

### JIRA Link

DA-370

### Checklist

### DIFF
--- a/src/middleware/queriesV2.ts
+++ b/src/middleware/queriesV2.ts
@@ -251,27 +251,35 @@ export function instructionsQuery(
 type EventArgs = 'moduleId' | 'eventId' | 'eventArg0' | 'eventArg1' | 'eventArg2';
 
 /**
- *  @hidden
+ * Create args and filters to be supplied to GQL query
+ *
+ * @param filters filters to be applied
+ * @param typeMap Map defining the types corresponding to each attribute.
+ * All missing attributes whose types are not defined are considered to be String
+ *
+ * @hidden
  */
-function createEventFilters(attributes: QueryArgs<Event, EventArgs>): {
+function createArgsAndFilters(
+  filters: Record<string, unknown>,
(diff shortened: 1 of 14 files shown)


src/middleware/queriesV2.ts
```diff
@@ -259,10 +259,12 @@ function createEventFilters(attributes: QueryArgs<Event, EventArgs>): {
 } {
   const args: string[] = ['$start: Int', '$size: Int'];
   const filters: string[] = [];
+  const typeMap: Record<string, string> = { moduleId: 'ModuleIdEnum', eventId: 'EventIdEnum' };
```
monitz87, line 262: These `createXFilter` functions are similar enough that it merits using a single `createQueryParams` or `createArgsAndFilters` function which takes the attributes and the type map as parameters. If you need to call the function more than once with the same type map, just assign it to a variable (or curry it if you want)

monitz87 (changes requested):

src/middleware/queriesV2.ts
```diff
@@ -251,27 +251,35 @@ export function instructionsQuery(
 type EventArgs = 'moduleId' | 'eventId' | 'eventArg0' | 'eventArg1' | 'eventArg2';
 
 /**
- *  @hidden
+ * Create args and filters to be supplied to GQL query
+ *
+ * @param filters filters to be applied
+ * @param typeMap Map defining the types corresponding to each attribute.
```
monitz87, line 257:
```suggestion
 * @param filters - filters to be applied
 * @param typeMap - Map defining the types corresponding to each attribute
```

monitz87 (approved):

PR #797, opened 2022-07-21
test: 💍 separate the two expectations into different tests to reduce ci flakiness on affirmation tests

polymath-eric:
### Description

I've been noticing my PR test runs have been flaky due to a particular test case.

For now I doubled the time for it to hopefully prevent the flakiness.

### Breaking Changes

N/A

### JIRA Link

[DA-380](https://polymath.atlassian.net/browse/DA-380)

### Checklist


monitz87 (commented):
I think I'd rather separate the two expectations into different tests and see how that turns out (instead of increasing the timeout)

polymath-eric:
Fair enough. I was considering that too, but figured the timeout was less work. 

I am not sure this will solve the issue, it would be this case for the few times I saw it, but Prashant had an example of another case. 

I have never been able to reproduce the error locally without lowering the timeout.

VictorVicente (approved):

PR #801, opened 2022-07-27
fix: 🐛 Correct validation for removing asset requirements

prashantasdeveloper:
### Description

* Correct the condition to check if given requirement id exists in existing requirements for an Asset 

### JIRA Link

[DA-384](https://polymath.atlassian.net/browse/DA-384)

### Checklist

### DIFF
--- a/src/api/procedures/removeAssetRequirement.ts
+++ b/src/api/procedures/removeAssetRequirement.ts
@@ -37,7 +37,7 @@ export async function prepareRemoveAssetRequirement(
 
   const { requirements } = await query.complianceManager.assetCompliances(rawTicker);
 
-  if (requirements.filter(({ id: rawId }) => u32ToBigNumber(rawId).eq(reqId)).length) {
+  if (!requirements.filter(({ id: rawId }) => u32ToBigNumber(rawId).eq(reqId)).length) {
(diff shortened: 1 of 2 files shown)


src/api/procedures/removeAssetRequirement.ts
```diff
@@ -37,7 +37,7 @@ export async function prepareRemoveAssetRequirement(
 
   const { requirements } = await query.complianceManager.assetCompliances(rawTicker);
 
-  if (requirements.filter(({ id: rawId }) => u32ToBigNumber(rawId).eq(reqId)).length) {
+  if (!requirements.filter(({ id: rawId }) => u32ToBigNumber(rawId).eq(reqId)).length) {
```
monitz87, line 40: `find` makes more sense here

monitz87 (changes requested):

src/api/procedures/removeAssetRequirement.ts
prashantasdeveloper: Since we don't actually need the value, I guess `some` fits better here

monitz87 (approved):

VictorVicente (approved):

PR #807, opened 2022-08-04
feat: 🎸 allow for exempted identities to be removed

polymath-eric:
### Description
when a user does not pass exempted identities that are exempt when
calling setTransferRestrictions the procedure will unset those
permissions

### Breaking Changes

BREAKING CHANGE: 🧨 non passed exempted identities are removed when calling
setTransferRestriction

### JIRA Link

[DA-317](https://polymath.atlassian.net/browse/DA-317)

### Checklist


src/api/entities/Asset/TransferRestrictions/ClaimCount.ts
```diff
@@ -36,6 +36,9 @@ export class ClaimCount extends TransferRestrictionBase<TransferRestrictionType.
   /**
    * Sets all Claim Count Transfer Restrictions on this Asset
    *
+   * @note this method set exempted Identities for restrictions as well. If an Identity is currently exempted from a Claim Count Transfer Restriction
+   * but not passed into this call then they will be removed
+   *
```
prashantasdeveloper, line 41:
```suggestion
   * @note this method sets exempted Identities for restrictions as well. If an Identity is currently exempted from a Claim Count Transfer Restriction
   * but not passed into this call, then it will be removed
   *
```

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,128 @@ import {
...
+type ClaimKey = keyof typeof ClaimType | 'None';
+type ExemptionRecords = Partial<Record<ClaimKey, Identity[]>>;
+
+const addExemptionIfNotPresent = (
+  toInsertId: Identity,
+  exemptionRecords: ExemptionRecords,
+  claimType: ClaimKey,
+  filterSet: Identity[] = []
+): void => {
+  const found = filterSet.find(currentId => currentId.did === toInsertId.did);
```
prashantasdeveloper, line 60:
```suggestion
  const found = filterSet.some(currentId => currentId.isEqual(toInsertId));
```
I think `some` makes more sense here, since we don't actually need the matching value. Also, we can directly use `isEqual` (can leave for direct did assertion as well)

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,128 @@ import {
...
+    conditions.push(rawCondition);
 
-    if (r.exemptedIdentities) {
-      r.exemptedIdentities.forEach(e => {
-        exemptions.push(e);
-      });
-    }
+    r.exemptedIdentities?.forEach(e => {
+      const key = claimType || 'None';
+      const id = typeof e === 'string' ? new Identity({ did: e }, context) : e;
```
prashantasdeveloper, line 121:
```suggestion
      const id = asIdentity(e, context);
```

src/utils/conversion.ts
```diff
@@ -3832,6 +3848,20 @@ export function statTypeToStatOpType(
...
+/**
+ * @hidden
+ */
+export function transferRestrictionTypeToStatOpType(
+  type: TransferRestrictionType,
+  context: Context
+): PolymeshPrimitivesStatisticsStatOpType {
+  if (type === TransferRestrictionType.Count || type === TransferRestrictionType.ClaimCount) {
+    return statisticsOpTypeToStatOpType(StatisticsOpType.Count, context);
+  } else {
```
prashantasdeveloper, line 3860: nit -> we don't explicitly need `else` keyword as we are returning from each branch

prashantasdeveloper (changes requested):
Left some nit comments and doc changes

src/api/entities/Asset/TransferRestrictions/ClaimCount.ts
monitz87: There is another very similar comment on the percentage side that should also be addressed

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -147,31 +195,40 @@ export async function prepareSetTransferRestrictions(
...
-        feeMultiplier: new BigNumber(exemptions.length),
-        args: [true, exemptKey, btreeIds],
+        feeMultiplier: new BigNumber(exemptedBtreeSet.size),
+        args: [true, exemptKey, exemptedBtreeSet],
       })
     );
-  }
+  });
+
+  Object.entries(toRemoveExemptions).forEach(([claimType, unExempted]) => {
```
monitz87, line 219: lodash `forEach`

monitz87: Also, these two blocks look very similar, you could probably refactor them into a named arrow function

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,128 @@ import {
...
   compareTransferRestrictionToInput,
-  getExemptedIds,
   neededStatTypeForRestrictionInput,
 } from '~/utils/internal';
 
 /**
  * @hidden
  */
+
+type ClaimKey = keyof typeof ClaimType | 'None';
```
monitz87, line 51: Shouldn't this be limited to the 3 claim types that are allowed in transfer restrictions?

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -286,14 +343,55 @@ export async function prepareStorage(
...
+      StorageKey<
+        [
+          PolymeshPrimitivesTransferComplianceTransferConditionExemptKey,
+          PolymeshPrimitivesIdentityId
+        ]
+      >,
+      bool
+    ][]
+  >[] = [];
+  claimTypes.forEach(claimType => {
```
monitz87, line 363:
```suggestion
  >[] = [];
  
  claimTypes.forEach(claimType => {
```

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -286,14 +343,55 @@ export async function prepareStorage(
...
+      statistics.transferConditionExemptEntities.entries({
+        asset: tickerKey,
+        op,
+        claimType,
+      })
+    );
+  });
+  const rawCurrentExemptions = flatten(await Promise.all(exemptionFetchers));
+  const currentExemptions: ExemptionRecords = {};
+  rawCurrentExemptions.forEach(([key]) => {
```
monitz87, line 374:
```suggestion
  });
  
  const rawCurrentExemptions = flatten(await Promise.all(exemptionFetchers));
  const currentExemptions: ExemptionRecords = {};
  
  rawCurrentExemptions.forEach(([key]) => {
```

You should destructure key and args here

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -286,14 +343,55 @@ export async function prepareStorage(
...
+    const { claimType: rawClaimType } = key.args[0];
+    const did = identityIdToString(key.args[1]);
+    let claimType: ClaimKey = 'None';
+    if (rawClaimType.isSome) {
```
monitz87, line 378:
```suggestion
    let claimType: ClaimKey = 'None';
    
    if (rawClaimType.isSome) {
```

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -286,14 +343,55 @@ export async function prepareStorage(
...
+      claimType = meshClaimTypeToClaimType(rawClaimType.unwrap());
+    }
+    const identity = new Identity({ did }, context);
+    addExemptionIfNotPresent(identity, currentExemptions, claimType);
```
monitz87, line 382:
```suggestion
    }
    
    const identity = new Identity({ did }, context);
    
    addExemptionIfNotPresent(identity, currentExemptions, claimType);
```

src/utils/conversion.ts
```diff
@@ -2980,11 +2981,26 @@ export function transferRestrictionToPolymeshTransferCondition(
...
 ): BTreeSet<PolymeshPrimitivesIdentityId> {
-  return context.createType('BTreeSet<PolymeshPrimitivesIdentityId>', scopeIds);
+  const rawIds = identities.map(({ did }) => stringToIdentityId(did, context));
+  return context.createType('BTreeSet<PolymeshPrimitivesIdentityId>', rawIds);
+}
+
+/**
+ * @hidden
+ */
+export async function getExemptedBtreeSet(
```
monitz87, line 2995: This one seems more like a procedure/internal util than conversion

src/utils/conversion.ts
```diff
@@ -2980,11 +2981,26 @@ export function transferRestrictionToPolymeshTransferCondition(
...
+  identities: (string | Identity)[],
+  ticker: string,
+  context: Context
+): Promise<BTreeSet<PolymeshPrimitivesIdentityId>> {
+  const exemptedIds = await getExemptedIds(identities, context, ticker);
+  const toIdentity = (did: string): Identity => new Identity({ did }, context);
```
monitz87, line 3001: You can use the existing `asIdentity` here

src/utils/conversion.ts
```diff
@@ -2980,11 +2981,26 @@ export function transferRestrictionToPolymeshTransferCondition(
...
+  const mapped = exemptedIds.map(toIdentity);
+  return identitiesToBtreeSet(mapped, context);
```
monitz87, line 3003:
```suggestion
  const mapped = exemptedIds.map(toIdentity);
  
  return identitiesToBtreeSet(mapped, context);
```

src/utils/conversion.ts
```diff
@@ -2980,11 +2981,26 @@ export function transferRestrictionToPolymeshTransferCondition(
...
  */
-export function scopeIdsToBtreeSetIdentityId(
-  scopeIds: PolymeshPrimitivesIdentityId[],
+export function identitiesToBtreeSet(
+  identities: Identity[],
   context: Context
 ): BTreeSet<PolymeshPrimitivesIdentityId> {
-  return context.createType('BTreeSet<PolymeshPrimitivesIdentityId>', scopeIds);
+  const rawIds = identities.map(({ did }) => stringToIdentityId(did, context));
+  return context.createType('BTreeSet<PolymeshPrimitivesIdentityId>', rawIds);
```
monitz87, line 2989:
```suggestion
  const rawIds = identities.map(({ did }) => stringToIdentityId(did, context));
  
  return context.createType('BTreeSet<PolymeshPrimitivesIdentityId>', rawIds);
```

monitz87 (changes requested):

src/api/procedures/setTransferRestrictions.ts
polymath-eric: On chain it's an option type, with `None` being for `Count` and `Percentage`.

There might be a better way than having a magic string of 'None' though

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,128 @@ import {
...
 
-  if (!someDifference) {
+  Object.entries(currentExemptions).forEach(([cType, currentIds]) => {
+    const given = givenExemptions[cType as ClaimKey] || [];
+    currentIds.forEach(id => {
+      addExemptionIfNotPresent(id, toRemove, cType as ClaimKey, given);
+    });
+  });
+
+  const sizeOf = (obj: Record<string, unknown>): number => Object.keys(obj).length;
```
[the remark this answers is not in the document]
polymath-eric: `forEach` was adding a `| undefined` type vs `Object.entries(...).forEach` giving the type.

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,128 @@ import {
...
   currentRestrictions: TransferCondition[];
+  currentExemptions: ExemptionRecords;
   occupiedSlots: BigNumber;
 }
-
 /**
  * @hidden
  */
-function transformRestrictions(
+function transformInput(
```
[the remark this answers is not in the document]
polymath-eric: looking at this now

src/api/procedures/setTransferRestrictions.ts
monitz87: What I meant was having `type ClaimKey = ClaimType.Accredited | ClaimType.Affiliate | ClaimType.Jurisdiction | 'None'`

src/api/procedures/setTransferRestrictions.ts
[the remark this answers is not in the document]
monitz87: If you do what I specified in https://github.com/PolymeshAssociation/polymesh-sdk/pull/807#discussion_r943732849, you can remove the `Partial` from the `ExemptionRecords` type, which will remove the `| undefined` from the `forEach`. This will mean that you will have to initialize your `ExemptionRecord` objects with empty arrays for each property and alter the logic slightly. It might seem a bit unnecessary but you'll get cleaner types

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -203,31 +211,23 @@ export async function prepareSetTransferRestrictions(
...
+  const pushExemptions = (exemptions: ExemptionRecords, exempt: boolean): void => {
+    Object.entries(exemptions).forEach(([claimType, exempted]) => {
+      const rawClaimType = claimType === 'None' ? undefined : (claimType as ClaimType);
+      const exemptKey = toExemptKey(tickerKey, op, rawClaimType);
+      const exemptedBtreeSet = identitiesToBtreeSet(exempted, context);
+      transactions.push(
+        checkTxType({
+          transaction: statistics.setEntitiesExempt,
+          feeMultiplier: new BigNumber(exemptedBtreeSet.size),
+          args: [exempt, exemptKey, exemptedBtreeSet],
```
monitz87, line 223: Just for consistency, we should use `booleanToBool` for `exempt` here

monitz87 (changes requested):

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -18,88 +25,136 @@ import {
...
+      addExemptionIfNotPresent(id, toSetExemptions, key, currentIds);
+    });
   });
 
-  if (!someDifference) {
+  // calculate exemptions to remove — for each current exemption check if it was in the input
+  Object.entries(currentExemptions).forEach(([cType, currentIds]) => {
+    const given = inputExemptions[cType as ClaimKey] || [];
+    currentIds.forEach(id => {
+      addExemptionIfNotPresent(id, toRemoveExemptions, cType as ClaimKey, given);
```
monitz87, line 144:
```suggestion
    const key = cType as ClaimKey;
    const given = inputExemptions[key] || [];
    currentIds.forEach(id => {
      addExemptionIfNotPresent(id, toRemoveExemptions, key, given);
```

monitz87 (changes requested):

src/api/procedures/setTransferRestrictions.ts
```diff
@@ -116,37 +121,40 @@ function transformInput(
     const rawCondition = transferRestrictionToPolymeshTransferCondition(condition, context);
     conditions.push(rawCondition);
 
-    r.exemptedIdentities?.forEach(e => {
+    restriction.exemptedIdentities?.forEach(exemption => {
       const key = claimType || 'None';
-      const id = typeof e === 'string' ? new Identity({ did: e }, context) : e;
-      addExemptionIfNotPresent(id, givenExemptions, key);
+      const id = asIdentity(exemption, context);
+      addExemptionIfNotPresent(id, inputExemptions, key);
```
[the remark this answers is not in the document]
polymath-eric: I think I could make exemption record be

`type ExemptionRecords = Record<ClaimKey, Set<Identity>>;`

and then use Set properties to keep it unique, and not need the method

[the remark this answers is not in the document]
polymath-eric: Actually Set would only work for primitives. We want uniquness on `did`, but it will do object reference.

Will rename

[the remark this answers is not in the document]
monitz87: Can't you define a custom comparator for sets?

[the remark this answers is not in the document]
polymath-eric: I looked at some articles and they say its not implemented, although there was a proposal to add it.

The solutions involve making your own Set, which is more involved / complicated than having the method.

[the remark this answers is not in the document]
polymath-eric: https://stackoverflow.com/questions/29759480/how-to-customize-object-equality-for-javascript-set

Update March 2022

prashantasdeveloper (approved):

monitz87 (changes requested):

src/utils/conversion.ts
```diff
@@ -3798,23 +3797,24 @@ export function statUpdatesToBtreeStatUpdate(
...
+      return StatType.ScopedBalance;
     }
   }
 }
 
 /**
  * @hidden
+ * @param type - On chain "Scoped" or not is determined by the presence of the claim, and not be distinct types
  */
 export function statisticsOpTypeToStatOpType(
```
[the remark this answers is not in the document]
polymath-eric: I inlined the createType instead

src/utils/conversion.ts
```diff
@@ -3795,29 +3795,20 @@ export function statUpdatesToBtreeStatUpdate(
...
-/**
- * @hidden
- * @param type - On chain "Scoped" or not is determined by the presence of the claim, and not be distinct types
- */
-export function statisticsOpTypeToStatOpType(
-  type: StatType.Count | StatType.Balance,
-  context: Context
-): PolymeshPrimitivesStatisticsStatOpType {
-  return context.createType('PolymeshPrimitivesStatisticsStatOpType', type);
+  if (rawStat.op.type === 'Count') {
```
monitz87, line 3807: We could use some destructuring here

monitz87 (changes requested):

monitz87 (approved):
LGTM

PR #813, opened 2022-08-10
feat: 🎸 Asset metadata

prashantasdeveloper:
### Description

* Add `Metadata` namespace to manager metadata related functionality . This namespace has the following members - 
  * `get` - method to get all (global + local) metadata for an asset [[DA-389](https://polymath.atlassian.net/browse/DA-389)]
  * `getOne` - method to fetch a specific metadata by its id & type [[DA-390](https://polymath.atlassian.net/browse/DA-390)]
  * `register` - procedure to register a new local asset metadata. This also allows optionally to set the value of metadata while registering [[DA-388](https://polymath.atlassian.net/browse/DA-388)]
* Add `MetadataEntry` entity to manage a specific metadata. It contains the following - 
  * `ticker` - asset to which this metadata belong
  * `type` - global or local type of metadata
  * `id` - unique ID of the metadata for a given ticker and type 
  * `details` - method to get name and specs for the metadata entry [[DA-393](https://polymath.atlassian.net/browse/DA-393)]
  * `value` - method to get value of the metadata with extra details about the value namely `expiry`, `lockStatus` and `lockedUntil`[[DA-387](https://polymath.atlassian.net/browse/DA-387)]
  * `set` - procedure to set value of the metadata [[DA-391](https://polymath.atlassian.net/browse/DA-391)]

### JIRA Link

[DA-387](https://polymath.atlassian.net/browse/DA-387), [DA-392](https://polymath.atlassian.net/browse/DA-392), [DA-393](https://polymath.atlassian.net/browse/DA-393), [DA-388](https://polymath.atlassian.net/browse/DA-388), [DA-389](https://polymath.atlassian.net/browse/DA-389), [DA-390](https://polymath.atlassian.net/browse/DA-390), [DA-391](https://polymath.atlassian.net/browse/DA-391)

### Checklist


src/api/entities/Asset/types.ts
```diff
@@ -57,5 +57,16 @@ export interface AgentWithGroup {
...
-export * from './CorporateActions/types';
+export interface AssetMetadataSpec {
+  url?: string;
+  description?: string;
+  typeDef?: string;
+}
+export interface AssetGlobalMetadata {
+  id: BigNumber;
+  name: string;
+  specs?: AssetMetadataSpec;
```
polymath-eric, line 68:
```suggestion
  specs: AssetMetadataSpec;
```
I think specs are required even if every field might be `None`.

src/api/client/Assets.ts
```diff
@@ -250,4 +257,77 @@ export class Assets {
...
+
+            if (rawDescription.isSome) {
+              specs.description = bytesToString(rawDescription.unwrap());
+            }
+
+            if (rawTypeDef.isSome) {
+              specs.typeDef = bytesToString(rawTypeDef.unwrap());
+            }
+
+            if (Object.keys(specs).length > 0) {
```
polymath-eric, line 323: I think its OK to assign an empty object to specs. Let me know if there is something I am not considering.

src/api/client/Assets.ts
```diff
@@ -250,4 +257,77 @@ export class Assets {
...
+          args: [rawId],
+        },
+        rawName,
+      ]) => {
+        const metadata: AssetGlobalMetadata = {
+          id: u64ToBigNumber(rawId),
+          name: bytesToString(rawName.unwrap()),
+        };
+
+        const specEntry = specsEntries.find(
```
polymath-eric, line 293: Asked this on Slack. If so, it will be more performant to use the index offset directly instead of iterating with find on every loop. Since we have lodash there is `zip` that could group them if they are ordered the same.

Otherwise build up a `Record<rawId, Spec>` before this loop.

With `find` in map this becomes O(n<sup>2</sup>), which I think we can avoid and make this function more readable.


https://polymathnetwork.slack.com/archives/C01RLA867C5/p1660750728813059

polymath-eric (changes requested):
overall looks good

src/testUtils/mocks/entities.ts
```diff
@@ -865,6 +880,53 @@ const MockAssetClass = createMockEntityClass<AssetOptions>(
...
+     */
+    public argsToOpts(...args: ConstructorParameters<typeof MetadataEntry>) {
+      return extractFromArgs(args, ['id', 'ticker', 'type']);
+    }
+
+    /**
+     * @hidden
+     */
+    public configure(opts: Required<MetadataEntryOptions>) {
+      this.uuid = 'sto';
```
polymath-eric, line 903: (nit) I think destructure on opts would clean this up a bit.

src/utils/conversion.ts
```diff
@@ -4067,3 +4074,189 @@ export function inputStatTypeToMeshStatType(
...
+      message: '"typeDef" length exceeded for given Asset Metadata spec',
+      data: {
+        maxLength: metadataTypeDefMaxLength,
+      },
+    });
+  }
+
+  return context.createType('PolymeshPrimitivesAssetMetadataAssetMetadataSpec', {
+    url: optionize(stringToBytes)(url, context),
+    description: optionize(stringToBytes)(description, context),
```
polymath-eric, line 4109: Is there a max length for description?

src/utils/conversion.ts
```diff
@@ -4067,3 +4074,189 @@ export function inputStatTypeToMeshStatType(
...
+    typeDef: optionize(stringToBytes)(typeDef, context),
+  });
+}
+
+/**
+ * @hidden
+ */
+export function meshMetadataSpecToMetadataSpec(
+  rawSpecs?: Option<PolymeshPrimitivesAssetMetadataAssetMetadataSpec>
```
polymath-eric, line 4118: Do we need `?` and `Option`? I would have though the Option covered the undefined case.

src/utils/conversion.ts
```diff
@@ -4067,3 +4074,189 @@ export function inputStatTypeToMeshStatType(
...
+}
+
+/**
+ * @hidden
+ */
+export function meshMetadataValueToMetadataValue(
+  rawValue: Option<Bytes>,
+  rawDetails: Option<PolymeshPrimitivesAssetMetadataAssetMetadataValueDetail>
+): MetadataValue | null {
+  if (rawValue.isEmpty) {
```
polymath-eric, line 4167: (style) should it be `isNone` to indicate an Option check vs the generic `isEmpty` thats on everything?

Same behavior either way.

src/utils/conversion.ts
```diff
@@ -4067,3 +4074,189 @@ export function inputStatTypeToMeshStatType(
...
+  }
+
+  return context.createType('PolymeshPrimitivesAssetMetadataAssetMetadataKey', metadataKey);
+}
+
+/**
+ * @hidden
+ */
+export function meshMetadataValueToMetadataValue(
+  rawValue: Option<Bytes>,
```
polymath-eric, line 4164: should these be passed as a single object?

polymath-eric (commented):
Good work. Left a few smaller comments, mostly nitty stuff.

src/api/client/Assets.ts
```diff
@@ -250,4 +257,53 @@ export class Assets {
...
+      context: {
+        polymeshApi: {
+          query: {
+            asset: { assetMetadataGlobalKeyToName, assetMetadataGlobalSpecs },
+          },
+        },
+      },
+    } = this;
+
+    const [keyToNameEntries, specsEntries] = await Promise.all([
```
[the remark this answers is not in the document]
prashantasdeveloper: I tried the queryMulti but it was getting tougher to make it type safe since the query fetch all entries instead of a specific value. I too saw that comment from Jere suggesting queryMulti after which I changed it at one of the places but wasn't able to replace this. Maybe later I'll give it another try

src/api/client/Assets.ts
```diff
@@ -250,4 +257,53 @@ export class Assets {
       next,
     };
   }
+
+  /**
+   * Retrieve all the Asset Global Metadata on chain.
+   * This includes metadata id, name and optional specs
+   */
+  public async getGlobalMetadataKeys(): Promise<GlobalMetadataKey[]> {
```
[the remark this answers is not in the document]
prashantasdeveloper: I think we can provide pagination but I don't think are going to be a lot of global asset metadata keys. Felt an overkill just to provide an option which very rarely will be used.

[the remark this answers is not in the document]
prashantasdeveloper: But still if you think this might be useful I can push in the change. Lets see what Victory thinks here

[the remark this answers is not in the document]
polymath-eric: I think it should be OK to leave. I think only the sudo user can add them also.

src/api/procedures/__tests__/setMetadata.ts
```diff
@@ -0,0 +1,290 @@
...
+  });
+
+  it('should return a set asset metadata transaction spec', async () => {
+    const proc = procedureMockUtils.getInstance<Params, MetadataEntry>(mockContext);
+
+    params = {
+      value: 'SOME_VALUE',
+      metadataEntry,
+    };
+    const metadataValueToMeshMetadataValueStub = sinon.stub(
```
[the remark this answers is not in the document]
prashantasdeveloper: Since this was only required in this I preferred putting it here.

src/utils/conversion.ts
prashantasdeveloper: No

src/utils/conversion.ts
prashantasdeveloper: We need `?`, otherwise we would have to handle the undefined case at the respective places from where this is invoked.

src/utils/conversion.ts
prashantasdeveloper: Since these were separate attributes every where else, I figured there wasn't that much of a need to put them as an object. Eventually we need destruct them here itself and this is used from one place only

src/api/client/Assets.ts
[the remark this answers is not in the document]
VictorVicente: Global metadata can be added by GC after some quorum. I agree it is not necessary to paginate this endpoint, at least now.

src/api/entities/MetadataEntry/index.ts
```diff
@@ -0,0 +1,178 @@
...
+    ]);
+
+    return meshMetadataValueToMetadataValue(rawValue, rawValueDetails);
+  }
+
+  /**
+   * Determine whether this MetadataEntry exists on chain
+   */
+  public async exists(): Promise<boolean> {
+    return true;
```
VictorVicente, line 163: Is it able to delete a MetadataEntry once configured?

VictorVicente (approved):
Awesome work Prashant!

src/api/entities/MetadataEntry/index.ts
prashantasdeveloper: Not right now

VictorVicente (approved):

polymath-eric (approved):

PR #822, opened 2022-08-21
fix: 🐛 Fix transfer with memo

polymath-eric:
### Description

Fix transfer with memo

### Breaking Changes

N/A

### Checklist

### DIFF
--- a/src/utils/__tests__/conversion.ts
+++ b/src/utils/__tests__/conversion.ts
@@ -515,7 +515,9 @@ describe('portfolioMovementToMovePortfolioItem', () => {
 
     expect(result).toBe(fakeResult);
 
-    context.createType.withArgs('Memo', padString(memo, 32)).returns(rawMemo);
+    context.createType
+      .withArgs('PolymeshCommonUtilitiesBalancesMemo', memo.padEnd(32))
+      .returns(rawMemo);
 
     context.createType
       .withArgs('MovePortfolioItem', {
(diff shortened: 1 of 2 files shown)


prashantasdeveloper:
Seems like this got left out with upgrade. My only concern here is if we are switching the `padString` with `padEnd` shouldn't we just do that in `conversion` util func? That should take care of any other possibility even if that gets overlooked by us

VictorVicente (approved):

polymath-eric:
@prashantasdeveloper What do you mean by:

> Seems like this got left out with upgrade. My only concern here is if we are switching the padString with padEnd shouldn't we just do that in conversion util func? That should take care of any other possibility even if that gets overlooked by us

The changes are in conversion, and the test file for conversion only

prashantasdeveloper:
My bad. I meant directly updating the `padString` function present in `internal.ts`. Are we sure other instances where `padString` is used aren't affected?

prashantasdeveloper (approved):

polymath-eric:
Made a ticket for Prashant's suggetion. For now I want to focus on webhooks rather then test all the callers of that method.

https://polymath.atlassian.net/browse/DA-403

F-OBrien:
> My bad. I meant directly updating the `padString` function present in `internal.ts`. Are we sure other instances where `padString` is used aren't affected? Maybe pick that up later and create a JIRA ticket instead.

Did we establish if the `padString` with null characters or the `Memo` createType was the root cause? `padString` is used in multiple other places. Primarily on tickers so editing it potentially has wider implications than just addressing memos.

As an aside, I prefer the way querying strings padded with spaces are reported in plain text instead of hex encoded strings when viewing events in Subscan and using the polkadot.js app interface. So if there is no reason to keep padding with null characters it might be worth considering spaces as the default padding.

PR #823, opened 2022-08-22
chore: 🤖 improve `assembleBatchTransaction` typing

monitz87:
### Description

This PR adds the `readonly` modifier to some of the types surrounding `assembleBatchTransactions`. This removes the need to use `tuple`, which can be replaced with `as const`. While in some cases it might require more explicit typing, it's idiomatic typescript and should be easier to understand for new devs

### Checklist


polymath-eric (approved):

VictorVicente (approved):

monitz87:
Closed in favor of #832

PR #831, opened 2022-08-25
fix: 🐛 fix bug where only the default signer was used

polymath-eric:
### Description

in Procedure `prepare` method, the passed in, non modified `context` was
used, when `ctx` was supposed to be used instead

### JIRA Link

[DA-405](https://github.com/PolymeshAssociation/polymesh-sdk/pull/831#issuecomment-1227689597)

### Checklist

### DIFF
--- a/src/base/Procedure.ts
+++ b/src/base/Procedure.ts
@@ -339,8 +339,8 @@ export class Procedure<Args = void, ReturnValue = void, Storage = Record<string,
 
       const procedureResult = await prepareTransactionsPromise;
 
-      const signingAddress = context.getSigningAddress();
-      const signer = context.getExternalSigner();
+      const signingAddress = ctx.getSigningAddress();
+      const signer = ctx.getExternalSigner();
 
       const spec = {
         ...procedureResult,


polymath-eric:
alternatively, we could change

`const ctx = await this.setup(procArgs, context, opts);`  to
`context = await this.setup(procArgs, context, opts);`

overriding the passed in context, so it would be harder to use by mistake. It should be safe since the given variable is a pointer and should leave the caller context unchanged as `setup` clones it.

polymath-eric:
Since the REST API lets the SDK initialize first and then set account it would always throw

```
InternalServerErrorException: There is no signing Account associated with the SDK instance
```

Which is probably better behavior than only using the same signer no matter what its told to use.

monitz87 (approved):
Can't believe this one got through unnoticed for so long

polymath-eric:
> Can't believe this one got through unnoticed for so long

It was only in alpha branch besides the eliminate TQ PR.

Usually with sandbox I always init the signer before the SDK so I probably wouldn't have caught it either. It also looks right when reviewing it.

PR #833, opened 2022-08-26
Update middleware types 

prashantasdeveloper:
### Description

* Update middleware types and fix ordering for native SQ queries

### Breaking Changes

NA

### JIRA Link

[DA-253](https://polymath.atlassian.net/browse/DA-253)

### Checklist


polymath-eric (approved):
`src/middleware/typesV2.ts` seems big to me at 31K lines.

Is that coming from SubQuery? Any ideas on being able to use a smaller file instead somehow?

prashantasdeveloper:
> src/middleware/typesV2.ts seems big to me at 31K lines.
Is that coming from SubQuery? Any ideas on being able to use a smaller file instead somehow?

Yeah it is coming from SQ, I am not sure what changed but apparently the unsafe aggregate functions got exposed somehow (without even specifying), which caused multiple new options to be generated. 
For now, I have used the staging-native GQL to generate the types and seems to have reduced again back to previous form. I think once I updated my local subql-node version that may have caused it, but I'm not sure how those got exposed.  (Can you try it in your local machine once @polymath-eric ?)

polymath-eric (approved):
types look better.

Running it myself against my polymesh-local gets something similar to whats committed here, but there is a different of linespaces.

PR #852, opened 2022-09-16
chore: 🤖 use api.at(<blockhash>) for querying old blocks

prashantasdeveloper:
### Description

Methods such as `<query>.at(..args)` and `<query>.entriesAt(...args)` were deprecated from Polkadot in favour of first getting a decorated api at any block with `api.at(<blockHash>)` and then using the `query` module within it to get the desired result. This change updates all occurrence to use the latest logic

### Breaking Changes

NA

### JIRA Link

[DA-410](https://polymesh.atlassian.net/browse/DA-410)

### Checklist

### DIFF
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -82,7 +82,7 @@ import {
   u16ToBigNumber,
   u32ToBigNumber,
 } from '~/utils/conversion';
-import { assertAddressValid, calculateNextKey, createClaim } from '~/utils/internal';
+import { assertAddressValid, calculateNextKey, createClaim, getApiAtBlock } from '~/utils/internal';
 
 interface ConstructorParams {
   polymeshApi: ApiPromise;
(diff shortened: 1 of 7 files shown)


src/base/Context.ts
```diff
@@ -190,13 +190,18 @@ export class Context {
...
+        query: { system },
       },
+      polymeshApi,
     } = this;
 
     try {
       const blockHash = await system.blockHash(bigNumberToU32(new BigNumber(0), this));
-      const balance = await balances.totalIssuance.at(blockHash);
+
+      const apiAt = await polymeshApi.at(blockHash);
```
VictorVicente, line 201: Why this one is not using `getApiAtBlock`?

VictorVicente (commented):
You've make it very simple. Congrats @prashantasdeveloper

VictorVicente (approved):

PR #857, opened 2022-09-21
feat: 🎸 add result to successfully ran transactions

polymath-eric:
### Description

Adds a getter `result` to PolymeshTransactionBase class. this will have the result resolved from run() so it can be accessed later

### Breaking Changes

None

### JIRA Link

✅ Closes: DA-407

### Checklist


polymath-eric:
I based off of beta, although this should probably have been on alpha.

I think we should bring the commits from master into beta then alpha (or at least beta -> alpha). Leaving this as draft until then

polymath-eric:
rebasing onto alpha

PR #858, opened 2022-09-21
Merge pull request #855 from PolymeshAssociation/beta

polymath-eric:
### Description

Update beta with master

### Checklist


polymath-eric:
this seems a bit redundant to me, but it seems better to ensure beta contains every commit of master just in case

polymath-eric:
Lots of code smells I think we ignored previously

prashantasdeveloper (approved):

PR #866, opened 2022-10-05
DA-423 Update type def and min spec and chain version to 5.1.0

prashantasdeveloper:
### Description

* Update types and definitions using latest polymesh_schema
* Refactors all ConfidentialIdentity types with ConfidentialIdentityV2 types
* Pass `IdentityId` instead of `Signatory` for target attribute for `createGroupAndAddAuth` tx
* Use `signatory` from entry arguments instead of the query response in `multiSig.multiSigSigners` query
* Update chain and spec version supported by SDK to 5.1.0

### Breaking Changes

NA

### JIRA Link

DA-423

### Checklist

### DIFF
--- a/scripts/transactions.json
+++ b/scripts/transactions.json
@@ -63,7 +63,8 @@
     "remove_secondary_keys_old": "remove_secondary_keys_old",
     "placeholder_legacy_set_permission_to_signer": "placeholder_legacy_set_permission_to_signer",
     "add_secondary_keys_with_authorization_old": "add_secondary_keys_with_authorization_old",
-    "set_secondary_key_permissions": "set_secondary_key_permissions"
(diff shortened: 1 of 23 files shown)


VictorVicente (commented):
Do we know how `Refactors all ConfidentialIdentity types with ConfidentialIdentityV2 types` affect us?

prashantasdeveloper:
Quoting Robert here 
> We have two versions of the confidential_identity crate used in Polymesh.  Before they didn't have the version number in the "package" name.  Both of those types are the V2 version

Those got renamed that's it

polymath-eric:
Should we release the stuff in alpha first? 

I think we should have a release for SDK v5.0.x that has metadata features, and then get this merged in.

polymath-eric (approved):

PR #868, opened 2022-10-06
feat: 🎸 Allow option to redeem from specific portfolio

prashantasdeveloper:
### Description

`RedeemTokenParams` now supports `from` attribute where a specific portfolio can be specified to redeem any Asset

### Breaking Changes

NA

### JIRA Link

DA-422

### Checklist

### DIFF
--- a/src/api/procedures/redeemTokens.ts
+++ b/src/api/procedures/redeemTokens.ts
@@ -1,7 +1,13 @@
-import { Asset, DefaultPortfolio, PolymeshError, Procedure } from '~/internal';
+import BigNumber from 'bignumber.js';
+
+import { Asset, DefaultPortfolio, NumberedPortfolio, PolymeshError, Procedure } from '~/internal';
 import { ErrorCode, RedeemTokensParams, TxTags } from '~/types';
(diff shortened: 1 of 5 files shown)


src/api/procedures/redeemTokens.ts
```diff
@@ -47,9 +52,19 @@ export async function prepareRedeemTokens(
     });
   }
 
+  const rawAmount = bigNumberToBalance(amount, context, isDivisible);
+
+  if (from) {
```
polymath-eric, line 57: It seems like `redeemFromPortfolio` accepts a default Portfolio and `portfolioToPortfolioKind` handles defaults as well. 

We should be able to use only `redeemFromPortfolio` and simplify the code here a bit. Like `addInstructionWithMemo` always being used now.

polymath-eric: That said, if someone had a secondary key permissioned with only the old extrinsic, the way you have it is 100% backward compatible.

polymath-eric (approved):
LGTM

Left a comment about a potential simplification, but there might be benefits to using the old extrinsic just in case someone has a restrictive permission then its could potentially be a breaking change.

VictorVicente (approved):

PR #871, opened 2022-10-12
fix: 🐛 update MovePortfolioItem conversion to long type name

polymath-eric:
### Description

fixes issue where memo on portfolio movement would cause an error

### Breaking Changes

None

### JIRA Link

N/A

### Checklist

### DIFF
--- a/src/utils/conversion.ts
+++ b/src/utils/conversion.ts
@@ -14,7 +14,9 @@ import {
   PalletCorporateActionsDistribution,
   PalletCorporateActionsInitiateCorporateActionArgs,
   PalletMultisigProposalStatus,
+  PalletPortfolioMovePortfolioItem,
   PalletStoFundraiser,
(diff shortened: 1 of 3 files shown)


VictorVicente (approved):

src/utils/conversion.ts
```diff
@@ -1437,7 +1437,7 @@ export function authorizationDataToAuthorization(
 /**
  * @hidden
  */
-export function stringToMemo(value: string, context: Context): Memo {
+export function stringToMemo(value: string, context: Context): PolymeshCommonUtilitiesBalancesMemo {
```
polymath-eric, line 1440: This isn't strictly needed for the fix, but is one less schema type to worry about. I tested `transferPolyx` method with this as well.

prashantasdeveloper (approved):

PR #873, opened 2022-10-18
feat: 🎸 Add `Instruction.onStatusChange`

prashantasdeveloper:
### Description

Add `Instruction.onStatusChange`. This allows to subscribe on the status of an instruction

### Breaking Changes

NA

### JIRA Link

DA-301

### Checklist

### DIFF
--- a/src/api/entities/Instruction/index.ts
+++ b/src/api/entities/Instruction/index.ts
@@ -182,6 +182,8 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
    * Retrieve current status of the Instruction. This can be subscribed to know if instruction fails
    *
    * @note can be subscribed to
+   * @note current status as `Executed` means that the Instruction has been executed/rejected and pruned from
+   *   the chain.
    */
   public async onStatusChange(callback: SubCallback<InstructionStatus>): Promise<UnsubCallback> {
     const {
@@ -199,16 +201,14 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
     const assembleResult = (rawStatus: MeshInstructionStatus): InstructionStatus => {
       const status = meshInstructionStatusToInstructionStatus(rawStatus);
 
-      if (status === InternalInstructionStatus.Unknown) {
-        throw new PolymeshError({
-          code: ErrorCode.DataUnavailable,
-          message: executedMessage,
-        });
+      if (status === InternalInstructionStatus.Pending) {
+        return InstructionStatus.Pending;
+      } else if (status === InternalInstructionStatus.Failed) {
+        return InstructionStatus.Failed;
+      } else {
+        // TODO @prashantasdeveloper remove this once the chain handles Executed/Rejected state separately
+        return InstructionStatus.Executed;
       }
-
-      return status === InternalInstructionStatus.Pending
-        ? InstructionStatus.Pending
-        : InstructionStatus.Failed;
     };
 
     return instructionDetails(bigNumberToU64(id, context), ({ status }) => {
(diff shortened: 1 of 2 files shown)


src/api/entities/Instruction/index.ts
```diff
@@ -173,6 +176,44 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
...
+
+    const assembleResult = (rawStatus: MeshInstructionStatus): InstructionStatus => {
+      const status = meshInstructionStatusToInstructionStatus(rawStatus);
+
+      if (status === InternalInstructionStatus.Unknown) {
+        throw new PolymeshError({
+          code: ErrorCode.DataUnavailable,
+          message: executedMessage,
+        });
+      }
```
prashantasdeveloper, line 205: This isn't a specific way of knowing if Instruction was rejected/executed. One way to do so (when we receive status from chain as `Unknown`) can be to query middleware to check so, but maybe the middleware isn't all synced upto the block. 

In a hacky way maybe we add some delay in querying middleware but that's feel right. Subscription should update the status as soon as it changes otherwise the whole purpose gets destroyed. 

Let me know you thoughts on how you guys think we can handle the `Rejected/Executed` case @VictorVicente @polymath-eric

prashantasdeveloper: As per the task DA-301, what it sounds is that `Failed` status was the one that was most concerning but since its a generic status change function, I think it should tell us about all

VictorVicente: Difference between being rejected and executed is very big to not be able to know it. We should discuss with Robert and team how would it be the best way to track it. 
One alternative could be using events, but it's not my preferred by default one. 
Maybe there is some alternative on the scheduler side. I'll kickoff a discussion with Robert to find the best solution here.

polymath-eric: I thought we able to subscribe to the events with polkadot and when we see a block with it we can resolve the promise / call the callback? I think the subscription can narrow to event type, but sure if we could subscribe to a specific instruction. Been a while since I worked on bridge watcher but it should have an example.

We shouldn't use the middleware for this since it can add latency and make things feel slow.

src/api/entities/Instruction/index.ts
```diff
@@ -175,6 +178,44 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
...
+    const assembleResult = (rawStatus: MeshInstructionStatus): InstructionStatus => {
+      const status = meshInstructionStatusToInstructionStatus(rawStatus);
+
+      if (status === InternalInstructionStatus.Pending) {
+        return InstructionStatus.Pending;
+      } else if (status === InternalInstructionStatus.Failed) {
+        return InstructionStatus.Failed;
+      } else {
+        // TODO @prashantasdeveloper remove this once the chain handles Executed/Rejected state separately
+        return InstructionStatus.Executed;
```
prashantasdeveloper, line 210: For now, I have added a comment and clarified in the docs about the status - Executed. Once the core team makes changes to handle the status separately, I think we can live with this

VictorVicente: I'd prefer to have a status for `ExecutedOrRejected` explaining we currently can't know which is the correct one. But assuming an instruction was executed when it was rejected could lead to other issues. 
Could you also ask Robert to check when is this feature coming to the chain?

polymath-eric: @VictorVicente Do you just mean change the status name for documentation purposes?

If the core team is going to add Rejected in the future, then I think its better to leave this as just `Executed`, otherwise when we separate them it will force a breaking change or else we'll be left with a name that's verbosely misleading.

We should add a TS Doc `@note` to the status explaining it though.

polymath-eric: On second though, adding in `Rejected` should still be a major version thing, since people should look at their code again, and if they do a `never` check it could break TS projects.

I am kind of liking `ExecutedOrRejected` now.

VictorVicente (approved):

src/api/entities/Instruction/index.ts
VictorVicente: This change will suppose a renaming change for current instruction status enum. We've decided to leave as it is, and to create a new one `Rejected` once it's provided directly from the chain.

PR #877, opened 2022-10-21
refactor: 💡 Use polkadot lookup types

prashantasdeveloper:
### Description

Make use of polkadot lookup types wherever necessary

### Breaking Changes

NA

### JIRA Link

DA-430

### Checklist

### DIFF
--- a/src/api/entities/Checkpoint.ts
+++ b/src/api/entities/Checkpoint.ts
@@ -1,7 +1,8 @@
+import { u64 } from '@polkadot/types';
+import { PolymeshPrimitivesIdentityId, PolymeshPrimitivesTicker } from '@polkadot/types/lookup';
 import BigNumber from 'bignumber.js';
 
 import { Asset, Context, Entity, Identity } from '~/internal';
-import { CheckpointId, IdentityId, Ticker } from '~/polkadot/polymesh';
 import { IdentityBalance, PaginationOptions, ResultSet } from '~/types';
 import { tuple } from '~/types/utils';
 import {
@@ -128,7 +129,8 @@ export class Checkpoint extends Entity<UniqueIdentifiers, HumanReadable> {
     });
 
     const currentDidBalances: { did: string; balance: BigNumber }[] = [];
-    const balanceUpdatesMultiParams: [Ticker, IdentityId][] = [];
+    const balanceUpdatesMultiParams: [PolymeshPrimitivesTicker, PolymeshPrimitivesIdentityId][] =
+      [];
 
     // Prepare the query for balance updates. Push to currentDidBalances to be used if there are no updates for the balance
     entries.forEach(([storageKey, balance]) => {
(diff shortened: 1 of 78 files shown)


polymath-eric (commented):
LGTM

I took a look for where the types file is still being used and noticed that there is: 
[`getTransactionArguments`](https://github.com/PolymathNetwork/polymesh-sdk/blob/9651381c33448eaff54966e33af69f06822089aa/src/base/Context.ts#L651) on Context.

I know it says `public`, but `Context` is an internal construct, so we should be safe removing that method without marking it as a breaking change. I only see it referenced in the test.

src/api/entities/Identity/index.ts
```diff
@@ -524,7 +522,7 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
       return null;
     }
 
-    return scopeIdToString(scopeId);
+    return identityIdToString(scopeId);
```
VictorVicente, line 525: Why ScopeId is now being generated everywhere as an IdentityId?

src/api/procedures/addInstruction.ts
```diff
@@ -63,30 +67,30 @@ export interface Storage {
  */
 type InternalAddAndAffirmInstructionParams = [
   u64,
-  SettlementType,
-  Moment | null,
-  Moment | null,
+  PalletSettlementSettlementType,
+  u64 | null,
```
VictorVicente, line 71: Same here, replacing `Moment` with `u64` directly seems a bit risky to me

src/api/entities/Checkpoint.ts
```diff
@@ -1,7 +1,8 @@
+import { u64 } from '@polkadot/types';
```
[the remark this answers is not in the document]
polymath-eric: All `CheckPointId` was an alias for `u64` `CheckPointId extends u64 {}` so technically a sub type I suppose.

I'd prefer having an explicit ID type for it, but if it changes then the types for the extrinsics should change and give type errors.

src/api/procedures/addInstruction.ts
polymath-eric: Moment was an alias for u64 like the ID.

I agree with `Moment` being better, but with the previous behavior and TS "structural" type system I think they could be used interchangeable before.

src/api/entities/Identity/index.ts
prashantasdeveloper: Under the hood both are the same and new lookup types generated point them as IdentityId

prashantasdeveloper:
> Could you remember me what's the advantage of using @polkadot/types/lookup instead of ~/polkadot/polymesh/types?

Main problem was wherever we tried creating type (`context.createType('SomeType', def)`) and if the type referenced was not of the form present in `@polkadot/types/lookup` it used to throw an error. 

This is what we faced in [this PR](https://github.com/PolymeshAssociation/polymesh-sdk/pull/871/files). Short name from `~/polkadot/polymesh/types` was used instead of long name from `@polkadot/types/lookup`. So, in order for that to work correctly we should have used types from lookup file instead of types file. 

That's why I didn't wanted to us to be dependent on the types anymore and directly use the lookup types. One less thing to worry about.

polymath-eric:
To add onto Prashant's comment, if we did 100% dynamic lookup I believe core could add new Option fields to extrinsics and the SDK would be able to handle that without an upgrade. (Not 100% sure and would take a while to test)

With RPC the types are still needed, but its only a few special cases we use them.

VictorVicente (approved):

PR #883, opened 2022-11-07
chore: 🤖 Modify `lookup.d.ts` in @polkadot/types to fix typings

prashantasdeveloper:
### Description

The types package for polkadot (`@polkadot/types`) only declares the lookup module and expects it to be implemented by the consumer. SDK uses the `yarn polkadot-types-from-defs` to generate the types correctly. The allows the compiler to understand that at runtime packages will be available but intellisense doesn't work with lookup types. This change helps fix that and compiler gets to know about the types at compile time.

### Breaking Changes

NA

### JIRA Link

DA-427

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,8 @@
     "lint": "eslint src --ext .js,.jsx,.ts,.tsx",
     "format": "cross-env prettier-eslint $PWD\"/src/**/*.{ts,tsx,js,jsx}\" --write",
     "ts-node": "yarn build:ts && ts-node --",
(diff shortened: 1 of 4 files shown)


polymath-eric (approved):
LGTM

Two nitty thing about the PR:

- the Jira link is to the old board
- It would be nice to have "add patch-package" somewhere in the commit message. I haven't thought of this before, but it would be nice to have a chance at finding when packages were added by looking at git log.

VictorVicente (approved):

PR #896, opened 2022-11-29
fix: 🐛 Handle queryMulti params correctly

prashantasdeveloper:
### Description

Following error was currently thrown 
```
2022-11-29 16:38:36             VEC: Unable to decode on index 1 Expected an array of 2 values as params to a Map query
2022-11-29 16:38:36        RPC-CORE: subscribeStorage(keys?: Vec<StorageKey>): StorageChangeSet:: createType(Vec<StorageKey>):: Expected an array of 2 values as params to a Map query
2022-11-29 16:38:36             DRR: createType(Vec<StorageKey>):: Expected an array of 2 values as params to a Map query

~/polymesh-rest-api-cloned/node_modules/@polkadot/types-create/cjs/create/type.js:73
    firstError = new Error(`createType(${type}):: ${error.message}`);
                 ^
Error: createType(Vec<StorageKey>):: Expected an array of 2 values as params to a Map query
```

`queryMulti` expects an array of arguments with query storage allows more than one arguments. This change fixes the generic type for `requestMulti` internal function to infer the same. 

### Breaking Changes

NA

### JIRA Link

DA-452

### Checklist

### DIFF
--- a/src/utils/internal.ts
+++ b/src/utils/internal.ts
@@ -528,7 +528,11 @@ export async function getApiAtBlock(
 
 type QueryMultiParam<T extends AugmentedQuery<'promise', AnyFunction>[]> = {
   [index in keyof T]: T[index] extends AugmentedQuery<'promise', infer Fun>
-    ? [T[index], ...Parameters<Fun>]
+    ? Fun extends (firstArg: infer First, ...restArg: infer Rest) => ReturnType<Fun>
+      ? Rest extends never[]
+        ? [T[index], First]
+        : [T[index], Parameters<Fun>]
+      : never
     : never;
 };
 
--- a/src/api/procedures/__tests__/registerMetadata.ts
+++ b/src/api/procedures/__tests__/registerMetadata.ts
@@ -105,9 +105,11 @@ describe('registerMetadata procedure', () => {
       name,
       specs: {},
     };
+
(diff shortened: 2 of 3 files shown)


polymath-eric (approved):
LGTM.

Ideally we'd be able to use queryMulti instead. Do you know what the problem is? I feel like its going to cause issues in the future.

prashantasdeveloper:
> LGTM.
> 
> Ideally we'd be able to use queryMulti instead. Do you know what the problem is? I feel like its going to cause issues in the future.

I interpreted the error incorrectly. Actually the queryMulti expects us to pass the arguments of query storage as an arrary if number of arguments is greater than 1 (Updated the error in the PR description). 
Right now, the arguments were flattened in all cases. I have updated the type to handle the same

src/utils/internal.ts
```diff
@@ -528,7 +528,11 @@ export async function getApiAtBlock(
 
 type QueryMultiParam<T extends AugmentedQuery<'promise', AnyFunction>[]> = {
   [index in keyof T]: T[index] extends AugmentedQuery<'promise', infer Fun>
-    ? [T[index], ...Parameters<Fun>]
+    ? Fun extends (firstArg: infer First, ...restArg: infer Rest) => ReturnType<Fun>
```
polymath-eric, line 531: Nice work, I like the API this provides. Even if its a bit hard to read people shouldn't really have to look at it.

PR #900, opened 2022-12-14
chore: 🤖 Update middleware V2 types

prashantasdeveloper:
### Description

The enum types have changed from schema names to postgres enum names, since SQ was updated to use latest node. This PR creates a script to export postgres name enums as human readable enums.

### Breaking Changes

NA

### JIRA Link

DA-470

### Checklist


src/api/client/Network.ts
```diff
@@ -169,8 +169,8 @@ export class Network {
 
     if (context.isMiddlewareV2Enabled()) {
       return this.getEventByIndexedArgsV2({
-        moduleId: moduleId as MiddlewareV2ModuleId,
-        eventId: eventId as MiddlewareV2EventId,
+        moduleId: moduleId as unknown as MiddlewareV2ModuleId,
```
polymath-eric, line 172: Don't we need to map back to the `PublicEnum7A0B4Cc03E` form before calling GraphQL?

Why no `Record<ModuleId, MiddlewareV2ModuleId>` to lookup the value?

prashantasdeveloper: We can safely assume that SQ types will always be superset of all the types/enum values mentioned in schema file of tooling. So we don't need to manually map all the tooling values to SQ values. 

Also, for quering the GQL on SQ, we need to define types as `PublicEnum.....`. This is handled in one common func - https://github.com/PolymeshAssociation/polymesh-sdk/blob/78a667cc8b0041dbf2cd21149472448b5c6ba9d1/src/middleware/queriesV2.ts#L275

prashantasdeveloper: This above function uses const which takes the  generic enum name that we are using throughout and maps it to corresponding hashed name of enum (or postgres enum name style)

polymath-eric: TypeScript is usually pretty smart, and its saying the types can never overlap.

I am not sure how, but we shouldn't need an `unknown` cast here. This is implying TS inference is wrong, and moduleId can take on any value and still satisfy the types.

polymath-eric: If runtime works and we do need to cast, I think a typeguard assertion with even a basic `instanceof` check would be better than an `unknown` inline cast

prashantasdeveloper: The enum values are mostly the same and I too agree that typescript is smart to know that both values are same but it isn't being smart right now. I would keep this assertion for now, anyhow it's not that long before we remove the v1 code, so all these will get removed eventually.

polymath-eric (approved):
LGTM

PR #902, opened 2022-12-20
feat: 🎸 Support fetching historical instructions for a DID

prashantasdeveloper:
### Description

Fetch historical instructions using middleware v2 where a DID is involved

### Breaking Changes

NA

### JIRA Link

DA-168

### Checklist

### DIFF
--- a/src/middleware/queriesV2.ts
+++ b/src/middleware/queriesV2.ts
@@ -324,12 +324,14 @@ export function instructionsQuery(
  *
  * Get Instructions where an identity is involved
  */
-export function instructionsByDidQuery(identityId: string): GraphqlQuery {
+export function instructionsByDidQuery(
+  identityId: string
+): GraphqlQuery<QueryArgs<Leg, 'fromId' | 'toId'>> {
   const query = gql`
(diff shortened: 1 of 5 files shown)


polymath-eric:
There is a smell:

`status: status as InstructionStatusEnum,` could be just `status` I think

I tried locally removing the assertion and things still built. guessing its in the base branch here.

src/middleware/queriesV2.ts
```diff
@@ -319,6 +319,70 @@ export function instructionsQuery(
...
+            }
+          }
+        }
+      }
+    }
+  `;
+
+  return {
+    query,
+    variables: undefined,
```
polymath-eric, line 382: (nit) Should `identityId` be a GraphQL variable? 

Looks like its getting string interpolated which doesn't seem as nice as the first class variable to me.

That said, if it needs to be like that for some reason or is more consistent its OK to leave

polymath-eric (approved):
LGTM, 

Left one nitty comment

polymath-eric (approved):

VictorVicente (approved):

PR #903, opened 2022-12-21
feat: 🎸 Add Subsidies namespace

prashantasdeveloper:
### Description

Adds `Subsidies` namespace which contains two methods -
* `getBeneficiaries` - to get list of all subsidy relationship along with allowance where the Account is the subsidizer
* `getSubsidizer` - get the Subsidy relationship along with allowance where the Account is the beneficiary

### Breaking Changes

NA

### JIRA Link

DA-196

### Checklist

### DIFF
--- a/src/api/entities/Account.ts
+++ b/src/api/entities/Account.ts
@@ -321,6 +321,8 @@ export class Account extends Entity<UniqueIdentifiers, string> {
    *   this Account isn't being subsidized, return null
    *
    * @note can be subscribed to
+   *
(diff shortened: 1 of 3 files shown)


prashantasdeveloper:
Since we do have `Subsidies` namespace now, should we mark `Account.getSubsidy` as deprecated in favour of `Subsidies.getSubsidizer` ? They both basically do the same thing

VictorVicente (approved):

polymath-eric (approved):
LGTM

I think the current `account.getSubsidy` should be marked as deprecated and we should take it out eventually.

polymath-eric (approved):

PR #914, opened 2023-01-17
fix: 🐛 Use middleware V2 as well for emitting `ProcessedByMiddleware`

prashantasdeveloper:
### Description

`onProcessedByMiddleware` only looks at middleware v1 to check if block has been processed or not. Modified the function to check for both v1 and v2 and use the latest block query accordingly

### Breaking Changes

NA

### JIRA Link

DA-485

### Checklist


src/base/PolymeshTransactionBase.ts
```diff
@@ -451,6 +453,34 @@ export abstract class PolymeshTransactionBase<
     };
   }
 
+  /**
+   * Get the latest processed block from the database
+   *
+   * @note uses the middleware
+   */
+  private async getLatestBlockFromDb(): Promise<BigNumber> {
```
polymath-eric, line 461: (nit) its getting it from a GraphQL server, not a DB. "FromDb" makes it seem like there is a direct DB connection to me.

```suggestion
  private async getLatestBlockFromMiddleware(): Promise<BigNumber> {
```
or
```suggestion
  private async getLatestMiddlewareBlock(): Promise<BigNumber> {
```

src/base/PolymeshTransactionBase.ts
```diff
@@ -451,6 +453,34 @@ export abstract class PolymeshTransactionBase<
...
+    let processedBlock: number;
+    if (context.isMiddlewareV2Enabled()) {
+      ({
+        data: {
+          blocks: {
+            nodes: [{ blockId: processedBlock }],
+          },
+        },
+      } = await context.queryMiddlewareV2<EnsuredV2<QueryV2, 'blocks'>>(latestBlockQuery()));
+    } else {
```
polymath-eric, line 473: This is private, so maybe this is over defensive, but should we check?

On on hand we already have check logic so it might be redundant. Thoughts?
```suggestion
    } else if (context.isMiddlewareEnabled()) {
    ...
    } else {
      throw new Error(...
    }
    
```

src/base/PolymeshTransactionBase.ts
```diff
@@ -437,7 +439,7 @@ export abstract class PolymeshTransactionBase<
   public onProcessedByMiddleware(listener: (err?: PolymeshError) => void): UnsubCallback {
     const { context, emitter } = this;
 
-    if (!context.isMiddlewareEnabled()) {
+    if (!context.isMiddlewareV2Enabled() && !context.isMiddlewareEnabled()) {
```
polymath-eric, line 442: I think this would be easier to read as a single Context method

polymath-eric (commented):
LGTM, a few small optional comments

src/base/PolymeshTransactionBase.ts
prashantasdeveloper: > On on hand we already have check logic so it might be redundant. Thoughts?

I too think it would be redundant. Also, we would be pretty soon be getting rid of all V1 methods when we will remove all related code.

polymath-eric (approved):

PR #916, opened 2023-01-18
fix: 🐛 ensure subscriptions are closed before disconnecting

polymath-eric:
### Checklist

### DIFF
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -128,6 +128,13 @@ export class Context {
   private nonce?: BigNumber;
 
   /**
+   * A counter for keeping track of internal subscriptions. Disconnect will wait until its zero
+   *
+   * @note it must be an "Object" so cloned Context will share the reference
+   */
+  private subscriptions = { chain: 0 };
+
+  /**
    * @hidden
    */
   private constructor(params: ConstructorParams) {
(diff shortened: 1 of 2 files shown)


polymath-eric:
The subscription was being [made](https://github.com/PolymathNetwork/polymesh-sdk/blob/67f7799c9546669425380363f6ed6e0c2c831d49/src/base/Context.ts#L1346), Which is [called here](https://github.com/PolymathNetwork/polymesh-sdk/blob/a456c72e6de49095300e98513521f9eda71889e8/src/base/PolymeshTransactionBase.ts#L469). The simplest fix is to remove middlewareV1 from the SDK.
I was wrong in the time needed to wait. Its only the time for a call to the chain to return (~10ms locally, ~100ms against testnet for me).
The reason for the issue is an optimization using the subscription API. On RPC there is only a way to get the finalized block hash, and then use that in another call to get the finalized block number.
MiddlewareV1 needs the latestNumber, so either we deoptimize the call, put a slight delay in disconnect, or [use this counter](https://github.com/PolymeshAssociation/polymesh-sdk/pull/916).
If we remove V1, then we can use the RPC + add a new query for blockByHash we wouldn’t have the pending sub issue.

polymath-eric:
Closing in favor of a simple sleep: https://github.com/PolymeshAssociation/polymesh-sdk/pull/918

That was a lot of reading for a simple change, but I am happy to not track state in a counter and worry about exactly once semantics.

PR #917, opened 2023-01-18
docs: ✏️ Add backtick around <T> and export missing v2 types

prashantasdeveloper:
### Description

Adds `\` for proper rendering of <T> in docs + exports missing v2 types needed in documentation

### Breaking Changes

NA

### JIRA Link

NA

### Checklist

### DIFF
--- a/src/types/utils/index.ts
+++ b/src/types/utils/index.ts
@@ -24,7 +24,7 @@ export type UnionOfProcedureFuncs<Args, ReturnValue, Storage> = Args extends unk
   : never;
 
 /**
- * Less strict version of Parameters<T\>
+ * Less strict version of `Parameters<T>`
  */
 export type ArgsType<T> = T extends (...args: infer A) => unknown ? A : never;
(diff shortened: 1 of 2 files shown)


src/types/utils/index.ts
```diff
@@ -24,7 +24,7 @@ export type UnionOfProcedureFuncs<Args, ReturnValue, Storage> = Args extends unk
   : never;
 
 /**
- * Less strict version of Parameters<T>
+ * Less strict version of Parameters<T\>
```
polymath-eric, line 27:
```suggestion
 * Less strict version of `Parameters<T>`
```

I thought we wanted to use `

prashantasdeveloper: Ohh when I checked and generated docs, it altogether skipped <T> and only had `Less strict version of Parameters` so I thought all we wanted was to display it correctly

polymath-eric (approved):

PR #918, opened 2023-01-18
fix: 🐛 allow 500ms for chain connections to drain

polymath-eric:
### Description

some subscriptions to the chain may reject on disconnect if WS is closed immediately after submitting a transaction. 500ms is conservative, generally a little over the ping time to the node should suffice.

polkadot.js will [reject any pending subscription](https://github.com/polkadot-js/api/pull/3197/files#diff-871752a27c655476527379eacbc4f66b71eb08a5bb1639acbcf8e380c57cda6cR355) when disconnect is called.

While reasonable and efficient, our procedure middleware emitter runs async and can prevent a clean shutdown if disconnect is called too quickly.

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-483](https://polymesh.atlassian.net/browse/DA-483)

### Checklist

### DIFF
--- a/src/base/Context.ts
+++ b/src/base/Context.ts
@@ -83,7 +83,13 @@ import {
   u16ToBigNumber,
   u32ToBigNumber,
 } from '~/utils/conversion';
-import { assertAddressValid, calculateNextKey, createClaim, getApiAtBlock } from '~/utils/internal';
+import {
+  assertAddressValid,
+  calculateNextKey,
+  createClaim,
+  delay,
+  getApiAtBlock,
+} from '~/utils/internal';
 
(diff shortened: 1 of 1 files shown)


prashantasdeveloper (approved):

VictorVicente (approved):
Is this marked as a breaking change? It should.

polymath-eric:
@VictorVicente why is it a breaking change? disconnect still returns a Promise, so the API didn't change, even though there is now an `async` annotation.

I don't think this is breaking. We never guaranteed disconnect would take a certain time, and the user still gets the same type returned.

VictorVicente (approved):

PR #923, opened 2023-01-20
fix: 🐛 allow non default signer to quit portfolios

polymath-eric:
### Description

fix bug where the procedure context differs from the portfolio context

The issue happens when using the non default signer to leave the default signer's portfolio since the `context` variable is different between the objects.

### Breaking Changes

n/a

### JIRA Link

n/a

### Checklist

### DIFF
--- a/src/api/procedures/quitCustody.ts
+++ b/src/api/procedures/quitCustody.ts
@@ -33,9 +33,9 @@ export async function prepareQuitCustody(
   const { portfolio } = args;
 
   const signer = await context.getSigningIdentity();
-  const isOwnedBy = await portfolio.isOwnedBy({ identity: signer });
+  const isOwnedBySigner = await portfolio.isOwnedBy({ identity: signer });
 
-  if (isOwnedBy) {
+  if (isOwnedBySigner) {
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
       message: 'The Portfolio owner cannot quit custody',
--- a/src/api/procedures/quitCustody.ts
+++ b/src/api/procedures/quitCustody.ts
@@ -32,7 +32,8 @@ export async function prepareQuitCustody(
 
   const { portfolio } = args;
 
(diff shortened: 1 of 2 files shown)


src/api/procedures/quitCustody.ts
```diff
@@ -32,7 +32,8 @@ export async function prepareQuitCustody(
 
   const { portfolio } = args;
 
-  const isOwnedBy = await portfolio.isOwnedBy();
+  const signer = await context.getSigningIdentity();
```
polymath-eric, line 35: I think a unit test is a bit hard for this. I found it with an integration one, so we'll have that coverage though.

The other solution is to destructure the given Portfolio and create a new one with the current context.

Its very edge case since you need to be a custodian that somehow have a key ring with your customer's private key as the default in your signing manager

polymath-eric:
```ts
  const createPortfolioTx = await sdk.identities.createPortfolio({ name: 'CUSTODY_PORTFOLIO' });
  const portfolio = await createPortfolioTx.run();

  const setCustodianTx = await portfolio.setCustodian({ targetIdentity: custodian });
  const authRequest = await setCustodianTx.run();
  const acceptTx = await authRequest.accept({ signingAccount: custodianAccount });
  await acceptTx.run();

  const custodiedPortfolios = await custodian.portfolios.getCustodiedPortfolios();

  // Quit being a custodian
  const [portfolioToQuit] = custodiedPortfolios.data;

  const quitCustodyTx = await portfolioToQuit.quitCustody({ signingAccount: custodianAccount }); // throws owner can't leave own portfolio
  await quitCustodyTx.run();
  ```

src/api/procedures/quitCustody.ts
prashantasdeveloper: Can we atleast assert that isOwnedBy was called with {identity: signer} param?

prashantasdeveloper (approved):
LGTM, left a small suggestion for updating test case

PR #931, opened 2023-01-30
feat: 🎸 add register did with cdd in `Identities` namespace

polymath-eric:
### Description

allow a DID to be created with a CDD claim using the new chain
extrinsic. Previously these were unbatchable since the CDD claim needed the DID as an argument

### Breaking Changes

none

### JIRA Link

✅ Closes: DA-498

### Checklist


polymath-eric:
I looked at reusing the existing `registerIdentity` procedure, which would certainly be a "DRYer
" way, but it would also require weird params being passed in when its being set up.

Its a bit more verbose, but easier to follow along, and less likely to introduce bugs. If there was a third method for DID creation, then the insides should probably be extracted to a shared function.

src/api/client/Identities.ts
```diff
@@ -65,6 +74,20 @@ export class Identities {
...
+   * Functions like {@link api/client/Identities!registerIdentity | registerIdentity} followed by a CDD claim being added, except in a single transaction.
+   *
+   * @note this may create {@link api/entities/AuthorizationRequest!AuthorizationRequest | Authorization Requests} which have to be accepted by the `targetAccount`.
+   *   An {@link api/entities/Account!Account} or {@link api/entities/Identity!Identity} can fetch its pending Authorization Requests by calling {@link api/entities/common/namespaces/Authorizations!Authorizations.getReceived | authorizations.getReceived}.
+   *   Also, an Account or Identity can directly fetch the details of an Authorization Request by calling {@link api/entities/common/namespaces/Authorizations!Authorizations.getOne | authorizations.getOne}
+   *
+   * @note required role:
+   *   - Customer Due Diligence Provider
+   */
+  public registerIdentityWithCdd: ProcedureMethod<RegisterIdentityWithCddParams, Identity>;
```
VictorVicente, line 89: Any reason for adding a new method instead of overloading the existing one? No preference, but I'd like to know your thoughts

polymath-eric:
@VictorVicente updated with one method. Quite a bit cleaner this way actually. Thanks for the suggestion.

VictorVicente (approved):

PR #932, opened 2023-01-31
feat: 🎸 Add procedure to manually execute instruction

prashantasdeveloper:
### Description

Adds procedure to manually execute an instruction

### Breaking Changes

NA

### JIRA Link

DA-503

### Checklist

### DIFF
--- a/src/api/procedures/executeManualInstruction.ts
+++ b/src/api/procedures/executeManualInstruction.ts
@@ -60,7 +60,6 @@ export async function prepareExecuteManualInstruction(
       owner: { did: ownerDid },
     } = await instructionDetails.venue.details();
 
-    console.log(ownerDid, signer);
     if (ownerDid !== signer) {
       throw new PolymeshError({
         code: ErrorCode.UnmetPrerequisite,
--- a/src/api/procedures/executeManualInstruction.ts
+++ b/src/api/procedures/executeManualInstruction.ts
@@ -26,7 +26,7 @@ export interface Storage {
   portfolios: (DefaultPortfolio | NumberedPortfolio)[];
   totalLegAmount: BigNumber;
   instructionDetails: InstructionDetails;
-  signer: string;
+  signerDid: string;
 }
 
 /**
@@ -46,7 +46,7 @@ export async function prepareExecuteManualInstruction(
       },
     },
     context,
-    storage: { portfolios, totalLegAmount, instructionDetails, signer },
+    storage: { portfolios, totalLegAmount, instructionDetails, signerDid },
   } = this;
 
   const { id } = args;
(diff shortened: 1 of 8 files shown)


src/api/procedures/executeManualInstruction.ts
```diff
@@ -0,0 +1,174 @@
...
+  const instruction = new Instruction({ id }, context);
+
+  await assertInstructionValidForManualExecution(instructionDetails, context);
+
+  if (!portfolios.length) {
+    const {
+      owner: { did: ownerDid },
+    } = await instructionDetails.venue.details();
+
+    console.log(ownerDid, signer);
```
polymath-eric, line 63:
```suggestion
```

src/api/procedures/executeManualInstruction.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (ownerDid !== signer) {
+      throw new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'The signing Identity is not involved in this Instruction',
```
polymath-eric, line 67: Isn't this about not owning the Venue?

prashantasdeveloper: It's like if you are not part of the legs but still own the venue you can execute that instruction. Felt correct this way to me

src/api/procedures/executeManualInstruction.ts
```diff
@@ -0,0 +1,174 @@
...
+  portfolioIdToMeshPortfolioId,
+  portfolioLikeToPortfolioId,
+  u64ToBigNumber,
+} from '~/utils/conversion';
+
+export interface Storage {
+  portfolios: (DefaultPortfolio | NumberedPortfolio)[];
+  totalLegAmount: BigNumber;
+  instructionDetails: InstructionDetails;
+  signer: string;
```
polymath-eric, line 29:
```suggestion
  signerDid: string;
```

Usually a "signer" is more associated to an address or key. I think Did makes it clearer what it is

src/api/procedures/executeManualInstruction.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  const { id } = args;
+
+  const instruction = new Instruction({ id }, context);
+
+  await assertInstructionValidForManualExecution(instructionDetails, context);
+
+  if (!portfolios.length) {
+    const {
+      owner: { did: ownerDid },
```
polymath-eric, line 60:
```suggestion
      owner: { did: venueOwner },
```
Here I think "owner" implies Identity enough we don't need the "did" part.

src/api/procedures/executeManualInstruction.ts
polymath-eric: I left two variable name suggestions. I think with those it will be a bit more clear here.

I think keeping it general as it is will be OK.

polymath-eric, line 66:
```suggestion
        message: 'The signing identity is not involved in this Instruction',
```
(very nit) To me this looks better. I am thinking capital letters should be for `new`able things (the "domain objects"), and since "signing identity" is an instance it doesn't get special capitalization rules.

polymath-eric (approved):

VictorVicente (approved):

PR #933, opened 2023-01-31
5.2.0 validations

polymath-eric:
### Description

Add additional validations to reflect new chain rules introduced in 5.2.0

### Breaking Changes

Although not breaking API changes, these deserve a mention:
  - Portfolio movements can only move the same Asset once
  - New tickers must consist of alpha numeric values only now

### JIRA Link

[DA-494](https://polymesh.atlassian.net/browse/DA-494)

For the point:

- Add ensure check for revoke_claim to make sure claim exist

I tried and we will return the error `PolymeshError: Attempt to edit claims that weren't issued by the signing Identity`

This is an accurate message, although in theory, it could be more precise for this specific case.

### Checklist

[DA-494]: https://polymesh.atlassian.net/browse/DA-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/procedures/moveFunds.ts
+++ b/src/api/procedures/moveFunds.ts
@@ -1,9 +1,9 @@
 import BigNumber from 'bignumber.js';
+import { uniq } from 'lodash';
 
 import { assertPortfolioExists } from '~/api/procedures/utils';
 import { DefaultPortfolio, NumberedPortfolio, PolymeshError, Procedure } from '~/internal';
 import {
-  Asset,
   ErrorCode,
   MoveFundsParams,
   PortfolioId,
@@ -17,7 +17,7 @@ import {
   portfolioLikeToPortfolioId,
   portfolioMovementToMovePortfolioItem,
 } from '~/utils/conversion';
-import { asTicker, hasDuplicates } from '~/utils/internal';
+import { asTicker } from '~/utils/internal';
 
 /**
  * @hidden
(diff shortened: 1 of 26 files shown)


src/utils/internal.ts
```diff
@@ -1391,6 +1401,22 @@ export function assertTickerValid(ticker: string): void {
   }
 }
 
+/**
+ * @hidden
+ *
+ * Validates a ticker value is valid to be created. Some tickers were created with more relaxed rules
+ */
+export function assertTickerCreatable(ticker: string): void {
```
VictorVicente, line 1409: Not a fan of this name. `assertTickerIsValid`?

polymath-eric: `assertTickerValid ` is already defined, and we need to leave that function as is to support grandfathered tickers.

polymath-eric: > The new restriction is only on new ticker registrations/create_asset calls.  All other usages of Ticker still have the old validation.

I think `create` captures the intended use case of where this validation should be used.

polymath-eric: Looking at it, this check can be inlined and I can get rid of this function

src/api/procedures/moveFunds.ts
```diff
@@ -82,6 +83,17 @@ export async function prepareMoveFunds(
     });
   }
 
+  const tickers = items.map(({ asset }) => (typeof asset === 'string' ? asset : asset.ticker));
```
prashantasdeveloper, line 86:
```suggestion
  const tickers = items.map(asTicker);
```

prashantasdeveloper (changes requested):
Left a small nit comment

prashantasdeveloper (approved):

VictorVicente (approved):

PR #949, opened 2023-03-01
fix: 🐛 Add new GQL codegen types for middleware V2

VictorVicente:
### Description

Add new GQL codegen types for middleware V2

### Breaking Changes

NA

### JIRA Link

DA-540

### Checklist


prashantasdeveloper (approved):
I checked in my local machine since github wasn't showing the diff. It seems we have new `distinct` param added for querying. Moving ahead this could help in some claims related queries I guess but I'm not sure. Something to keep an eye on. LGTM for now

PR #952, opened 2023-03-09
fix: 🐛 Update compatible version to 5.2.x

VictorVicente:
### Description

Update compatible version to 5.2.x

### Breaking Changes

NA

### JIRA Link

NA

### Checklist

### DIFF
--- a/src/utils/internal.ts
+++ b/src/utils/internal.ts
@@ -1318,8 +1318,6 @@ function handleSpecVersionResponse(
     return false;
   }
 
-  console.log(specVersionAsSemver);
-  console.log(SUPPORTED_SPEC_VERSION_RANGE);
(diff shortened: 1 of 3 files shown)


polymath-eric (approved):

src/utils/internal.ts
```diff
@@ -1318,7 +1318,7 @@ function handleSpecVersionResponse(
     return false;
   }
 
-  if (!satisfies(specVersionAsSemver, SUPPORTED_SPEC_VERSION_RANGE)) {
+  if (gtr(specVersionAsSemver, SUPPORTED_SPEC_VERSION_RANGE)) {
```
prashantasdeveloper, line 1321: By changing it to `gtr` this will only log when specVersion from chain is greater than the specified range

prashantasdeveloper (approved):

PR #953, opened 2023-03-12
feat: 🎸 Add support for spec range 5.3.x

prashantasdeveloper:
### Description

Updates supported spec version and node version to 5.3.x 

### Breaking Changes

NA

### JIRA Link

DA-571

### Checklist

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -1,4 +1,5 @@
 import BigNumber from 'bignumber.js';
+import { coerce } from 'semver';
 
 import { TransactionArgumentType } from '~/types';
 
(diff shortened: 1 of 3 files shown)


src/utils/internal.ts
```diff
@@ -1316,7 +1318,7 @@ function handleSpecVersionResponse(
     return false;
   }
 
-  if (!satisfies(specVersionAsSemver, SUPPORTED_SPEC_VERSION_RANGE)) {
+  if (gtr(specVersionAsSemver, SUPPORTED_SPEC_VERSION_RANGE)) {
```
polymath-eric, line 1321: I think there should another be a "less than" check for the minimum version that will work.

src/utils/constants.ts
```diff
@@ -109,12 +110,18 @@ export const ROOT_TYPES = rootTypes;
 /**
  * The Polymesh RPC node version range that is compatible with this version of the SDK
  */
-export const SUPPORTED_NODE_VERSION_RANGE = '5.2.0';
+export const SUPPORTED_NODE_VERSION_RANGE = '5.3.0';
```
polymath-eric, line 113: I think we should have two variables, `MAX_SUPPORTED_NODE_VERSION` and `MIN_SUPPORTED_NODE_VERSION`.

For the first 5.3 release we leave MIN at 5.2 and set MAX to 5.3. Then when we add NFT features we bump the MIN to 5.3.

polymath-eric (commented):
I think we should do a minimum check as well + add a min variable.

VictorVicente (approved):

PR #957, opened 2023-03-21
feat: 🎸 add validate account method

polymath-eric:
### Description

add sync method to validate address strings. `getAccount` implicitly validates the address, however is async making it less versatile

When I added multisig, I made `getAccount` async, which adds extra checks. This lets people check just the address is OK

### Checklist

### DIFF
--- a/src/api/client/AccountManagement.ts
+++ b/src/api/client/AccountManagement.ts
@@ -27,7 +27,7 @@ import {
   UnsubCallback,
 } from '~/types';
 import { stringToAccountId } from '~/utils/conversion';
-import { asAccount, createProcedureMethod } from '~/utils/internal';
+import { asAccount, assertAddressValid, createProcedureMethod } from '~/utils/internal';
 
 /**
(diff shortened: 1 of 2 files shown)


src/api/client/AccountManagement.ts
```diff
@@ -276,4 +276,17 @@ export class AccountManagement {
 
     return new Subsidy({ beneficiary: beneficiaryAddress, subsidizer: subsidizerAddress }, context);
   }
+
+  /**
+   * Returns validation error if @param args.address is not a valid ss58 string for the connected network
+   */
+  public validateAddress(args: { address: string }): Error | null {
```
VictorVicente, line 283: Should it be an assertions or should it return a boolean? I have no clear answer.

prashantasdeveloper: Something like isValid return true or false would be better IMO

VictorVicente (approved):

PR #961, opened 2023-03-31
chore: 🤖 update build to use node v18

polymath-eric:
### Description

update tooling to use node v18 since node v14 is approaching end of life. Also add some package.json fields

there shouldn't be any user impact because we only reference node version for our build tooling

I saw there was an "engines" field for package.json, but it seems to cause some issues with the browser. We know the SDK works with v14 at the moment, so I don't think its worth risking issues with front end repos.

This also lead me to discovering some other fields I think we should have though.

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-546](https://polymesh.atlassian.net/browse/DA-546)

### Checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: '14.x'
+          node-version: '18.x'
       - name: install dependencies
         run: yarn --frozen-lockfile
       - name: generate docs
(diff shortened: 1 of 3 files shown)


polymath-eric:
This makes it seem like we shouldn't use engines:

https://stackoverflow.com/questions/73758534/is-npm-package-engines-field-intended-for-consumers-of-a-package-or-developers

VictorVicente (approved):

prashantasdeveloper (approved):
LGTM. Just one question on this, should we update @types/node to version >=18?

PR #963, opened 2023-04-05
fix: 🐛 multisig.getIdentity to return associated Identity

polymath-eric:
### Description

previous behavior was a `MultiSig.getIdentity` would return its creator, which is what `getCreator()` is for

### Breaking Changes

User's may need to switch to the correct call if the coupled to the incorrect implementation (i.e. they expected the creator identity and relied on that for further work)

### JIRA Link

[DA-623](https://polymesh.atlassian.net/browse/DA-623)

### Checklist

[DA-623]: https://polymesh.atlassian.net/browse/DA-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/conversion.ts
```diff
@@ -432,13 +432,6 @@ export function stringToEcdsaSignature(
   return context.createType('PolymeshPrimitivesEthereumEcdsaSignature', signature);
 }
 
-/**
- * @hidden
- */
-export function accountIdToAccount(accountId: AccountId, context: Context): Account {
```
polymath-eric, line 438: This was an unused method, only referenced in tests

prashantasdeveloper (approved):

VictorVicente (approved):

PR #982, opened 2023-05-11
feat: 🎸 add procedure for setVenuFiltering

sansan:
### Description

adds procedure for enabling/disabling venue filtering

### JIRA Link

https://polymesh.atlassian.net/browse/DA-190

### Checklist


src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,62 @@
+import { Asset, PolymeshError, Procedure } from '~/internal';
+import { ErrorCode, SetVenueFilteringParams, TxTags } from '~/types';
+import { ExtrinsicParams, ProcedureAuthorization, TransactionSpec } from '~/types/internal';
+
+/**
```
prashantasdeveloper, line 5: You don't need to set this type

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,62 @@
...
+    context: {
+      polymeshApi: { tx, query },
+    },
+  } = this;
+
+  const { ticker, enabled: setEnabled } = args;
+
+  const isEnabled = await query.settlement.venueFiltering(ticker);
+
+  console.log({ isEnabled, setEnabled });
```
prashantasdeveloper, line 27:
```suggestion

```

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (isEnabled.valueOf() === setEnabled) {
+    throw new PolymeshError({
+      code: ErrorCode.NoDataChange,
+      message: `Venue filtering is already ${setEnabled ? 'enabled' : 'disabled'}`,
+    });
+  }
+
+  return {
+    transaction: tx.settlement.setVenueFiltering,
+    args: [ticker, setEnabled],
```
prashantasdeveloper, line 38: Instead of directly passing `ticker` and `setEnabled` you need to first convert them to specific chain types using some existing conversion functions. `stringToTicker` for ticker and `booleanToBool` for setEnabled

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,62 @@
...
+  this: Procedure<Params, void>,
+  args: Params
+): Promise<TransactionSpec<void, ExtrinsicParams<'settlement', 'setVenueFiltering'>>> {
+  const {
+    context: {
+      polymeshApi: { tx, query },
+    },
+  } = this;
+
+  const { ticker, enabled: setEnabled } = args;
```
prashantasdeveloper, line 23:
```suggestion
  const { ticker, enabled } = args;
```
IMO there is no need for introducing a new variable here

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,62 @@
...
+export function getAuthorization(
+  this: Procedure<Params, void>,
+  { ticker }: Params
+): ProcedureAuthorization {
+  return {
+    permissions: {
+      transactions: [TxTags.settlement.SetVenueFiltering],
+      assets: [new Asset({ ticker }, this.context)],
+    },
+  };
```
prashantasdeveloper, line 55:
```suggestion
  const { context } = this;
  return {
    permissions: {
      transactions: [TxTags.settlement.SetVenueFiltering],
      assets: [new Asset({ ticker }, context)],
    },
  };
```

src/api/procedures/__tests__/setVenueFiltering.ts
```diff
@@ -0,0 +1,139 @@
...
+      });
+
+      args.enabled = false;
+
+      expect(boundFunc(args)).toEqual({
+        permissions: {
+          transactions: [TxTags.settlement.SetVenueFiltering],
+          assets: [expect.objectContaining({ ticker: enabledTicker })],
+        },
+      });
```
prashantasdeveloper, line 136: Since the `getAuthorization` method doesn't depend on `enabled` param, you can skip this.

prashantasdeveloper (changes requested):
Left some comments. Also, this should be not based out of `master`. IMO `alpha` but @VictorVicente should confirm

src/api/procedures/types.ts
```diff
@@ -974,3 +974,8 @@ export type RegisterMetadataParams =
       value: string;
       details?: MetadataValueDetails;
     };
+
+export type SetVenueFilteringParams = {
+  ticker: string;
+  enabled: boolean;
+};
```
prashantasdeveloper, line 981: This is not required here and can be directly handled in the procedure file

prashantasdeveloper (changes requested):
I just realised looking at the other PR that the `setVenueFiltering` is not invoked from anywhere. You need to add couple of methods(enableVenueFiltering, disableVenueFiltering) in `Asset/index.ts` file that will invoke this procedure

prashantasdeveloper (approved):
LGTM. For uniformity I suggested having two separate methods, but one's also OK for me. @VictorVicente @polymath-eric do you think we should have enableVenueFiltering and disableVenueFiltering as two endpoints ?

PR #984, opened 2023-05-16
feat: 🎸 allow and disallow venues for an asset

sansan:
### Description

adds method to allow and dissalow venues for an asset

### Breaking Changes

- none

### JIRA Link

https://polymesh.atlassian.net/browse/DA-192

### Checklist


polymath-eric (approved):

src/api/procedures/types.ts
```diff
@@ -974,3 +974,9 @@ export type RegisterMetadataParams =
       value: string;
       details?: MetadataValueDetails;
     };
+
+export type ManageAllowedVenuesParams = {
+  action: 'allow' | 'disallow';
+  ticker: string;
+  venues: number[];
+};
```
prashantasdeveloper, line 982: Since this type will be exported, and assuming the procedure to manage allowedVenues will sit under the Asset class, the `ManageAllowedVenueParams` should only have a list of venues like -
```
export type ManageAllowedVenueParams {
    venues: BigNumber[];
}
```
Notice that I made the type as `BigNumber` instead of number, that is because we have made sure in all of SDK to change number to BigNumber for uniformity

src/api/procedures/manageAllowedVenues.ts
```diff
@@ -0,0 +1,53 @@
+import { Asset, Procedure } from '~/internal';
+import { ManageAllowedVenuesParams, TxTags } from '~/types';
+import { ExtrinsicParams, ProcedureAuthorization, TransactionSpec } from '~/types/internal';
+
+/**
+ * @hidden
+ */
+export type Params = ManageAllowedVenuesParams;
```
prashantasdeveloper, line 8:
```suggestion
export type Params = ManageAllowedVenuesParams & {
    action: 'allow' | 'disallow';
    ticker: string;
};
```
Once you make the change for `ManageAllowedVenueParams` as mentioned in the other comment, you can change the above as suggested

src/api/procedures/manageAllowedVenues.ts
```diff
@@ -0,0 +1,53 @@
...
+    context: {
+      polymeshApi: { tx },
+    },
+  } = this;
+
+  const { ticker, venues, action } = args;
+
+  return {
+    transaction: action === 'allow' ? tx.settlement.allowVenues : tx.settlement.disallowVenues,
+    args: [ticker, venues],
```
prashantasdeveloper, line 27: We do not directly add in the ticker as params, we first need to serialize it in a way that chain won't complain. SDK can accept plain string as ticker values, but if it is < 12 characters the chain will complain.

Use `tickerToString(ticker)` to get the rawTicker value that should be sent.

prashantasdeveloper: Similarly, convert the venues using `bigNumberToU64` and then send in the values

src/api/procedures/manageAllowedVenues.ts
```diff
@@ -0,0 +1,53 @@
...
+): ProcedureAuthorization {
+  return {
+    permissions: {
+      transactions: [
+        action === 'allow' ? TxTags.settlement.AllowVenues : TxTags.settlement.DisallowVenues,
+      ],
+      assets: [new Asset({ ticker }, this.context)],
+    },
+  };
+}
```
prashantasdeveloper, line 47:
```suggestion
export function getAuthorization(
  { context }: Procedure<Params, void>,
  { ticker, action }: Params
): ProcedureAuthorization {
  return {
    permissions: {
      transactions: [
        action === 'allow' ? TxTags.settlement.AllowVenues : TxTags.settlement.DisallowVenues,
      ],
      assets: [new Asset({ ticker }, context)],
    },
  };
}
```

src/api/entities/Asset/index.ts
```diff
@@ -659,4 +680,8 @@ export class Asset extends Entity<UniqueIdentifiers, string> {
   public toHuman(): string {
     return this.ticker;
   }
+
+  public allowVenues: ProcedureMethod<ManageAllowedVenuesParams, void>;
+
+  public disallowVenues: ProcedureMethod<ManageAllowedVenuesParams, void>;
```
prashantasdeveloper, line 686: You forgot to add comments here

src/api/procedures/manageAllowedVenues.ts
```diff
@@ -0,0 +1,64 @@
...
+): Promise<TransactionSpec<void, ExtrinsicParams<'settlement', 'allowVenues'>>> {
+  const { context } = this;
+  const {
+    polymeshApi: { tx },
+  } = context;
+
+  const { ticker, venues, action } = args;
+
+  const rawTicker = stringToTicker(ticker, context);
+  const rawVenues = venues.map(v => bigNumberToU64(v, context));
```
prashantasdeveloper, line 29:
```suggestion
  const rawVenues = venues.map(venue => bigNumberToU64(venue, context));
```

src/api/procedures/manageAllowedVenues.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @hidden
+ */
+export async function prepareManageAllowedVenues(
+  this: Procedure<Params, void>,
+  args: Params
+): Promise<TransactionSpec<void, ExtrinsicParams<'settlement', 'allowVenues'>>> {
+  const { context } = this;
+  const {
+    polymeshApi: { tx },
+  } = context;
```
prashantasdeveloper, line 24:
```suggestion
  const {
    context: {
      polymeshApi: { tx },
    },
    context,
  } = this;
```

prashantasdeveloper (changes requested):

PR #988, opened 2023-05-18
feat: 🎸 allow unsetting of SDK signing manager

polymath-eric:
### Description

pass `null` to `setSigningManager` to unset the current signing manager

### Breaking Changes

None

### JIRA Link

✅ Closes: DA-765

### Checklist


src/base/Context.ts
```diff
@@ -205,7 +205,15 @@ export class Context {
    * @note the signing Account will be set to the Signing Manager's first Account. If the Signing Manager has
    *   no Accounts yet, the signing Account will be left empty
    */
-  public async setSigningManager(signingManager: SigningManager): Promise<void> {
+  public async setSigningManager(signingManager: SigningManager | null): Promise<void> {
+    if (signingManager === null) {
+      this._signingManager = undefined;
+      this.signingAddress = undefined;
+      // note polkadot API does not currently support unsetting a signer
```
polymath-eric, line 212: I looked at the upstream code, and we could safely do `this_.polymeshApi.setSigner(undefined as any)`, however we would be giving up type safety with that.

We could submit a PR upstream allowing this, but we'd need to provide reasoning and need to wait for its release + updating our code to use latest.

I think this is OK, although exposed publicly as an escape hatch the polkadotAPI is meant to be mainly an internal API.

src/base/Context.ts
```diff
@@ -205,7 +205,17 @@ export class Context {
...
    *   no Accounts yet, the signing Account will be left empty
    */
-  public async setSigningManager(signingManager: SigningManager): Promise<void> {
+  public async setSigningManager(signingManager: SigningManager | null): Promise<void> {
+    if (signingManager === null) {
+      this._signingManager = undefined;
+      this.signingAddress = undefined;
+      // TODO remove cast when polkadot/api >= v10.7.2
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this._polymeshApi.setSigner(undefined as any);
```
polymath-eric, line 214: https://github.com/polkadot-js/api/releases/tag/v10.7.2

Upstream shouldn't diverge, this cast should be OK, but ideally we get around to upgrading polkadot api sooner than later.

PR #991, opened 2023-05-24
feat: 🎸 add procedure for setVenuFiltering and allowing/disallowing venues for Asset

sansan:
### Description

creates a single method in sdk to setVenueFiltering and allo and disallow venues for an asset

as discussed in: https://polymesh.slack.com/archives/C044Z31TE56/p1684762708251399?thread_ts=1684760999.220589&cid=C044Z31TE56

### Breaking Changes

none

### JIRA Link

- https://polymesh.atlassian.net/browse/DA-190
- https://polymesh.atlassian.net/browse/DA-192

### Checklist

### DIFF
--- a/src/api/procedures/setVenueFiltering.ts
+++ b/src/api/procedures/setVenueFiltering.ts
@@ -18,15 +18,18 @@ export async function prepareVenueFiltering(
   this: Procedure<Params, void>,
   args: Params
 ): Promise<BatchTransactionSpec<void, unknown[][]>> {
-  const { context } = this;
   const {
-    polymeshApi: { tx, query },
-  } = context;
+    context: {
+      polymeshApi: { tx, query },
+    },
+    context,
+  } = this;
+
   const { ticker, enabled, allowedVenues, disallowedVenues } = args;
   const rawTicker = stringToTicker(ticker, context);
   const transactions = [];
 
-  const isEnabled = await query.settlement.venueFiltering(ticker);
+  const isEnabled = await query.settlement.venueFiltering(rawTicker);
 
   if (enabled !== undefined && isEnabled.valueOf() !== enabled) {
     transactions.push(
@@ -41,7 +44,7 @@ export async function prepareVenueFiltering(
     transactions.push(
       checkTxType({
         transaction: tx.settlement.allowVenues,
-        args: [rawTicker, allowedVenues.map(v => bigNumberToU64(v, context))],
+        args: [rawTicker, allowedVenues.map(venue => bigNumberToU64(venue, context))],
       })
     );
   }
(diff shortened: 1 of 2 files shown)


src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,96 @@
...
+): Promise<BatchTransactionSpec<void, unknown[][]>> {
+  const { context } = this;
+  const {
+    polymeshApi: { tx, query },
+  } = context;
+  const { ticker, enabled, allowedVenues, disallowedVenues } = args;
+  const rawTicker = stringToTicker(ticker, context);
+  const transactions = [];
+
+  const isEnabled = await query.settlement.venueFiltering(ticker);
```
prashantasdeveloper, line 29:
```suggestion
  const isEnabled = await query.settlement.venueFiltering(ticker);
```

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  if (enabled !== undefined && isEnabled.valueOf() !== enabled) {
+    transactions.push(
+      checkTxType({
+        transaction: tx.settlement.setVenueFiltering,
+        args: [rawTicker, booleanToBool(enabled, context)],
+      })
+    );
+  }
```
prashantasdeveloper, line 38: In a scenario, where I only pass in the allowed venues my expectation is that even if the ticker has not enabled the venue filtering it will automatically enable it. IMO, in case when `enabled` is undefined and `allowedVenues`/`disallowedVenues` are passed, `setVenueFiltering` transaction should also be triggered (given that existing value is `isEnabled` is false

prashantasdeveloper: To explain a bit more, IMO we should have restrictions on the args such as 
```
{
   enabled: true | undefined, 
   allowedVenues?: BigNumber[], 
   disallowedVenues?: BigNumber[]
} || {
  enabled: false
}
```
With disabling the filtering, I should not be allowed to add/remove venues

Thoughts @polymath-eric @VictorVicente

src/api/procedures/setVenueFiltering.ts
```diff
@@ -0,0 +1,96 @@
...
+    transactions.push(TxTags.settlement.SetVenueFiltering);
+  }
+
+  if (allowedVenues?.length) {
+    transactions.push(TxTags.settlement.AllowVenues);
+  }
+
+  if (disallowedVenues?.length) {
+    transactions.push(TxTags.settlement.DisallowVenues);
+  }
```
prashantasdeveloper, line 82: This should not be done when `enabled` param value is false

prashantasdeveloper (changes requested):

src/api/procedures/setVenueFiltering.ts
prashantasdeveloper: I realised this is possible from chain, so we can keep as is

src/api/procedures/setVenueFiltering.ts
prashantasdeveloper: I thought chain would prevent this, but since it allows it maybe we can leave as is

prashantasdeveloper (approved):

PR #995, opened 2023-06-06
feat: 🎸 Add method to check if middleware is synced with chain

prashantasdeveloper:
### Description

Add method to check if middleware is synced with chain

### Breaking Changes

NA

### JIRA Link

DA-742

### Checklist

### DIFF
--- a/src/api/client/Network.ts
+++ b/src/api/client/Network.ts
@@ -551,7 +551,8 @@ export class Network {
   }
 
   /**
-   * Get the number of blocks the middleware needs to process to be synced with chain
+   * Get the number of blocks the middleware needs to process to be synced with chain.
+   * The lag can be around somewhere upto 15 blocks, but this can increase if the block size being processed by the Middleware is too large.
    *
    * @note uses the middleware V2
    */
--- a/src/api/client/Network.ts
+++ b/src/api/client/Network.ts
@@ -549,4 +549,23 @@ export class Network {
   public async getMiddlewareMetadata(): Promise<MiddlewareMetadata | null> {
     return this.context.getMiddlewareMetadata();
   }
+
+  /**
+   * Get the number of blocks the middleware needs to process to be synced with chain
+   *
+   * @note uses the middleware V2
+   */
+  public async getMiddlewareLag(): Promise<BigNumber> {
+    let lastProcessedBlockFromMiddleware = new BigNumber(0);
+    const [latestBlockFromChain, middlewareMetadata] = await Promise.all([
+      this.context.getLatestBlock(),
+      this.context.getMiddlewareMetadata(),
+    ]);
+
+    if (middlewareMetadata) {
+      lastProcessedBlockFromMiddleware = middlewareMetadata.lastProcessedHeight;
+    }
+
+    return latestBlockFromChain.minus(lastProcessedBlockFromMiddleware);
+  }
 }
--- a/src/api/client/Network.ts
+++ b/src/api/client/Network.ts
@@ -553,6 +553,7 @@ export class Network {
   /**
    * Get the number of blocks the middleware needs to process to be synced with chain.
    * The lag can be around somewhere upto 15 blocks, but this can increase if the block size being processed by the Middleware is too large.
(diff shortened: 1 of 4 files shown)


src/api/client/Network.ts
```diff
@@ -579,4 +579,27 @@ export class Network {
...
+    const [latestBlockFromChain, middlewareMetadata] = await Promise.all([
+      this.context.getLatestBlock(),
+      this.getMiddlewareMetadata(),
+    ]);
+
+    if (middlewareMetadata) {
+      lastProcessedBlockFromMiddleware = middlewareMetadata.lastProcessedHeight;
+    }
+
+    return latestBlockFromChain.minus(lastProcessedBlockFromMiddleware).isLessThanOrEqualTo(20);
```
VictorVicente, line 603: Could you extract the number to a const? Also I think 20 is too big. Have you checked in practice how many blocks away is it?

VictorVicente (changes requested):

src/api/client/Network.ts
VictorVicente: If you have seen it on 10, 5 doesn't look safe. In my mind 10 would be enough but if you saw it at that value, maybe 12 or 15 is safer.

src/api/client/Network.ts
```diff
@@ -549,4 +549,29 @@ export class Network {
   public async getMiddlewareMetadata(): Promise<MiddlewareMetadata | null> {
     return this.context.getMiddlewareMetadata();
   }
+
+  /**
+   * Check if the middleware is up to date with the chain.
+   *
+   * @note Since the middleware can take some time in processing large blocks,
+   * the function returns true if the number of processed blocks by the middleware doesn't lag the finalized chain blocks by more than 20.
```
polymath-eric, line 557:
```suggestion
   * the function returns true if the number of processed blocks by the middleware doesn't lag the finalized chain blocks by more than @link{MAX_MIDDLEWARE_LAG}.
```

I forget the syntax, but the number could become inaccurate easily

src/utils/constants.ts
```diff
@@ -136,3 +136,8 @@ export const MAX_LEGS_LENGTH = 10;
  * Default CDD ID associated with an Identity on chain. Used for Identities onboarded without PUIS
  */
 export const DEFAULT_CDD_ID = '0x0000000000000000000000000000000000000000000000000000000000000000';
+
+/**
+ * Maximum lag of block processing allowed in the middleware to indicate that the middleware is in a synced state
+ */
+export const MAX_MIDDLEWARE_LAG = new BigNumber(5);
```
polymath-eric, line 143: Maybe 10? IDK, but I think ~1 minute is reasonable.

30 seconds seems like it could happen on big batch blocks. IDK the right value though

src/api/client/Network.ts
```diff
@@ -549,4 +549,29 @@ export class Network {
...
+   * More details about the last processed block from middleware can be fetched using the `getMiddlewareMetadata` method.
+   *
+   * @note uses the middleware V2
+   */
+  public async isMiddlewareSynced(): Promise<boolean> {
```
polymath-eric, line 562: What about returning how many blocks middleware is laging, instead of returning a true/false value here?

I don't think we can find one value that is right for every application

src/api/client/Network.ts
polymath-eric: I think we should return how many blocks its lagging and let the consumer know about it. Seems like an impossible task to find a single number that is OK for every app imo

Comments should mention expected block time is 6 seconds as well

src/api/client/Network.ts
```diff
@@ -549,4 +549,23 @@ export class Network {
   public async getMiddlewareMetadata(): Promise<MiddlewareMetadata | null> {
     return this.context.getMiddlewareMetadata();
   }
+
+  /**
+   * Get the number of blocks the middleware needs to process to be synced with chain
```
VictorVicente, line 554: I think we should add a comment here recommending a number. To be useful for user, they should know it shouldn't exceed 10 or 15 blocks.

VictorVicente (approved):

PR #1001, opened 2023-06-28
feat: 🎸 Add portfolio params for `modifyInstructionAffirmation`

prashantasdeveloper:
### Description

Add portfolio params for `modifyInstructionAffirmation` procedure
- Adds optional `portfolios` param for `affirm` and `withdraw` methods for Instruction entity
- Adds optional `portfolio` params for `reject` method of Instruction entity

### Breaking Changes

🧨 `Instruction.affirm`, `Instruction.withdraw` and `Instruction.reject` now takes a parameter containing optional portfolio(s) for performing respective operation

### JIRA Link

DA-754

### Checklist

### DIFF
--- a/src/api/entities/Instruction/__tests__/index.ts
+++ b/src/api/entities/Instruction/__tests__/index.ts
@@ -736,7 +736,7 @@ describe('Instruction class', () => {
         )
         .mockResolvedValue(expectedTransaction);
 
-      const tx = await instruction.reject();
+      const tx = await instruction.reject({});
       expect(tx).toBe(expectedTransaction);
     });
   });
@@ -760,7 +760,7 @@ describe('Instruction class', () => {
         )
         .mockResolvedValue(expectedTransaction);
 
-      const tx = await instruction.affirm();
+      const tx = await instruction.affirm({});
 
       expect(tx).toBe(expectedTransaction);
     });
@@ -785,7 +785,7 @@ describe('Instruction class', () => {
         )
         .mockResolvedValue(expectedTransaction);
 
-      const tx = await instruction.withdraw();
+      const tx = await instruction.withdraw({});
 
       expect(tx).toBe(expectedTransaction);
     });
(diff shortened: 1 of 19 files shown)


src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -175,16 +185,104 @@ export async function getAuthorization(
...
+
+  const isParam = (legPortfolio: DefaultPortfolio | NumberedPortfolio): boolean => {
+    const { did: legPortfolioDid, number: legPortfolioNumber } =
+      portfolioLikeToPortfolioId(legPortfolio);
+    return (
+      !portfolioIdParams.length ||
+      portfolioIdParams.some(
+        ({ did, number }) =>
+          did === legPortfolioDid &&
+          new BigNumber(legPortfolioNumber || 0).eq(new BigNumber(number || 0))
```
polymath-eric, line 251: I think here we don't need to convert to BigNumber to compare since these aren't exposed anywhere, just a logic check with something that should fit into normal JS numbers.

OK to leave

polymath-eric (approved):
LGTM. sonar cloud complained about complexity, its giving big penalties for nested ifs and you are only over by 1 so I think refactor is better than ignoring, but either option works

polymath-eric:
linting on CI got stuck installing dependencies for some reason. I haven't seen that before. I restarted it for now

src/api/entities/Instruction/__tests__/index.ts
```diff
@@ -736,7 +736,7 @@ describe('Instruction class', () => {
         )
         .mockResolvedValue(expectedTransaction);
 
-      const tx = await instruction.reject();
+      const tx = await instruction.reject({});
```
VictorVicente, line 739: This is a breaking change. Is there any way to avoid requiring an explicit empty arg?

VictorVicente (changes requested):

src/api/entities/Instruction/__tests__/index.ts
prashantasdeveloper: As of now, the way `createProcedureMethod` is designed, we can't support that. https://github.com/PolymeshAssociation/polymesh-sdk/blob/alpha/src/utils/internal.ts#L641. Similar type of change is supposed to go for [this PR](https://github.com/PolymeshAssociation/polymesh-sdk/pull/1004) as well.

VictorVicente (approved):

PR #996, opened 2023-06-06
feat: 🎸 Add middleware V2 compatibility check on initialization

prashantasdeveloper:
### Description

Adds compatibility check for Middleware V2 URL with chain node URL. Throws if URL is invalid or points to another chain.

### Breaking Changes

🧨 `Polymesh.connect` now throws if middleware V2 URL is not working or is incompatible with the node URL

### JIRA Link

DA-761

### Checklist

- [ ] Updated the Readme.md (if required) ?

PR #1004, opened 2023-06-29
feat: 🎸 allow for skiping affirm for manual execution

polymath-eric:
### Description

instruction `executeManually` method now accepts optional `skipAffirmationCheck` param. Useful when processing the affirmation earlier in the batch transaction.

### Breaking Changes

None

### JIRA Link

[DA-756](https://polymesh.atlassian.net/browse/DA-756)

### Checklist

[DA-756]: https://polymesh.atlassian.net/browse/DA-756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/procedures/types.ts
+++ b/src/api/procedures/types.ts
@@ -526,9 +526,12 @@ export type ModifyInstructionAffirmationParams = AffirmInstructionParams &
       } & RejectInstructionParams)
   );
 
-export interface ExecuteManualInstructionParams {
-  id: BigNumber;
-}
+export type ExecuteManualInstructionParams = AffirmInstructionParams & {
+  /**
+   * (optional) Set to `true` to skip affirmation check, useful for batch transactions
+   */
+  skipAffirmationCheck?: boolean;
+};
 
 export interface CreateVenueParams {
   description: string;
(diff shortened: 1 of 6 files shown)


polymath-eric:
I made this a breaking change after trying out adding a new endpoint.

I am thinking an integrator using `executeManual` probably wants control, so shouldn't be too inconvenienced at inspecting the call site and deciding on the correct behavior.

By principle I dislike having boolean params that toggle internal logic, but I think this case is worth while making an exception for. `instruction.executeManualButSkipAffirmationCheck` doesn't look good as a name. Having to specify details when opting for manual execution seems reasonable and should be easy to update.

We have a breaking change here, so I am hoping these can go out together since they are both relatively minor: https://github.com/PolymeshAssociation/polymesh-sdk/pull/996

prashantasdeveloper (approved):

VictorVicente (changes requested):
Standby to check if this breaking change can be avoided

src/api/procedures/types.ts
```diff
@@ -488,7 +488,7 @@ export type AddInstructionWithVenueIdParams = AddInstructionParams & {
   venueId: BigNumber;
 };
 
-export interface AffirmInstructionParams {
+export interface InstructionIdParams {
```
polymath-eric, line 491: I think this is a more accurate name for it. In the affirm method the intent is still clear due to the method name.

PR #1007, opened 2023-07-03
feat: 🎸 Add method to get involved instructions

prashantasdeveloper:
### Description

This adds `Identity.getInvolvedInstructions` which returns all instructions where identity is involved grouped by their role (owner/custodian) and status

### Breaking Changes

NA

### JIRA Link

DA-755

### Checklist

### DIFF
--- a/src/api/entities/Identity/index.ts
+++ b/src/api/entities/Identity/index.ts
@@ -6,7 +6,7 @@ import {
 } from '@polkadot/types/lookup';
 import BigNumber from 'bignumber.js';
 import P from 'bluebird';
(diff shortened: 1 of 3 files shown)


src/api/entities/Identity/index.ts
```diff
@@ -605,13 +630,76 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
       });
     });
 
+    console.log({
```
VictorVicente, line 633: These should be removed

VictorVicente (approved):

PR #1017, opened 2023-08-02
fix: 🐛 instruction affirmation does not need leg count

polymath-eric:
### Description

6.0 removes the need to specify leg count for affirmations. Previously would fail with expecting 2 got 3 args error

### Breaking Changes

None

### JIRA Link

N/A

### Checklist


VictorVicente (approved):

VictorVicente (commented):
There arar some `never used` imports to be removed

PR #1020, opened 2023-08-10
feat: 🎸 support chain v5.4 - v6.0

polymath-eric:
### Description

add multi major version support for the SDK to function with either 5.4 and 6.0 chains

this version supports chains 5.4 and 6.0, with a feature flag type approach to dynamically switch between the different calls as soon as the update happens

this version undos the breaking changes introduced in 78c95ae. This should be API compatible with SDK v21.0.0

When creating checkpoints there will only be a finite set of points after the 6.0 upgrade

WARNING: Get Checkpoint Schedule details will have incorrect period details after 6.0 upgrade occurs

### Checklist


polymath-eric:
The coverage is low because `isV5` is meant to be ripped out ASAP, which will bring up our code.coverage again

src/api/entities/Instruction/types.ts
```diff
@@ -6,8 +6,7 @@ import { EventIdentifier } from '~/types';
 export enum InstructionStatus {
   Pending = 'Pending',
   Failed = 'Failed',
-  Rejected = 'Rejected',
-  Success = 'Success',
+  Executed = 'Executed',
```
VictorVicente, line 9: This is a good example on what awful dual version supporting is. 
When we decide to remove v5 support, we'll need to reintroduce this breaking change.

VictorVicente (approved):

PR #1025, opened 2023-08-23
feat: 🎸 select portfolio on asset issuance

sansan:
### Description

Adds posibility to set portfolio to which the Asset will be issued when creating asset (if the initial supply gt 0) and when issuing tokens;

### JIRA Link

https://polymesh.atlassian.net/browse/DA-833

### Checklist

### DIFF
--- a/src/api/procedures/issueTokens.ts
+++ b/src/api/procedures/issueTokens.ts
@@ -51,7 +51,6 @@ export async function prepareIssueTokens(
     });
   }
 
-  console.log({ portfolioId: portfolioId?.toString() });
 
   const portfolio = portfolioId
     ? await signingIdentity.portfolios.getPortfolio({ portfolioId })
--- a/src/api/procedures/issueTokens.ts
+++ b/src/api/procedures/issueTokens.ts
@@ -9,6 +9,7 @@ import { bigNumberToBalance, portfolioToPortfolioKind, stringToTicker } from '~/
 export interface IssueTokensParams {
   amount: BigNumber;
   ticker: string;
+  portfolioId?: BigNumber;
 }
 
 export interface Storage {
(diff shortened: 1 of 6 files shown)


src/api/procedures/issueTokens.ts
```diff
@@ -50,11 +51,15 @@ export async function prepareIssueTokens(
     });
   }
 
-  const defaultPortfolio = await signingIdentity.portfolios.getPortfolio();
+  console.log({ portfolioId: portfolioId?.toString() });
```
polymath-eric, line 54:
```suggestion
```

src/api/procedures/issueTokens.ts
```diff
@@ -50,11 +51,15 @@ export async function prepareIssueTokens(
...
+
+  const portfolio = portfolioId
```
polymath-eric, line 56: I guess we could make this 5.4 compatible by also checking the `isV5` param, but imo its better to wait for us to release the dual version, clean up, and then merge this PR in vs offering an API that won't work as expected on 5.4

src/api/procedures/issueTokens.ts
```diff
@@ -9,6 +9,7 @@ import { bigNumberToBalance, portfolioToPortfolioKind, stringToTicker } from '~/
 export interface IssueTokensParams {
   amount: BigNumber;
   ticker: string;
+  portfolioId?: BigNumber;
```
polymath-eric, line 12: Do we have to worry about custodian portfolios? We might want to allow for a full portfolio ID here (`portfolio: {did: string; id: BigNumber}`).

You might need to check with core what the full intention is here.

polymath-eric: seems like its OK inferring the DID here

polymath-eric (approved):
lgtm

VictorVicente (approved):

PR #1027, opened 2023-08-23
fix: 🐛 unwrap fundraiser for 5.4 chains

polymath-eric:
### Description

fundraiser is an option if 5.4 so should be unwrapped there as well

### Checklist


polymath-eric:
In our API `offerings.getOne` returns `Offering` but `offerings.get` returns `OfferingWithDetails`. My last PR I tested locally only with `get` so missed the real details method.

VictorVicente (approved):

prashantasdeveloper (approved):

PR #1031, opened 2023-08-29
feat: update batch transaction

sansan:
### Description

Updates the tag for batch tx

### JIRA Link

https://polymesh.atlassian.net/browse/DA-834
### Checklist


polymath-eric (commented):
What about this when the batch fails?

https://github.com/PolymeshAssociation/polymesh-sdk/blob/alpha/src/base/PolymeshTransactionBatch.ts#L251

I was expecting a different event to get published in this case

VictorVicente (approved):

PR #1034, opened 2023-08-29
perf: ⚡️ improve startup time

polymath-eric:
add polkadot options so people can provide metadata to improve startup time. optimizes the rest of `Polymesh.connect` to further improve startup time

- parallizes version checks on connect
- lazy evaluates `isArchiveNode` when needed
- add `polkadot` option to connect. consumer can provide metadata

✅ Closes: DA-838

### Checklist


VictorVicente:
Is there any reason to keep it in `alpha`? Also there is a line we still need to cover.

PR #1040, opened 2023-09-06
fix: 🐛 Use `9.0.1` version for `bignumber.js`

prashantasdeveloper:
### Description

The `9.1.2` version of `bignumber.js` doesn't export `./bignumber` path in it's package.json file which errors as `Error
[ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './bignumber' is not defined by "exports"`

### Breaking Changes

NA

### JIRA Link

NA

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "@polkadot/api": "10.9.1",
     "@polkadot/util": "12.4.2",
     "@polkadot/util-crypto": "12.4.2",
-    "bignumber.js": "9.1.2",
+    "bignumber.js": "9.0.1",
     "bluebird": "^3.7.2",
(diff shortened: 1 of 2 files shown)


VictorVicente (approved):

polymath-eric (approved):

polymath-eric:
where did you run into this error?

polymath-eric (changes requested):
lets see if there is a better way than downgrading for now

polymath-eric:
Seems like the error happens on node `18.13.0` but is OK on node `18.16.0`

Makes sense to downgrade for now so others don't run into this issue

polymath-eric (approved):

PR #1041, opened 2023-09-07
fix: readme file (minimum node version support)

MehdiRaash:
### Description

NodeJS version 16 (LTS) is the minimum supported version for executing Polymesh SDK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ This document assumes you are already familiar with [Security Tokens](https://th
 
 ### Technical Pre-requisites
 
-In order to use the Polymath SDK, you must install [node](https://nodejs.org/) \(version 14\) and [npm](https://www.npmjs.com/). The library is written in [typescript](https://www.typescriptlang.org/), but can also be used in plain javascript. This document will assume you are using typescript, but the translation to javascript is very simple.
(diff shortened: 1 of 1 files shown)


polymath-eric (approved):
LGTM. Thanks for the PR

VictorVicente (approved):

polymath-eric (commented):
how did you make this commit? We use a bot that is picky about its commit format.

Instead of doing normal `git commit` you should have ran `yarn commit` and follow the steps there.

Should be OK to merge as is, but tooling is supposed to stop it from being committed too.

MehdiRaash:
> how did you make this commit? We use a bot that is picky about its commit format.
> 
> Instead of doing normal `git commit` you should have ran `yarn commit` and follow the steps there.
> 
> Should be OK to merge as is, but tooling is supposed to stop it from being committed too.

I did it using the web based VS code on Github.
I can change the commit title to `fix:...`, Would that fix the bot's complaint?

polymath-eric:
We are OK to merge this in as it is.

the CI is stuck because we don't allow for the sonar token to be imported into a fork, but its just a readme update so there isn't anything to catch here.

PR #1043, opened 2023-09-15
feat: 🎸 Adds new `ChildIdentity` entity and other related functions

prashantasdeveloper:
### Description

Adds new `ChildIdentity` entity and other related functions.

Changes in `Identities` namespace - 
* Adds new procedure to create a child identity
* Adds new method to get an instance of `ChildIdentity`

Changes in `Identity` entity - 
* Adds new method to get all child identities
* Adds new procedure to unlink a child

Adds new `ChildIdentity` entity which extends the existing `Identity` entity with addition of following - 
* Method to get parent 
* Procedure to unlink from parent

### Breaking Changes

NA

### JIRA Link

DA-864, DA-866, DA-867, DA-868

### Checklist


src/base/Context.ts
```diff
@@ -575,6 +583,30 @@ export class Context {
...
+   *
+   * Returns an Child Identity when given a DID string
+   *
+   * @throws if the Child Identity does not exist
+   */
+  public async getChildIdentity(child: ChildIdentity | string): Promise<ChildIdentity> {
+    if (child instanceof ChildIdentity) {
+      return child;
+    }
+    const id = new ChildIdentity({ did: child }, this);
```
polymath-eric, line 597:
```suggestion
    const childIdentity = new ChildIdentity({ did: child }, this);
```
or `childId`

Just `id` made me think its a number at first

polymath-eric (commented):
left some comments. Only issue I really see is calling `parentDid.entries` when there should be an inverse storage like `childDids` to use.

We'll probably be OK for a while, but if child identities start to get used it will start to get slow for everyone. Maybe its OK for `alpha`, but we shouldn't have performance traps for `master` or `lts`.

src/api/entities/Identity/ChildIdentity.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    const parentDid = identityIdToString(rawParentDid.unwrap());
+
+    return new Identity({ did: parentDid }, context);
+  }
+
+  /**
+   * @hidden
+   * since a child Identity doesn't has any other children, this method overrides the base implementation to return empty array
```
VictorVicente, line 55: Should it return an empty array or should it throw? Not sure about it.

src/base/Context.ts
```diff
@@ -575,6 +583,30 @@ export class Context {
...
 
+  /**
+   * @hidden
+   *
+   * Returns an Child Identity when given a DID string
+   *
+   * @throws if the Child Identity does not exist
+   */
+  public async getChildIdentity(child: ChildIdentity | string): Promise<ChildIdentity> {
+    if (child instanceof ChildIdentity) {
```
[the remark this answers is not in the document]
prashantasdeveloper: I just kept it similar to `getIdentity`. Leaving this as is

src/api/entities/Identity/ChildIdentity.ts
prashantasdeveloper: I'm OK with both, but I preferred an empty array since throw requires the end user to handle this separately (which personally I think is an overkill)

src/api/entities/Identity/__tests__/ChildIdentity.ts
```diff
@@ -0,0 +1,118 @@
...
+    procedureMockUtils.reset();
+  });
+
+  afterAll(() => {
+    dsMockUtils.cleanup();
+    procedureMockUtils.reset();
+  });
+
+  it('should extend Entity', () => {
+    expect(ChildIdentity.prototype instanceof Entity).toBe(true);
```
polymath-eric, line 50:
```suggestion
    expect(ChildIdentity).toBeInstanceOf(Entity);
```

src/api/entities/Identity/index.ts
```diff
@@ -800,4 +812,42 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
       middlewareInstructionToHistoricInstruction(instruction!, context)
     );
   }
+
+  /**
+   * Returns the list of all child identities
+   *
+   * @note this query can be potentially **SLOW** depending on the number of parent Identities present on the chain
```
polymath-eric, line 819:
```suggestion
   * @note this query can be potentially slow depending on the number of parent Identities present on the chain
```

I am not sure if we should even mention this. Hopefully we have a better chain method before its an issue, and its not like developers can do much other than complain and us add to SubQuery or ask core to add a storage value.

polymath-eric: I guess we could monitor with a cronjob, like:

```ts
const start = new Date()
await identity.getChildIdentities()

if(new Date() - start > 1000) postToSlack()
```

polymath-eric (approved):

src/api/entities/Identity/index.ts
prashantasdeveloper: I personally didn't wanted to add the comment but it was discussed in the call to add in to explicitly specify the slowness so here it is. Also, the slow comment was mentioned the same way somewhere else that's why I followed the caps bold pattern for **SLOW**. Also, where do we add the cron?

PR #1044, opened 2023-09-18
feat: 🎸 add NftCollection entity + create procedure

polymath-eric:
### Description

Add `NftCollection` entity and  `assets.createNftCollection` method to Assets namespace to create it

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-876](https://polymesh.atlassian.net/browse/DA-876)

### Checklist


src/api/entities/Asset/Metadata.ts
```diff
@@ -123,4 +123,24 @@ export class Metadata extends Namespace<Asset> {
 
     return new MetadataEntry({ ticker, type, id }, context);
   }
+
+  /**
+   * Gets the next local metadata ID for the Asset
+   */
+  public async getNextLocalId(): Promise<BigNumber> {
```
VictorVicente, line 130: Is it useful to be public? Do we expect users to need it?

src/api/entities/Asset/NonFungible/NftCollection.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Properties that uniquely identify an Asset
+ */
+export interface UniqueIdentifiers {
+  /**
+   * ticker of the NftCollection
+   */
+  ticker: string;
+}
```
VictorVicente, line 30: Should we import it from `BaseAsset` too? To avoid duplication

src/api/entities/Asset/NonFungible/NftCollection.ts
```diff
@@ -0,0 +1,143 @@
...
+      })
+    );
+
+    return optionize(middlewareEventDetailsToEventIdentifier)(asset?.createdBlock, asset?.eventIdx);
+  }
+
+  /**
+   * Retrieve the amount of unique investors that hold this Nft
+   */
+  public async investorCount(): Promise<BigNumber> {
```
VictorVicente, line 101: `investorCount` doesn't sound right for NFTs. I know we shouldn't change it but could you ask Product team if wording is OK?

VictorVicente (changes requested):

src/api/entities/Asset/Metadata.ts
polymath-eric: Users shouldn't need it but if we don't expose it as `public` then I won't be able to call it from where I am calling it.

I could inline it where its called, but it makes sense to me here. Otherwise we'd duplicate this. I am putting the `plus(1)` in this function to hide the "next" id is actually "latest" id

I could add a `@hidden` so it isn't visible in generated documentation.

src/api/entities/Asset/NonFungible/NftCollection.ts
polymath-eric: There is also "totalSupply" in details, and that doesn't sound right to me.

I think its better to keep the same name to minimize the need for type casting, but we could name this something else if we really wanted to.

prashantasdeveloper (dismissed):

src/api/entities/Asset/__tests__/Metadata.ts
```diff
@@ -209,4 +209,27 @@ describe('Metadata class', () => {
...
+    let id: BigNumber;
+    let rawId: u64;
+
+    beforeEach(() => {
+      id = new BigNumber(1);
+      rawId = dsMockUtils.createMockU64(id);
+    });
+
+    afterAll(() => {
+      jest.restoreAllMocks();
```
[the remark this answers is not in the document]
polymath-eric: Made a story for this. There is quite a few files with it so will be a big diff, but easy to review when isolated.

https://polymesh.atlassian.net/browse/DA-889

src/api/entities/Asset/Metadata.ts
VictorVicente: Yeah, I think adding `@hidden` is the best solution here.

VictorVicente (approved):

PR #1047, opened 2023-09-26
feat: 🎸 add NonFungible details method

polymath-eric:
### Description

moves current `asset.details` method into `BaseAsset`. `NftCollection` wraps the base call to return `totalSupply` property.

NftCollection needs to make an additional call to the chain and iterate all holders to calculate `totalSupply` of the Nfts, which is the number of held NFTs in the collection.

A deferred optimization is to leverage middleware (SubQuery) when provided. However it is expected a single collection would need hundreds of unique holders before performance improvements become noticable.

### Checklist

### DIFF
--- a/src/api/entities/Asset/BaseAsset/index.ts
+++ b/src/api/entities/Asset/BaseAsset/index.ts
@@ -1,9 +1,23 @@
-import { Context, Entity } from '~/internal';
-import { SecurityIdentifier, SubCallback, UnsubCallback } from '~/types';
+import { Bytes, Option, StorageKey } from '@polkadot/types';
+import {
+  PalletAssetSecurityToken,
+  PolymeshPrimitivesAgentAgentGroup,
+  PolymeshPrimitivesIdentityId,
+  PolymeshPrimitivesTicker,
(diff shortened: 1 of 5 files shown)


src/api/entities/Asset/BaseAsset/index.ts
```diff
@@ -122,4 +136,99 @@ export abstract class BaseAsset extends Entity<UniqueIdentifiers, string> {
...
     return boolToBoolean(result);
   }
+
+  /**
+   * Retrieve the Asset's data
+   *
+   * @note can be subscribed to
+   */
+  protected internalDetails(): Promise<AssetDetails>;
+  protected internalDetails(callback: SubCallback<AssetDetails>): Promise<UnsubCallback>;
```
polymath-eric, line 146: At the time of writing I thought I wasn't able to access `super` in `NftCollection`, but it seems like that is possible after all. Going to see if I can leave this as just `details`

polymath-eric:
This is included in this PR: https://github.com/PolymeshAssociation/polymesh-sdk/pull/1048

PR #1048, opened 2023-09-26
refactor: 💡 rename `Asset` to `FungibleAsset`

polymath-eric:
### Description

Share functionality between fungible and nft collections with `BaseAsset` entity. Introduce `getFungibleAsset` and `getNftCollection` methods in the assets namespace. `getAsset` and `get` now return `FungibleAsset | NftCollection`. These results need to be narrowed for actions unique to each asset (e.g. issuance), but should be unnessesary for common actions (e.g. calling `.details()`). `FungibleAsset` retains all behavior of the previous `Asset` class

### Breaking Changes

BREAKING CHANGE: 🧨 Asset namespace get and getAssets now returns `FungibleAsset | NftCollection`. 
BREAKING CHANGE: 🧨`Asset` entity was renamed to `FungibleAsset`. 
BREAKING CHANGE: 🧨`FungibleAsset` procedures no longer return `FungibleAsset` model. Affected methods include:  `documents.set`, `trustedClaimIssuers.{set, add, remove}` , `requirements.{set, add, remove, reset, pause, unpause}`. The entity remains unchanged after these operations and could cause unnecessary change detection if used. To get the newly set values the appropriate get function should be called instead (e.g. `await setDocsTx.run(); const newDocs = documents.get()`)

### JIRA Link

✅ Closes: [DA-881](https://polymesh.atlassian.net/browse/DA-881)

### Checklist


src/api/entities/Asset/NonFungible/NftCollection.ts
```diff
@@ -35,10 +51,6 @@ export class NftCollection extends BaseAsset {
       { getProcedureAndArgs: args => [transferAssetOwnership, { ticker, ...args }] },
       context
     );
-    this.controllerTransfer = createProcedureMethod(
-      { getProcedureAndArgs: args => [controllerTransfer, { ticker, ...args }] },
-      context
-    );
```
VictorVicente, line 41: Why `transferOwnership` is not defined on `BaseAsset`?

src/utils/typeguards.ts
```diff
@@ -153,13 +152,6 @@ export function isNumberedPortfolio(value: unknown): value is NumberedPortfolio
   return value instanceof NumberedPortfolio;
 }
 
-/**
- * Return whether value is an Asset
- */
-export function isAsset(value: unknown): value is Asset {
-  return value instanceof Asset;
-}
-
```
VictorVicente, line 162: Should we provide TypeGuards for both kind of assets?

VictorVicente (approved):

src/api/entities/Asset/NonFungible/NftCollection.ts
polymath-eric: I think I got mixed up with `controllerTransfer` which was mentioned to only work for FungibleAssets atm.

Agreed `transferOwnership` should be on `BaseAsset`

src/utils/typeguards.ts
polymath-eric: Yeah, thats probably a good idea

PR #1051, opened 2023-09-30
feat: 🎸 registerCustomClaimType

sansan:
### Description

Adds procedure to register CustomClaimType

### JIRA Link

https://polymesh.atlassian.net/browse/DA-882

### Checklist

### DIFF
--- a/src/api/procedures/__tests__/registerCustomClaimType.ts
+++ b/src/api/procedures/__tests__/registerCustomClaimType.ts
@@ -74,7 +74,7 @@ describe('registerCustomClaimType procedure', () => {
     jest.restoreAllMocks();
   });
 
-  it('should throw an error if attempting to add a CustomClaimType already present', () => {
+  it('should throw an error if attempting to add a CustomClaimType with a name exceeding the max length', () => {
     params = {
       ...params,
       name: 'NAME_EXCEEDING_MAX_LENGTH',
@@ -84,9 +84,6 @@ describe('registerCustomClaimType procedure', () => {
     const expectedError = new PolymeshError({
       code: ErrorCode.ValidationError,
       message: 'CustomClaimType name length exceeded',
-      data: {
-        maxLength: customClaimTypeNameMaxLength,
-      },
     });
 
     return expect(prepareRegisterCustomClaimType.call(proc, params)).rejects.toThrowError(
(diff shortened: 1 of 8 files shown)


src/api/procedures/__tests__/registerCustomClaimType.ts
```diff
@@ -0,0 +1,169 @@
...
+    dsMockUtils.reset();
+  });
+
+  afterAll(() => {
+    procedureMockUtils.cleanup();
+    dsMockUtils.cleanup();
+    jest.restoreAllMocks();
+  });
+
+  it('should throw an error if attempting to add a CustomClaimType already present', () => {
```
polymath-eric, line 77:
```suggestion
  it('should throw an error if attempting to add a CustomClaimType with a name exceeding the max length', () => {
```

src/api/procedures/__tests__/registerCustomClaimType.ts
```diff
@@ -0,0 +1,169 @@
...
+        maxLength: customClaimTypeNameMaxLength,
+      },
+    });
+
+    return expect(prepareRegisterCustomClaimType.call(proc, params)).rejects.toThrowError(
+      expectedError
+    );
+  });
+
+  it('should throw an error if attempting to register a CustomClaimType with duplicate name', async () => {
```
polymath-eric, line 97:
```suggestion
  it('should throw an error if attempting to add a CustomClaimType already present', async () => {
```

Nit, but I prefer the wording from the above case description

src/api/procedures/__tests__/registerCustomClaimType.ts
```diff
@@ -0,0 +1,169 @@
...
+    params = {
+      ...params,
+      name: 'NAME_EXCEEDING_MAX_LENGTH',
+    };
+    const proc = procedureMockUtils.getInstance<Params, BigNumber>(mockContext);
+
+    const expectedError = new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message: 'CustomClaimType name length exceeded',
+      data: {
```
polymath-eric, line 87: From my testing `toThrow`/`toThrowError` assertions only check `code` and `message`, they don't examine the `data` property. 

I'd recommend removing it here it since its less lines, and makes it clear this isn't a property that is actually being asserted on.


```suggestion
```

polymath-eric (changes requested):
Overall looks good to me. I agree it should be moved to the claims namespace and not identities though.

Its OK to come in a later PR, but we will also need a getter so people can check if its registered or not.

VictorVicente (approved):

PR #1052, opened 2023-10-02
feat: 🎸 get CustomClaimType by id or name

sansan:
### Description

adds a method to check if CustomClaimType exists using either id or name
returns { id, name } if found or null if not

### JIRA Link

https://polymesh.atlassian.net/browse/DA-894

### Checklist

### DIFF
--- a/src/api/client/Claims.ts
+++ b/src/api/client/Claims.ts
@@ -4,7 +4,7 @@ import { filter, flatten, isEqual, uniqBy, uniqWith } from 'lodash';
 import { Context, Identity, modifyClaims, registerCustomClaimType } from '~/internal';
 import { ClaimTypeEnum } from '~/middleware/enums';
 import { claimsGroupingQuery, claimsQuery } from '~/middleware/queries';
-import { ClaimsGroupBy, ClaimsOrderBy, Query } from '~/middleware/types';
+import { ClaimsGroupBy, ClaimsOrderBy, CustomClaimType, Query } from '~/middleware/types';
 import {
   CddClaim,
   ClaimData,
@@ -35,11 +35,6 @@ import {
   u32ToBigNumber,
 } from '~/utils/conversion';
 import { calculateNextKey, createProcedureMethod, getDid, removePadding } from '~/utils/internal';
-
-type CustomClaimTypeToHuman = {
-  name: string;
-  id: string;
-};
 
 /**
  * Handles all Claims related functionality
@@ -493,13 +488,11 @@ export class Claims {
   public registerCustomClaimType: ProcedureMethod<RegisterCustomClaimTypeParams, BigNumber>;
 
   /**
-   * Retrieves a custom claim type based on its name.
+   * Retrieves a custom claim type based on its name
    *
-   * @param name - The name of the custom claim type to retrieve.
-   *
-   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
+   * @param name - The name of the custom claim type to retrieve
    */
-  public async getCustomClaimTypeByName(name: string): Promise<CustomClaimTypeToHuman | null> {
+  public async getCustomClaimTypeByName(name: string): Promise<CustomClaimType | null> {
     const {
       context: {
         polymeshApi: {
@@ -510,21 +503,19 @@ export class Claims {
 
     const customClaimTypeIdOpt = await identity.customClaimsInverse(name);
 
-    if (!customClaimTypeIdOpt.isSome) {
+    if (customClaimTypeIdOpt.isEmpty) {
       return null;
     }
 
-    return { id: u32ToBigNumber(customClaimTypeIdOpt.value).toString(), name };
+    return { id: u32ToBigNumber(customClaimTypeIdOpt.value), name };
   }
 
   /**
-   * Retrieves a custom claim type based on its ID.
+   * Retrieves a custom claim type based on its ID
    *
-   * @param id - The ID of the custom claim type to retrieve.
-   *
-   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
+   * @param id - The ID of the custom claim type to retrieve
    */
-  public async getCustomClaimTypeById(id: BigNumber): Promise<CustomClaimTypeToHuman | null> {
+  public async getCustomClaimTypeById(id: BigNumber): Promise<CustomClaimType | null> {
     const {
       context: {
         polymeshApi: {
@@ -535,10 +526,10 @@ export class Claims {
 
     const customClaimTypeIdOpt = await identity.customClaims(bigNumberToU32(id, this.context));
 
-    if (!customClaimTypeIdOpt.isSome) {
+    if (customClaimTypeIdOpt.isEmpty) {
       return null;
     }
 
-    return { id: id.toString(), name: bytesToString(customClaimTypeIdOpt.value) };
+    return { id, name: bytesToString(customClaimTypeIdOpt.value) };
   }
 }
--- a/src/api/client/Claims.ts
+++ b/src/api/client/Claims.ts
@@ -1,13 +1,7 @@
 import BigNumber from 'bignumber.js';
 import { filter, flatten, isEqual, uniqBy, uniqWith } from 'lodash';
 
-import {
-  Context,
-  Identity,
-  modifyClaims,
-  PolymeshError,
-  registerCustomClaimType,
-} from '~/internal';
+import { Context, Identity, modifyClaims, registerCustomClaimType } from '~/internal';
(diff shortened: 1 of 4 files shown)


src/api/client/Claims.ts
```diff
@@ -483,4 +503,77 @@ export class Claims {
...
+   * @param opts - The options object used for retrieving the custom claim type.
+   * @param opts.name - The name of the custom claim type to retrieve.
+   * @param opts.id - The ID of the custom claim type to retrieve.
+   *
+   * @throws {PolymeshError}
+   * Throws a PolymeshError if neither `name` nor `id` is provided in the options.
+   *
+   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
+   */
+  public async getCustomClaimType(
```
polymath-eric, line 521: Personally I think exposing `getCustomClaimTypeByName` and `getCustomClaimTypeById` is a more intuitive API.

src/api/client/Claims.ts
```diff
@@ -483,4 +503,77 @@ export class Claims {
...
+  ): Promise<CustomClaimTypeToHuman | null> {
+    if (opts.name) {
+      return this.getCustomClaimTypeByName(opts.name);
+    }
+
+    if (opts.id) {
+      return this.getCustomClaimTypeById(opts.id);
+    }
+
+    throw new PolymeshError({
```
polymath-eric, line 532: This is a case that TS would be able to validate at compile time with the other methods

src/api/client/Claims.ts
```diff
@@ -483,4 +503,77 @@ export class Claims {
...
+   * @param opts.name - The name of the custom claim type to retrieve.
+   * @param opts.id - The ID of the custom claim type to retrieve.
+   *
+   * @throws {PolymeshError}
+   * Throws a PolymeshError if neither `name` nor `id` is provided in the options.
+   *
+   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
+   */
+  public async getCustomClaimType(
+    opts: Partial<CustomClaimType>
```
polymath-eric, line 522:
```suggestion
    args: {id: BigNumber} | {name: string}
```

I think something like this is a clearer type, but TS will make it a pain to differentiate them internally

polymath-eric (changes requested):
I think it would be better to make the `byName` and `byId` methods public and get rid of the routing method `getCustomClaimType`. 

I don't really see the benefit since a developer needs to know what they have, but the signature gives them less TS checking.

src/api/client/Claims.ts
polymath-eric: Did you try with 
```ts
  // eslint-disable-next-line require-jsdoc
 ```
 
 We use overloads in other places and that's the only line that's needed to be added. 
 
Maybe `never` can help here, but I am not sure if this is the right situation for it.
 
 ```ts
 {id: BigNumber; name: never} | {id: never; name: string}
 ```
 
 I think this lets TS know about the other field, but also not let someone pass both in.

File: src/api/client/Claims.ts [hunk 4]
```diff
@@ -23,16 +23,24 @@ import {
...
   stringToIdentityId,
   toIdentityWithClaimsArray,
+  u32ToBigNumber,
 } from '~/utils/conversion';
 import { calculateNextKey, createProcedureMethod, getDid, removePadding } from '~/utils/internal';
 
+type CustomClaimTypeToHuman = {
+  name: string;
+  id: string;
+};
```
prashantasdeveloper, line 42:
```suggestion
export type CustomClaimType = {
  name: string;
  id: BigNumber;
};
```
Why not just have the name to `CustomClaimType`? Also since this is used as part of return type of a method we need to export it so that docs don't complain (so maybe move it to types file)

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
    *  - a custom claim type with the same `name` already exists
    */
   public registerCustomClaimType: ProcedureMethod<RegisterCustomClaimTypeParams, BigNumber>;
+
+  /**
+   * Retrieves a custom claim type based on its name.
+   *
+   * @param name - The name of the custom claim type to retrieve.
+   *
+   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
```
prashantasdeveloper, line 500:
```suggestion
   * Retrieves a custom claim type based on its name
   *
   * @param name - The name of the custom claim type to retrieve
```

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
...
+      context: {
+        polymeshApi: {
+          query: { identity },
+        },
+      },
+    } = this;
+
+    const customClaimTypeIdOpt = await identity.customClaimsInverse(name);
+
+    if (!customClaimTypeIdOpt.isSome) {
```
prashantasdeveloper, line 513:
```suggestion
    if (customClaimTypeIdOpt.isEmpty) {
```

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
...
+      return null;
+    }
+
+    return { id: u32ToBigNumber(customClaimTypeIdOpt.value).toString(), name };
```
prashantasdeveloper, line 517:
```suggestion
    return { id: u32ToBigNumber(customClaimTypeIdOpt.value), name };
```
if we change the type of id to `BigNumber`

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
...
+  }
+
+  /**
+   * Retrieves a custom claim type based on its ID.
+   *
+   * @param id - The ID of the custom claim type to retrieve.
+   *
+   * @returns A promise that resolves to the `CustomClaimTypeToHuman` object if found, or null otherwise.
```
prashantasdeveloper, line 525:
```suggestion
   * Retrieves a custom claim type based on its ID
   *
   * @param id - The ID of the custom claim type to retrieve
```

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
...
+      context: {
+        polymeshApi: {
+          query: { identity },
+        },
+      },
+    } = this;
+
+    const customClaimTypeIdOpt = await identity.customClaims(bigNumberToU32(id, this.context));
+
+    if (!customClaimTypeIdOpt.isSome) {
```
prashantasdeveloper, line 538:
```suggestion
    if (customClaimTypeIdOpt.isEmpty) {
```

src/api/client/Claims.ts
```diff
@@ -483,4 +491,54 @@ export class Claims {
...
+      return null;
+    }
+
+    return { id: id.toString(), name: bytesToString(customClaimTypeIdOpt.value) };
```
prashantasdeveloper, line 542:
```suggestion
    return { id, name: bytesToString(customClaimTypeIdOpt.value) };
```

prashantasdeveloper (changes requested):
Some nitty changes

src/api/client/Claims.ts
polymath-eric: My nitty preference is `isNone` for options, but I looked it up and `isEmpty` and `isNone` are the same thing for `Option`. I agree about getting rid of the `!` is an improvement

src/middleware/types.ts
```diff
@@ -34355,3 +34357,8 @@ export type _Metadata = {
   specName?: Maybe<Scalars['String']>;
   targetHeight?: Maybe<Scalars['Int']>;
 };
+
+export type CustomClaimType = {
+  name: string;
+  id: BigNumber;
+};
```
prashantasdeveloper, line 34364: We never add anything here. We should add in `src/types/index.ts`

src/middleware/types.ts
```diff
@@ -1,3 +1,5 @@
+import BigNumber from 'bignumber.js';
+
```
prashantasdeveloper, line 2: This will be removed once you move the below type to other file

src/api/client/Claims.ts
```diff
@@ -483,4 +486,50 @@ export class Claims {
...
+  public async getCustomClaimTypeById(id: BigNumber): Promise<CustomClaimType | null> {
+    const {
+      context: {
+        polymeshApi: {
+          query: { identity },
+        },
+      },
+    } = this;
+
+    const customClaimTypeIdOpt = await identity.customClaims(bigNumberToU32(id, this.context));
```
prashantasdeveloper, line 527: `this.context` can also be destructed above and directly `context` can be used here

prashantasdeveloper (changes requested):

prashantasdeveloper (approved):

PR #1061, opened 2023-10-10
feat: 🎸 register custom claim type

sansan:
### Description

Adds necessary changes to be able to register a CustomClaim
Fixes bug with not being able to revoke claim issued by the same did.

### JIRA Link

https://polymesh.atlassian.net/browse/DA-895 

### Checklist

### DIFF
--- a/src/api/entities/MultiSigProposal/__tests__/index.ts
+++ b/src/api/entities/MultiSigProposal/__tests__/index.ts
@@ -66,14 +66,14 @@ describe('MultiSigProposal class', () => {
         }),
       });
 
+      const mockProposal = dsMockUtils.createMockCall({
+        args: ['ABC'],
+        method: 'reserveTicker',
+        section: 'asset',
+      });
+      mockProposal.toJSON = jest.fn().mockReturnValue({ args: ['ABC'] });
       dsMockUtils.createQueryMock('multiSig', 'proposals', {
-        returnValue: createMockOption(
-          dsMockUtils.createMockCall({
-            args: ['ABC'],
-            method: 'reserveTicker',
-            section: 'asset',
-          })
-        ),
+        returnValue: createMockOption(mockProposal),
       });
 
       let result = await proposal.details();
--- a/src/api/entities/MultiSigProposal/index.ts
+++ b/src/api/entities/MultiSigProposal/index.ts
@@ -71,21 +71,22 @@ export class MultiSigProposal extends Entity<UniqueIdentifiers, HumanReadable> {
         expiry: rawExpiry,
         autoClose: rawAutoClose,
       },
-      proposal,
+      proposalOpt,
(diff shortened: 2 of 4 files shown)


src/utils/internal.ts
```diff
@@ -262,6 +269,13 @@ export function createClaim(
         id: cddId as string,
       };
     }
+    case ClaimType.Custom: {
+      return {
+        type,
+        customClaimTypeId: customClaimTypeId as BigNumber,
```
polymath-eric, line 275: It seems like non null assertion works on the type and looks more natural to me. OK to leave if you prefer `as BigNumber` or if we need it for some reason though
```suggestion
        customClaimTypeId: customClaimTypeId!,
```

Also, should this throw an internal error if `customClaimTypeId` is falsy here? TS isn't going to stop this being undefined.

src/utils/internal.ts
```diff
@@ -241,7 +247,8 @@ export function createClaim(
   claimType: string,
   jurisdiction: Falsyable<string>,
   middlewareScope: Falsyable<MiddlewareScope>,
-  cddId: Falsyable<string>
+  cddId: Falsyable<string>,
+  customClaimTypeId: Falsyable<BigNumber>
```
polymath-eric, line 251: Seems easy to forget to update the last arg of an object when it can be `..., null, null)`. To me the test case using this makes it seem easy to forget to pass in a `customClaimTypeId`. I think a check that throws could help with this API.

Could be nice to have this function accept an object so undefined things don't need to be passed, but I think its OK as is, just worried we'll make a custom claim with `id: undefined` by mistake someday.

polymath-eric (commented):
overall lgtm

src/middleware/enums.ts
```diff
@@ -1,37 +1,37 @@
...
-  AuthTypeEnum: 'PublicEnumCf84Da6Fd1',
-  ClaimTypeEnum: 'PublicEnumEc3243D777',
+  SettlementResultEnum: 'SettlementResultEnum',
+  StatOpTypeEnum: 'StatOpTypeEnum',
+  CallIdEnum: 'CallIdEnum',
+  BalanceTypeEnum: 'BalanceTypeEnum',
+  ProposalStateEnum: 'ProposalStateEnum',
+  ModuleIdEnum: 'ModuleIdEnum',
+  EventIdEnum: 'EventIdEnum',
+  StoStatus: 'StoStatus',
```
prashantasdeveloper, line 29: We can get rid of this now

prashantasdeveloper (changes requested):

prashantasdeveloper (changes requested):
You need to remove the code for enum map generation from `scripts/processMiddlewareTypes.js` as well.

prashantasdeveloper (approved):

PR #1075, opened 2023-10-23
feat: 🎸 add `Identity.getHeldNfts` method

polymath-eric:
### Description

add method to get historically held NFTs. `AssetPermissions.enabledAt` input now accepts `NftCollection`

### Breaking Changes

None

### JIRA Link

[DA-905](https://polymesh.atlassian.net/browse/DA-905)

### Checklist

### DIFF
--- a/src/api/entities/Identity/AssetPermissions.ts
+++ b/src/api/entities/Identity/AssetPermissions.ts
@@ -21,6 +21,7 @@ import {
   ErrorCode,
   EventIdentifier,
   ModuleName,
+  NftCollection,
   PermissionType,
   ProcedureMethod,
   ResultSet,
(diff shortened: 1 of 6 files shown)


prashantasdeveloper (approved):
I'm approving this but we need to update the `MINIMUM_SQ_VERSION` to the released SQ version here. One thing that I worry is that we won't be able to notify users that they need to resync their SQ and get the NFT data in. Maybe we should hold off merging this so that atleast our PRI instances have been updated which would take sometime I guess.

prashantasdeveloper:
We can merge this and worry later to get the min version field updated. Anyhow there won't be large chunk of SDK users using this I guess

PR #1097, opened 2023-11-28
feat: 🎸 add toSignablePayload() to procedures

polymath-eric:
### Description

add toSignablePayload method to procedure, this returns a JSON object that can be signed by an offline signer. Adds a submitTransaction method to the network namespace to submit the transaction with its signature

### Breaking Changes

BREAKING CHANGE: 🧨 procedure will not throw if signingAccount isn't present in the signing manager upon preparation. The check is now performed during `.run()`

### JIRA Link

✅ Closes: [DA-936](https://polymesh.atlassian.net/browse/DA-936)

### Checklist


polymath-eric:
made this a draft for two issues:

- immortal transactions are "bad signature"
- a signing manager is required, even if it has no addresses. SDK should be able to make payloads with out one.

prashantasdeveloper (approved):

PR #1099, opened 2023-12-05
feat: 🎸 Add procedures to clear and remove a MetadataEntry

prashantasdeveloper:
### Description

* Add procedure to clear an Asset Metadata value
* Add procedure to remove any Local Asset Metadata
* Also, adds a method `isModifiable` to MetadataEntry to return whether any metadata entry can be modified

### Breaking Changes

NA

### JIRA Link

DA-885, DA-950, DA-951

### Checklist

### DIFF
--- a/src/api/entities/MetadataEntry/index.ts
+++ b/src/api/entities/MetadataEntry/index.ts
@@ -1,7 +1,8 @@
+import { Bytes, Option } from '@polkadot/types-codec';
 import BigNumber from 'bignumber.js';
 
-import { BaseAsset, Context, Entity, setMetadata } from '~/internal';
-import { ProcedureMethod, SetMetadataParams } from '~/types';
+import { BaseAsset, clearMetadata, Context, Entity, setMetadata } from '~/internal';
+import { NoArgsProcedureMethod, ProcedureMethod, SetMetadataParams } from '~/types';
 import {
   bigNumberToU64,
   bytesToString,
@@ -71,6 +72,10 @@ export class MetadataEntry extends Entity<UniqueIdentifiers, HumanReadable> {
       { getProcedureAndArgs: args => [setMetadata, { ...args, metadataEntry: this }] },
       context
     );
+    this.clear = createProcedureMethod(
+      { getProcedureAndArgs: () => [clearMetadata, { metadataEntry: this }], voidArgs: true },
+      context
+    );
   }
 
   /**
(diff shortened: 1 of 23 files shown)


src/api/procedures/utils.ts
```diff
@@ -710,3 +712,51 @@ export async function addManualFees(
 
   return fees.reduce((prev, { fees: nextFees }) => prev.plus(nextFees), currentFee);
 }
+
+/**
+ * Checks that Metadata exists and that it's value is not locked
+ * @hidden
+ */
+export async function assertMetadataValueIsModifiable(metadataEntry: MetadataEntry): Promise<void> {
```
polymath-eric, line 720: Could this be a method on the MetadataEntry it self?

I think it could be exposed as `isModifiable` and then for the procedure throw the error in the procedure if false..

That way on a UI the same logic can be used to disable a button and for stopping to procedure

src/api/entities/MetadataEntry/index.ts
```diff
@@ -80,6 +85,15 @@ export class MetadataEntry extends Entity<UniqueIdentifiers, HumanReadable> {
...
   public set: ProcedureMethod<SetMetadataParams, MetadataEntry>;
 
+  /**
+   * Removes the asset metadata value
+   *
+   * @throws
+   *   - if the Metadata doesn't exists
+   *   - if the Metadata value is locked
+   */
+  public clear: NoArgsProcedureMethod<void>;
```
polymath-eric, line 95: What about adding something like:

```suggestion
  public clear: NoArgsProcedureMethod<void>;
  
  public isModifiable(): Promise<{canModify: boolean; reason?: PolymeshError}> {
    // assertIsModifiable logic here
  }
```

src/api/procedures/clearMetadata.ts
```diff
@@ -0,0 +1,77 @@
...
+      type,
+      asset: { ticker },
+    },
+    metadataEntry,
+  } = params;
+
+  const rawTicker = stringToTicker(ticker, context);
+  const rawMetadataKey = metadataToMeshMetadataKey(type, id, context);
+
+  await assertMetadataValueIsModifiable(metadataEntry);
```
polymath-eric, line 40:
```suggestion
  const {canModify; reason} = await metadataEntry.isModifable()
  if(!canModify)  throw reason
```

polymath-eric (commented):
I am thinking it would be good to expose a "can modify" to the metadata so a UI can disable a button easily with the same logic. Otherwise its hard to tell if the can be changed until modify request is about to be submitted.

PR #1111, opened 2024-01-05
feat: 🎸 Add method to get owner of a NFT

prashantasdeveloper:
### Description

* Updates the polkadot dependencies with chain 6.1.1
* Adds a new method `getOwner` to Nft to get the holder of the NFT

### Breaking Changes

NA

### JIRA Link

DA-997

### Checklist

### DIFF
--- a/src/api/entities/Asset/NonFungible/Nft.ts
+++ b/src/api/entities/Asset/NonFungible/Nft.ts
@@ -1,8 +1,9 @@
 import BigNumber from 'bignumber.js';
 
-import { Context, Entity, NftCollection, redeemNft } from '~/internal';
+import { Context, Entity, NftCollection, PolymeshError, redeemNft } from '~/internal';
 import {
   DefaultPortfolio,
+  ErrorCode,
   NftMetadata,
   NumberedPortfolio,
   OptionalArgsProcedureMethod,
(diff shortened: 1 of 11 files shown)


VictorVicente (approved):

src/api/entities/Asset/NonFungible/Nft.ts
```diff
@@ -198,6 +208,71 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
     return null;
   }
 
+  /**
+   * Get owner of the NFT
+   *
+   * @note This method returns `null` if there is no existing holder for the token. This may happen even if the token has been redeemed/burned
+   */
+  public async getOwner(): Promise<DefaultPortfolio | NumberedPortfolio | null> {
```
polymath-eric, line 216: nft.exists should use this method and return false for redeemed NFTs

src/api/entities/Asset/NonFungible/Nft.ts
```diff
@@ -198,6 +208,71 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
...
+      id,
+      context: {
+        polymeshApi: {
+          query: { portfolio },
+        },
+      },
+      context,
+    } = this;
+
+    const owner = await this.getOwner();
```
polymath-eric, line 259: I think this check should be encapsulated in the `exists` method, and here just check that method

polymath-eric (changes requested):
should use `owner` method in `exists` check, then in `isLocked` we can use `exists` for the DataUnavailable error check

src/api/entities/Asset/NonFungible/Nft.ts
prashantasdeveloper: We need the owner value to check locked status

polymath-eric (approved):

PR #1112, opened 2024-01-05
fix: 🐛 Correct the logic of `nft.exists`

prashantasdeveloper:
### Description

This adds in condition to correctly return false when checking existence for NFT id 0

### Breaking Changes

NA

### JIRA Link

DA-998

### Checklist

### DIFF
--- a/src/api/entities/Asset/NonFungible/Nft.ts
+++ b/src/api/entities/Asset/NonFungible/Nft.ts
@@ -116,6 +116,11 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
       collection,
       id,
     } = this;
+
+    if (id.lte(new BigNumber(0))) {
+      return false;
+    }
+
     const collectionId = await collection.getCollectionId();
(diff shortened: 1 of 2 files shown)


VictorVicente (approved):

polymath-eric (approved):

src/api/entities/Asset/NonFungible/Nft.ts
```diff
@@ -116,6 +116,11 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
       collection,
       id,
     } = this;
+
+    if (id.lte(new BigNumber(0))) {
+      return false;
+    }
+
```
polymath-eric, line 123: Now that we have owner storage this method should also check if the NFT has an onwer and return false if there is no owner, since that represents a redeemed/burned NFT

polymath-eric: I left a comment on your other PR since that has the owner method implemented

PR #1113, opened 2024-01-09
feat: getTransactionHistory for all Portfolios

sansan:
### Description

adds a method to fetch transaction history by calling `identity.portfolios.getTransactionHistory`

### JIRA Link

https://polymesh.atlassian.net/browse/DA-626

### Checklist

### DIFF
--- a/src/api/entities/Identity/Portfolios.ts
+++ b/src/api/entities/Identity/Portfolios.ts
@@ -1,33 +1,29 @@
 import BigNumber from 'bignumber.js';
 
 import {
-  Account,
   Context,
   DefaultPortfolio,
   deletePortfolio,
-  FungibleAsset,
   Identity,
   Namespace,
   NumberedPortfolio,
   PolymeshError,
 } from '~/internal';
 import { portfoliosMovementsQuery, settlementsForAllPortfoliosQuery } from '~/middleware/queries';
-import { Query, SettlementResultEnum } from '~/middleware/types';
+import { Query } from '~/middleware/types';
 import {
   ErrorCode,
   HistoricSettlement,
   PaginationOptions,
   ProcedureMethod,
   ResultSet,
-  SettlementDirectionEnum,
 } from '~/types';
 import { Ensured } from '~/types/utils';
 import {
   addressToKey,
   identityIdToString,
-  keyToAddress,
-  middlewarePortfolioToPortfolio,
   stringToIdentityId,
+  toHistoricalSettlements,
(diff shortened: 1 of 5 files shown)


src/api/entities/Identity/Portfolios.ts
```diff
@@ -157,4 +176,118 @@ export class Portfolios extends Namespace<Identity> {
...
+              from: middlewarePortfolioToPortfolio(from!, context),
+              to: middlewarePortfolioToPortfolio(to!, context),
+            },
+          ],
+        });
+      }
+    );
+    /* eslint-enable @typescript-eslint/no-non-null-assertion */
+
+    return data;
```
polymath-eric, line 291: Maybe sort by block number?

src/middleware/typesV1.ts
```diff
@@ -12,4 +12,5 @@ export enum SettlementDirectionEnum {
   None = 'None',
   Incoming = 'Incoming',
   Outgoing = 'Outgoing',
+  Internal = 'Internal',
```
polymath-eric, line 15: Can we use "None" instead? Seems like it would fit to me for a self transfer

polymath-eric (approved):
overall lgtm

Try to get the duplicate line check on sonar cloud to go away, its different than a normal smell and hard to ignore. Ideally refactor into a helper, but even a variable rename of line swap can fix it.

src/utils/conversion.ts
```diff
@@ -4712,3 +4716,95 @@ export function toCustomClaimTypeWithIdentity(
...
+            from: middlewarePortfolioToPortfolio(from!, context),
+            to: middlewarePortfolioToPortfolio(to!, context),
+          },
+        ],
+      });
+    }
+  );
+  /* eslint-enable @typescript-eslint/no-non-null-assertion */
+
+  return data.sort((a, b) => a.blockNumber.minus(b.blockNumber).toNumber());
```
polymath-eric, line 4809: (nit) I was thinking most recent first, since our returned order will be the default order for consumers.

Sure people can sort themselves, but I think block number DESC is most common for a default.

```ts
[1, 2, 3].sort((a, b) => a - b) // 1,2,3
[1, 2, 3].sort((a, b) => b - a) // 3,2,1
```

FWIW I think ASC order makes more sense then clumping settlements and portfolio movements together though like it was before.

polymath-eric (approved):
lgtm

prashantasdeveloper (approved):

PR #1127, opened 2024-01-26
feat: 🎸 add `addTransaction` method to confidential venue

polymath-eric:
### Description

allows venue owners to create confidential transactions with `addTransaction` and `addTransctions` methods

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-981](https://polymesh.atlassian.net/browse/DA-981)

### Checklist

[DA-981]: https://polymesh.atlassian.net/browse/DA-981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/procedures/addConfidentialTransaction.ts
+++ b/src/api/procedures/addConfidentialTransaction.ts
@@ -1,0 +1,317 @@
+import { u64 } from '@polkadot/types';
+import {
+  PalletConfidentialAssetTransactionLeg,
+  PolymeshPrimitivesMemo,
+} from '@polkadot/types/lookup';
+import { ISubmittableResult } from '@polkadot/types/types';
+import BigNumber from 'bignumber.js';
+
+import {
+  assertConfidentialAccountExists,
+  assertConfidentialAssetExists,
+  assertConfidentialVenueExists,
+  assertIdentityExists,
+} from '~/api/procedures/utils';
+import {
+  ConfidentialTransaction,
+  Context,
+  DefaultPortfolio,
+  NumberedPortfolio,
+  PolymeshError,
+  Procedure,
+} from '~/internal';
+import {
+  AddConfidentialTransactionParams,
+  AddConfidentialTransactionsParams,
+  ConfidentialAssetTx,
+  ErrorCode,
+  RoleType,
+} from '~/types';
+import { BatchTransactionSpec, ProcedureAuthorization } from '~/types/internal';
+import { MAX_LEGS_LENGTH } from '~/utils/constants';
+import {
+  auditorsToBtreeSet,
+  bigNumberToU64,
+  confidentialAccountToMeshPublicKey,
+  confidentialAssetsToBtreeSet,
+  confidentialLegToMeshLeg,
+  identitiesToBtreeSet,
+  stringToMemo,
+  u64ToBigNumber,
+} from '~/utils/conversion';
+import {
+  asConfidentialAccount,
+  asConfidentialAsset,
+  asIdentity,
+  assembleBatchTransactions,
+  filterEventRecords,
+  optionize,
+} from '~/utils/internal';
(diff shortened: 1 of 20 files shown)


src/api/procedures/addConfidentialTransaction.ts
```diff
@@ -0,0 +1,317 @@
...
+      polymeshApi: {
+        tx: { confidentialAsset },
+      },
+    },
+    context,
+  } = this;
+  const { transactions, venueId } = args;
+
+  await assertConfidentialVenueExists(venueId, context);
+
```
prashantasdeveloper, line 224: We also need to add check if venue filtering allows that particular venue to add transaction for a given asset id

polymath-eric: added a check on line 164

src/api/procedures/addConfidentialTransaction.ts
```diff
@@ -0,0 +1,320 @@
...
+ * @hidden
+ */
+export const createConfidentialTransactionResolver =
+  (context: Context) =>
+  (receipt: ISubmittableResult): ConfidentialTransaction[] => {
+    const events = filterEventRecords(receipt, 'confidentialAsset', 'TransactionCreated');
+
+    const result = events.map(
+      ({ data }) =>
+        new ConfidentialTransaction({ id: confidentialTransactionIdToBigNumber(data[2]) }, context)
```
polymath-eric, line 85: @prashantasdeveloper FYI, I had a bug here on the original PR. I did `data[1]`, which is the venue ID.

src/api/procedures/utils.ts
```diff
@@ -216,6 +216,41 @@ export async function assertConfidentialAccountExists(account: ConfidentialAccou
...
 
+/**
+ * @hidden
+ * */
+export async function assertConfidentialAssetsEnabledForVenue(
+  venueId: BigNumber,
+  assets: ConfidentialAsset[]
+): Promise<void> {
+  const filterDetails = await Promise.all(
+    assets.map(async asset => {
```
prashantasdeveloper, line 227: In the past, I have encountered a few scenarios where .map with async doesn't work as expected. Either we map the elements to a Promise array and then await on all or use a for-of loop instead of map and async. Not sure if this is working as expected.

prashantasdeveloper (approved):
LGTM. Left a doubt around using Array.map with async. Approving this for now

src/api/procedures/utils.ts
polymath-eric: I agree its a bit of a tricky construct. 

> Either we map the elements to a Promise array and then await on all 

This is what is happening here. Adding `async` to the inner map function means the inferred return type gets wrapped in `Promise`. The `.map` executes immediately with no blocking and becomes Promise<{asset, details}>[].

`await Promise.all(` on the above line is key here.


`P.map` is from Bluebird, which they themselves recommend using native promise support where possible.

> If there is a feature that keeps you using bluebird. Please let us know so we can try and upstream it :)
Currently - it is only recommended to use Bluebird if you need to support old browsers or EoL Node.js or as an intermediate step to use warnings/monitoring to find bugs. 
https://github.com/petkaantonov/bluebird/?tab=readme-ov-file#%EF%B8%8Fnote%EF%B8%8F

Bluebird isn't deprecated exactly, so I don't think its worth removing at this point, but I think we should avoid using it, since it largely exists due to deficiencies with older versions of JS.

polymath-eric:
@prashantasdeveloper I agree we need to be careful with `async` + map, but as long as its combined with `Promise.all` or `Promise.allSettled` it should be OK from my experience.

PR #1137, opened 2024-02-08
feat: 🎸 add asset required mediators functions

polymath-eric:
### Description

add create, get, remove functions for asset required mediators

### Breaking Changes

None

### JIRA Link

✅ Closes: [DA-1019](https://polymesh.atlassian.net/browse/DA-1019)

### Checklist

[DA-1019]: https://polymesh.atlassian.net/browse/DA-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


prashantasdeveloper (approved):
LGTM, but IMO we could have merged both the procedures. From what I can see on the TxTags + the condition to check if already present/removed is different which could be handled by a boolean (say operation - add/remove). 

I'm approving this knowing that this is needed sooner than later. Maybe something we can address later

PR #1149, opened 2024-02-21
feat: 🎸 getTransactionHistory for ConfidentialAccount

sansan:
### Description

adds method getTransactionHistory for ConfidentialAccount

### JIRA Link

https://polymesh.atlassian.net/browse/DA-977

### Checklist


src/api/entities/confidential/ConfidentialAccount/index.ts
```diff
@@ -313,4 +320,57 @@ export class ConfidentialAccount extends Entity<UniqueIdentifiers, string> {
...
+    );
+
+    const data: ConfidentialAssetHistoryEntry[] = nodes.map(
+      ({ id, assetId, transactionId, amount, eventId }) => {
+        return {
+          id,
+          asset: new ConfidentialAsset({ id: assetId }, context),
+          ...(transactionId && {
+            transaction: new ConfidentialTransaction({ id: new BigNumber(transactionId) }, context),
+          }),
```
polymath-eric, line 361: what is this doing? If transactionId is falsy wouldn't this cause a crash? I am assuming this is a work around from where we didn't add in the not null constraint in SQ.

maybe filter for where transactionId is not null instead? Or at least leave a comment explaining how we know this is safe.

polymath-eric (approved):

PR #1187, opened 2024-04-05
fix: 🐛 Remove future block constraint on `endAfterBlock`

prashantasdeveloper:
### Description

This will allow to set an older block as `endAfterBlock` while creating an instruction, eventually allowing the manual execution at any time

### Breaking Changes

NA

### JIRA Link

DA-1129

### Checklist


prashantasdeveloper:
> Is it ok to remove it completely or it should just accepts 0 as a value?

It works for any past block, so I removed it

PR #1203, opened 2024-04-18
feat: 🎸 get historical multisig proposals

sansan:
### Description

- adds method on MultiSig to get HistoricalMultiSigProposal(s) for that MultiSig

### JIRA Link

https://polymesh.atlassian.net/browse/DA-129

### Checklist


src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+enum MultiSigStatusEnum {
+  Pending = 'Pending',
+  Success = 'Success',
+  Failed = 'Failed',
+}
+
+/**
+ * Represents historical MultiSigProposal that no longer exists on chain
```
polymath-eric, line 47:
```suggestion
 * Represents a historical MultiSigProposal
```

(nit) I think these are made in SQ when the propsal is added, so it could exist on chain still. Maybe mention this will persist after execution instead.

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
+import BigNumber from 'bignumber.js';
+
+import { Context, Entity } from '~/internal';
+import { multiSigProposalVotesQuery } from '~/middleware/queries';
+import { Query } from '~/middleware/types';
+import { MultiSigProposalVote, SignerType } from '~/types';
+import { Ensured } from '~/types/utils';
+import { middlewareEventDetailsToEventIdentifier, signerValueToSigner } from '~/utils/conversion';
+
+export type MiddlewareMultiSigProposal = {
```
polymath-eric, line 10: I think exported types are supposed to go in the types file for project consistency: src/api/entities/Account/MultiSig/types.ts

polymath-eric, line 10: (nit)
```suggestion
export interface MiddlewareMultiSigProposal {
```

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+  approvalCount: number;
+  rejectionCount: number;
+  creatorDid?: string;
+  status: string;
+  createdBlockId: string;
+  updatedBlockId: string;
+  datetime: string;
+}
+
+enum MultiSigStatusEnum {
```
polymath-eric, line 40: I think this should be in `types.ts` as well

polymath-eric (commented):
lgtm, left some small comments

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+import { MultiSigProposalVote, SignerType } from '~/types';
+import { Ensured } from '~/types/utils';
+import { middlewareEventDetailsToEventIdentifier, signerValueToSigner } from '~/utils/conversion';
+
+export type MiddlewareMultiSigProposal = {
+  id: string;
+  proposalId: number;
+  multisigId: string;
+  approvalCount: number;
+  rejectionCount: number;
```
prashantasdeveloper, line 15: If this is to be exported, then return type should be BigNumber instead of number

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+  Failed = 'Failed',
+}
+
+/**
+ * Represents historical MultiSigProposal that no longer exists on chain
+ */
+export class HistoricalMultiSigProposal extends Entity<
+  MiddlewareMultiSigProposalUniqueIdentifiers,
+  HumanReadable
+> {
```
prashantasdeveloper, line 52: IMO this should not be a Entity. A simple interface would do the trick

prashantasdeveloper: Anyhow other entity related methods would not be used anywhere

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Get the MultiSigProposalVotes associated with this MultiSigProposal
+   */
+  public async getVotes(): Promise<MultiSigProposalVote[]> {
+    const {
+      context: { queryMiddleware },
+      id: proposalId,
+      context,
+    } = this;
+
```
prashantasdeveloper, line 134: This can be added as a method named `getVotes` in MultiSigProposal entity

src/middleware/queries.ts
```diff
@@ -1690,3 +1690,72 @@ export function nftCollectionHolders(
...
 }
+
+type MultiSigProposalQueryParameters = {
+  multisigId: string;
+};
+
+/**
+ *  @hidden
+ */
+export function createmultiSigProposalsQueryFilters(): {
```
prashantasdeveloper, line 1701:
```suggestion
export function createMultiSigProposalsQueryFilters(): {
```

src/middleware/queries.ts
```diff
@@ -1690,3 +1690,72 @@ export function nftCollectionHolders(
...
+      multisigId,
+      approvalCount,
+      status,
+      creator {
+        did,
+      }
+      rejectionCount,
+      createdBlockId,
+      updatedBlockId,
+      datetime,
```
prashantasdeveloper, line 1750: nit -> the query is not formatted. Use prettier extension with VS Code to get it formatted automatically on save

src/api/entities/HistoricalMultiSigProposal.ts
prashantasdeveloper: I double checked and this is already present as method named `votes` in `MultiSigProposal` entity

src/api/entities/HistoricalMultiSigProposal.ts
```diff
@@ -0,0 +1,177 @@
...
+import { Context, Entity } from '~/internal';
+import { multiSigProposalVotesQuery } from '~/middleware/queries';
+import { Query } from '~/middleware/types';
+import { MultiSigProposalVote, SignerType } from '~/types';
+import { Ensured } from '~/types/utils';
+import { middlewareEventDetailsToEventIdentifier, signerValueToSigner } from '~/utils/conversion';
+
+export type MiddlewareMultiSigProposal = {
+  id: string;
```
prashantasdeveloper, line 11:
```suggestion
```
This is not needed to be exposed to end user. ID from subquery is something we would use to just query stuff. It is generally a combination of a few values which would confuse the user to about what it is

src/middleware/queries.ts
```diff
@@ -1690,3 +1690,72 @@ export function nftCollectionHolders(
...
+ */
+export function multiSigProposalsQuery(
+  multisigId: string,
+  size?: BigNumber,
+  start?: BigNumber
+): QueryOptions<PaginatedQueryArgs<MultiSigProposalQueryParameters>> {
+  const { args, filter } = createmultiSigProposalsQueryFilters();
+
+  const query = gql`
+  query MultiSigProposalsQuery
```
prashantasdeveloper, line 1730: There is no need for separate query. Already we have `multiSigProposalQuery` in line 1525. You can just extend that to suffice our use case

prashantasdeveloper (changes requested):
The new entity is not needed. Also, queries were already written. You can just extend the existing one to give all the data that you need. `getVotes` method is already present as `votes` in `MultiSigProposal`.

src/api/entities/HistoricalMultiSigProposal.ts
prashantasdeveloper: I still think that is not needed. We would be using the same MultiSigProposal and HistoricalMultiSigProposal to do the exact same things. Also, for consistency, we always return a interface while getting the historical info

src/api/entities/Account/MultiSig/index.ts
```diff
@@ -113,6 +123,44 @@ export class MultiSig extends Account {
     );
   }
 
+  /**
+   * Return all { @link api/entities/HistoricalMultiSigProposal!HistoricalMultiSigProposal } for this MultiSig Account
+   */
+  public async getHistoricalProposals(opts: {
+    size?: BigNumber;
+    start?: BigNumber;
+  }): Promise<ResultSet<HistoricalMultiSigProposal>> {
```
prashantasdeveloper, line 132: IMO, we still don't need a `HistoricalMultiSigProposal` entity. They way you are doing is just creating a replica of `MultiSigProposal` as `HistoricalMultiSigProposal` and adding wrapper methods to call `MultiSigProposal`. Instead you can change the return type directly here to `Promise<ResultSet<MultiSigProposal>>` or like we do with other entities, call the SQ and return a detailed information in a interface call `HistoricalMultiSigProposal` which will have all the details of multisig fetched beforehand

prashantasdeveloper (changes requested):

prashantasdeveloper (approved):

PR #1205, opened 2024-04-23
feat: 🎸 Refactor types into module specific files

prashantasdeveloper:
### Description

Some types needed to be re-exported from the private SDK, which required re-arranging them in SDK. This PR moves the constants and types in separate files making them easier to be exported.

### Breaking Changes

NA

### JIRA Link

DA-1151

### Checklist


src/types/index.ts
```diff
@@ -1,43 +1,10 @@
...
-  Instruction,
-  KnownPermissionGroup,
-  Nft,
-  NftCollection,
-  NumberedPortfolio,
-  Offering,
-  PolymeshTransaction,
-  PolymeshTransactionBatch,
-} from '~/internal';
-import { Modify } from '~/types/utils';
```
polymath-eric, line 4: Should this get `export * from "module/types." files?

If someone was importing types it would probably be from this file, so moving them without a reexport could be a breaking change.

src/api/entities/entityTypes.ts
```diff
@@ -0,0 +1,774 @@
+import BigNumber from 'bignumber.js';
```
polymath-eric, line 1: thought about putting these below the rexports in `types` file? Seems a bit redundant to have two type files beside each other.

polymath-eric (approved):

src/types/index.ts
polymath-eric: Already being exported transitively.

PR #1227, opened 2024-05-17
feat: 🎸 adjust permissioned account type for account

sansan:
### Description

- adjusts `PermissionedAccount` so that 
 account is either Account or MultiSig type

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1163

### Checklist

### DIFF
--- a/src/utils/internal.ts
+++ b/src/utils/internal.ts
@@ -1810,7 +1810,7 @@ export function assertStatIsSet(
  * @param args.identity optional. If passed, Accounts that are not part of the given Identity will be filtered out
  */
 export async function getSecondaryAccountPermissions(
-  args: { accounts: (Account | MultiSig)[]; identity?: Identity },
+  args: { accounts: Account[]; identity?: Identity },
   context: Context,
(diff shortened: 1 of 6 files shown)


src/utils/internal.ts
```diff
@@ -1809,19 +1810,19 @@ export function assertStatIsSet(
  * @param args.identity optional. If passed, Accounts that are not part of the given Identity will be filtered out
  */
 export async function getSecondaryAccountPermissions(
-  args: { accounts: Account[]; identity?: Identity },
+  args: { accounts: (Account | MultiSig)[]; identity?: Identity },
```
polymath-eric, line 1813: The input type shouldn't need to be a union.

MultiSig extends Account so it should satisfy the smaller `Account` interface.

polymath-eric (commented):
lgtm, but not sure why the tests are failing. maybe it needs to have an entity mock import at the top?

polymath-eric (approved):

polymath-eric:
rebased so we can ff into alpha

PR #1240, opened 2024-05-29
feat: 🎸 Use `currentAssetMetadataLocalKey` to get next local ID

prashantasdeveloper:
### Description

Replaces the deprecated usage of `assetMetadataNextLocalKey` with `currentAssetMetadataLocalKey`. This is supported with 6.3 changes

### Breaking Changes

NA

### JIRA Link

DA-1200

### Checklist

### DIFF
--- a/src/api/entities/Asset/Base/Metadata/index.ts
+++ b/src/api/entities/Asset/Base/Metadata/index.ts
@@ -155,15 +155,19 @@ export class Metadata extends Namespace<BaseAsset> {
       context: {
         polymeshApi: {
           query: {
-            asset: { assetMetadataNextLocalKey },
+            asset: { currentAssetMetadataLocalKey },
           },
         },
       },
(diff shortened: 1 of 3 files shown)


polymath-eric (commented):
why directly into master? Shouldn't this go into the alpha release?

this change shouldn't be in master until 6.3 is deployed to mainnet so nothing breaks for people. The supported version should get update to 6.3 as well, since 6.2 is no longer supported with this change.

prashantasdeveloper:
> why directly into master? Shouldn't this go into the alpha release?
> 
> this change shouldn't be in master until 6.3 is deployed to mainnet so nothing breaks for people. The supported version should get update to 6.3 as well, since 6.2 is no longer supported with this change.

My bad, didn't change the branch while creating the PR

VictorVicente (approved):

PR #1241, opened 2024-05-29
chore: 🤖 add support for fast-forward github action

prashantasdeveloper:
### Description

This adds support for fast forwarding with PR comments

### Breaking Changes

NA

### JIRA Link

DA-1215

### Checklist

### DIFF
--- a/.github/workflows/fast-forward.yml
+++ b/.github/workflows/fast-forward.yml
@@ -1,0 +1,27 @@
+name: fast-forward
+on:
+  issue_comment:
+    types: [created, edited]
+jobs:
+  fast-forward:
+    # Only run if the comment contains the /fast-forward command.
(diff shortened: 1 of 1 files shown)


polymath-eric (commented):
shouldn't this go into alpha?

I guess its OK now since both are synced atm, but if this gets merged into master, then another alpha PR merge could mess up the linear history requirement.

I know we don't have it enabled on this repo so maybe its OK?

prashantasdeveloper:
> shouldn't this go into alpha?
> 
> I guess its OK now since both are synced atm, but if this gets merged into master, then another alpha PR merge could mess up the linear history requirement.
> 
> I know we don't have it enabled on this repo so maybe its OK?

For actions to work in alpha, the action should be present in master first. Then only the workflow gets triggered in other branches.

VictorVicente (approved):

PR #1252, opened 2024-06-20
fix: 🐛 get portfolio transactions error

sansan:
### Description

- fixes error with graphql query that results in error when getting portfolio transactions 

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1237

### Checklist


prashantasdeveloper (commented):
I'm not sure what the problem was here. For a `fromId` there should be a `from { number identityId }`. Which case was this failing?

prashantasdeveloper (approved):

PR #1254, opened 2024-07-04
feat: 🎸 allow MultiSig signers to submit proposals

polymath-eric:
### Description

add `runAsProposal` method to procedures. This is to be used when the signingAccount is a MultiSig signer. The transaction will be wrapped as a multiSig proposal, which will be executed when enough signers approve

### Breaking Changes

None

### JIRA Link

[DA-1249](https://polymesh.atlassian.net/browse/DA-1249)

### Checklist

[DA-1249]: https://polymesh.atlassian.net/browse/DA-1249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/base/PolymeshTransactionBase.ts
+++ b/src/base/PolymeshTransactionBase.ts
@@ -10,13 +10,15 @@ import {
   handleTransactionSubmissionError,
   pollForTransactionFinalization,
 } from '~/base/utils';
-import { Context, Identity, PolymeshError } from '~/internal';
+import { Context, Identity, MultiSigProposal, PolymeshError } from '~/internal';
 import { latestBlockQuery } from '~/middleware/queries';
 import { Query } from '~/middleware/types';
 import {
   ErrorCode,
   GenericPolymeshTransaction,
   MortalityProcedureOpt,
+  MultiSig,
+  MultiSigProcedureOpt,
   PayingAccount,
   PayingAccountFees,
   PayingAccountType,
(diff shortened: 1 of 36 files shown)


polymath-eric:
This should mostly be good to look at, but there is some work todo around the procedure's `_result`.

I didn't set it for MultiSig since its supposed to be the Transformed type, but really it should be set to the proposal.

Either the getter can throw if its MultiSig, or we introduce a breaking change by widening the type to include `MultiSigProposal`.

src/base/utils.ts
```diff
@@ -281,3 +282,16 @@ export const pollForTransactionFinalization = async (
     events: relatedEvents,
   });
 };
+
+/**
+ * @hidden
+ */
+const multiSigNoWrapTxs = ['multiSig.approveAsKey', 'multiSig.rejectAsKey'] as TxTag[];
```
prashantasdeveloper, line 289:
```suggestion
const multiSigNoWrapTxs = [MultiSigTx.ApproveAsKey, MultiSigTx.RejectAsKey] as TxTag[];
```

src/base/PolymeshTransactionBase.ts
```diff
@@ -436,7 +532,9 @@ export abstract class PolymeshTransactionBase<
    *   chain related factors (like modifications to a specific subsidizer relationship or a chain upgrade)
    */
   public async getTotalFees(): Promise<PayingAccountFees> {
-    const { signingAddress } = this;
+    const { signingAddress: signer } = this;
+
+    const signingAddress = signer;
```
prashantasdeveloper, line 537:
```suggestion
    const { signingAddress } = this;
```

prashantasdeveloper (commented):
Overall LGTM, left couple of nit comments

PR #1258, opened 2024-07-11
fix: 🐛 `multiSig.getHistoricalProposal` errors

polymath-eric:
### Description

fixed GraphQL error when fetching historical proposals. Also add `stack` attribute to PolymeshError

### Checklist

### DIFF
--- a/src/api/entities/Account/MultiSig/index.ts
+++ b/src/api/entities/Account/MultiSig/index.ts
@@ -157,26 +157,22 @@ export class MultiSig extends Account {
   }
 
   /**
-   * Return all { @link api/entities/MultiSigProposal!MultiSigProposal } for this MultiSig Account
+   * Return a set of { @link api/entities/MultiSigProposal!MultiSigProposal | MultiSigProposal } for this MultiSig Account
    *
    * @note uses the middlewareV2
    */
-  public async getHistoricalProposals(opts: {
+  public async getHistoricalProposals(opts?: {
     size?: BigNumber;
     start?: BigNumber;
   }): Promise<ResultSet<MultiSigProposal>> {
-    const {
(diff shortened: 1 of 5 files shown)


src/api/entities/Account/MultiSig/index.ts
```diff
@@ -157,26 +157,22 @@ export class MultiSig extends Account {
...
-    const { size, start } = opts;
+    const { context, address } = this;
+    const { size, start } = opts ?? {};
 
     const {
       data: {
         multiSigProposals: { nodes, totalCount },
       },
-    } = await queryMiddleware<Ensured<Query, 'multiSigProposals'>>(
+    } = await context.queryMiddleware<Ensured<Query, 'multiSigProposals'>>(
```
polymath-eric, line 175: `this` was undefined in `queryMiddleware` when destructuring.

src/base/PolymeshError.ts
```diff
@@ -24,14 +24,19 @@ export class PolymeshError extends Error {
     message,
     code,
     data,
+    stack,
   }: {
     message?: string;
     code: ErrorCode;
     data?: Record<string, unknown>;
+    stack?: Error['stack'];
```
polymath-eric, line 32: > Note: The stack property is de facto implemented by all major JavaScript engines, and [the JavaScript standards committee is looking to standardize it](https://github.com/tc39/proposal-error-stacks). You cannot rely on the precise content of the stack string due to implementation inconsistencies, but you can generally assume it exists and use it for debugging purposes.

~ [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack)

VictorVicente (approved):

PR #1259, opened 2024-07-12
feat: 🎸 Add support for SQ settlements V2

prashantasdeveloper:
### Description

This adds support for new SQ settlement changes while keeping the old SQ support as well

### JIRA Link

DA-1254

### Checklist


src/utils/internal.ts
```diff
@@ -1421,7 +1421,17 @@ export async function assertExpectedSqVersion(context: Context): Promise<void> {
     },
   } = await context.queryMiddleware<Ensured<Query, 'subqueryVersions'>>(latestSqVersionQuery());
 
-  if (!sqVersion || lt(sqVersion.version, MINIMUM_SQ_VERSION)) {
+  return sqVersion?.version || '1.0.0';
```
polymath-eric, line 1424: Maybe we should cache this value with a TTL? 

From the portal it takes me ~500ms to query middleware.  SQ could get updated so rechecking sometimes makes sense. 

Two round trips for a total of 1 second latency is slow but still usable. On the otherhand this is temporary, so might not be worth the complexity.

src/utils/constants.ts
```diff
@@ -174,6 +174,8 @@ export const DEFAULT_CDD_ID = '0x00000000000000000000000000000000000000000000000
  */
 export const MINIMUM_SQ_VERSION = '12.2.0-alpha.2';
 
+export const SETTLEMENTS_V2_SQ_VERSION = '16.0.0-alpha.1';
```
polymath-eric, line 177: Should these versions only reference official releases? I think they do show to the user if they have a bad version.

I get why we'd need `alpha.1` for testing in dev, but it should be OK to make this a plain 16 as long as we coordinate the release timing properly

polymath-eric (commented):
I like having the queries in organized into file. Thanks for refactoring!

src/utils/internal.ts
polymath-eric: I guess its only for the few breaking queries though. I was originally thinking it would be before every query.

OK to leave it, but I figure I would mention it.

src/api/client/__tests__/Network.ts
```diff
@@ -5,7 +5,8 @@ import { when } from 'jest-when';
 import { Network } from '~/api/client/Network';
 import * as baseUtils from '~/base/utils';
 import { Context, PolymeshError, PolymeshTransaction } from '~/internal';
-import { eventsByArgs, extrinsicByHash } from '~/middleware/queries';
+import { eventsByArgs } from '~/middleware/queries/events';
+import { extrinsicByHash } from '~/middleware/queries/extrinsics';
```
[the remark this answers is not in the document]
prashantasdeveloper: Somehow it messed up the working. While running the SDK the queries were not working with the old imports.

src/utils/constants.ts
prashantasdeveloper: But coordinating and updating on the right time can be tricky. It's OK to leave as is with this to be removed soon. And also, this is not getting logged anywhere. Its just for checking, so anyone using middleware alpha version would also get the correct answer to the queries

polymath-eric (approved):

src/utils/constants.ts
polymath-eric, line 177: It is getting logged: https://github.com/PolymeshAssociation/polymesh-sdk/blob/8689f633a8b6807a4a194ff53dafde5f90056ab4/src/utils/internal.ts#L1426

Also, I think we should leave it as `12.2.0-alpha.2` for now, since this has dual version support. It should probably get a TODO to update like the version check calls.

polymath-eric: I thought this was replacing minimum SQ version 😅

PR #1260, opened 2024-07-15
fix: 🐛 throw catchable connect errors

polymath-eric:
### Description

`Polymesh.connect` throws catchable connect errors. Previously it would crash on node, or clutter the console with `Abnormal closure` in the browser

### JIRA Link

✅ Closes: [DA-1251](https://polymesh.atlassian.net/browse/DA-1251)

### Checklist

[DA-1251]: https://polymesh.atlassian.net/browse/DA-1251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/api/client/Polymesh.ts
+++ b/src/api/client/Polymesh.ts
@@ -20,9 +20,9 @@ import {
 import { signerToString } from '~/utils/conversion';
 import {
   assertExpectedChainVersion,
-  assertExpectedSqVersion,
   createProcedureMethod,
   extractProtocol,
+  warnUnexpectedSqVersion,
 } from '~/utils/internal';
 
 import { AccountManagement } from './AccountManagement';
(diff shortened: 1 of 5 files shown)


src/api/client/Polymesh.ts
```diff
@@ -151,8 +151,7 @@ export class Polymesh {
 
     const { metadata, noInitWarn, typesBundle } = polkadot ?? {};
 
-    // Defer `await` on any checks to minimize total startup time
```
polymath-eric, line 154: If we make polkadot api with a bad node URL it will keep on spewing out the "abnormal closure" errors.

In theory we could parallelize the middleware check here as well, but since it depends on `Context`, which depends on polkadotApi, there is a circular dependency issue in this instance with our current setup.

src/utils/internal.ts
```diff
@@ -1412,7 +1412,7 @@ function assertExpectedSpecVersion(
  *
  * Checks SQ version compatibility with the SDK
  */
-export async function assertExpectedSqVersion(context: Context): Promise<void> {
```
polymath-eric, line 1415: This is only warning now, so I changed the name. `assert` makes it feel like it should throw to me

prashantasdeveloper (approved):

PR #1261, opened 2024-07-18
feat: 🎸 Add off chain authorizations support

prashantasdeveloper:
### Description

Add procedure to add secondary accounts to an Identity. 
Also, adds in method to get off chain authorization nonce for an Identity

### Breaking Changes

NA

### JIRA Link

DA-198

### Checklist


src/api/procedures/types.ts
```diff
@@ -600,6 +600,16 @@ export interface RemoveSecondaryAccountsParams {
   accounts: Account[];
 }
 
+export interface AccountWithSignature {
+  secondaryAccount: PermissionedAccount;
+  authSignature: string;
```
polymath-eric, line 605: How is someone supposed to get an `authSignature`? I think it needs more documentation

src/api/client/AccountManagement.ts
```diff
@@ -122,6 +141,13 @@ export class AccountManagement {
    */
   public removeSecondaryAccounts: ProcedureMethod<RemoveSecondaryAccountsParams, void>;
 
+  /**
+   * Adds a list of secondary Accounts to the signing Identity
+   *
+   * @throws if the signing Account is not the primary Account of the Identity
+   */
+  public addSecondaryAccounts: ProcedureMethod<AddSecondaryAccountsParams, Identity>;
```
polymath-eric, line 149: The comment should mention how to get the auth signatures, and why its better than invite account, since they perform similar functions.

src/api/client/AccountManagement.ts
```diff
@@ -291,4 +317,35 @@ export class AccountManagement {
...
+    target: string | Identity;
+    expiry: Date;
+  }): Promise<`0x${string}`> {
+    const { context } = this;
+
+    const { target, signer, expiry } = args;
+
+    const targetIdentity = asIdentity(target, context);
+
+    const offChainAuthNonce = await targetIdentity.getOffChainAuthorizationNonce();
```
polymath-eric, line 339: If the nonce is on the target, how is this supposed to work with multiple accounts being added in `addSecondaryAccounts`? Shouldn't the user be able to specify the nonce as an option, or is it something like the same nonce is fine since they will get processed in a batch?

polymath-eric (commented):
code looks good to me, but some more documentation about how and why to use the method would help imo

src/api/client/AccountManagement.ts
prashantasdeveloper: same nonce is fine for all

polymath-eric: It should still support a manually specified nonce and default to this value, like we do for transaction options.

I am thinking there could be other auths to sign right after joining an identity and the nonce would let the secondary key do all of its signing at once.

PR #1278, opened 2024-08-23
feat: 🎸 update polkadot types to latest

polymath-eric:
### Description

Build works, but test mocks needs to be updated + left some TODOs in permissions

### Checklist


polymath-eric:
@prashantasdeveloper I put the date in since I meant this to be only for the day. We should use the 7.0 release branch imo

PR #1280, opened 2024-08-29
feat: 🎸 allow optional venues for 7.0 chains

polymath-eric:
### Description

allow no venue id to be passed when creating instructions. None of the assets can have venue filtering enabled

### JIRA Link

[DA-1282](https://polymesh.atlassian.net/browse/DA-1282)

### Checklist


polymath-eric:
There is no guard for v6 here, but maybe better to wait until we have the dual version in alpha at least before merging in new features.

prashantasdeveloper:
IMO we can put in guard for v6 to throw error if no venue is provided and add in docs explaining that

src/api/procedures/__tests__/addInstruction.ts
```diff
@@ -712,6 +712,36 @@ describe('addInstruction procedure', () => {
     ).rejects.toThrow(expectedError);
   });
 
+  it.only('should throw an error if the no venue is specified and an asset has filtering enabled', () => {
```
polymath-eric, line 715:
```suggestion
  it('should throw an error if the no venue is specified and an asset has filtering enabled', () => {
```

prashantasdeveloper (approved):

PR #1286, opened 2024-09-03
feat: 🎸 handle optional venues when processing instruction

polymath-eric:
### Description

Handle optional venues for instruction processing

### Checklist

### DIFF
--- a/src/api/procedures/executeManualInstruction.ts
+++ b/src/api/procedures/executeManualInstruction.ts
@@ -71,11 +71,9 @@ export async function prepareExecuteManualInstruction(
   await assertInstructionValidForManualExecution(instructionDetails, context);
 
   if (!portfolios.length) {
-    const {
-      owner: { did: venueOwner },
-    } = await instructionDetails.venue.details();
+    const details = await instructionDetails.venue?.details();
 
-    if (venueOwner !== signerDid && !offChainParties.has(signerDid)) {
+    if (details?.owner.did !== signerDid && !offChainParties.has(signerDid)) {
       throw new PolymeshError({
         code: ErrorCode.UnmetPrerequisite,
         message: 'The signing identity is not involved in this Instruction',
--- a/src/api/procedures/modifyInstructionAffirmation.ts
+++ b/src/api/procedures/modifyInstructionAffirmation.ts
@@ -122,7 +122,7 @@ const assertReceipts = async (
   const instruction = new Instruction({ id: instructionId }, context);
 
   const { venue } = await instruction.details();
-  const allowedSigners = await venue.getAllowedSigners();
+  const allowedSigners = await venue?.getAllowedSigners();
 
   const invalidReceipts: OffChainAffirmationReceipt[] = [];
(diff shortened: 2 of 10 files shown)


src/api/procedures/executeManualInstruction.ts
```diff
@@ -70,11 +70,9 @@ export async function prepareExecuteManualInstruction(
   await assertInstructionValidForManualExecution(instructionDetails, context);
 
   if (!portfolios.length) {
-    const {
-      owner: { did: venueOwner },
-    } = await instructionDetails.venue.details();
+    const details = await instructionDetails.venue?.details();
 
-    if (venueOwner !== signerDid && !offChainParties.has(signerDid)) {
+    if (details?.owner.did !== signerDid && !offChainParties.has(signerDid)) {
```
prashantasdeveloper, line 75:
```suggestion
    if (!offChainParties.has(signerDid) && (!details || details.owner.did !== signerDid) {
```
If there is no venue associated with the instruction, non-offchain parties should not be allowed in that cases

prashantasdeveloper: I was looking at the core code but wasn't sure about the logic

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -144,7 +144,9 @@ const assertReceipts = async (
 
     const { address: signerAddress } = asAccount(signer, context);
 
-    const isAllowedSigner = allowedSigners.some(({ address }) => signerAddress === address);
+    const isAllowedSigner = allowedSigners
+      ? allowedSigners.some(({ address }) => signerAddress === address)
+      : true;
```
prashantasdeveloper, line 149:
```suggestion
      : false;
```
Shouldn't this be false? Not sure again, better to confirm with core team

prashantasdeveloper (commented):
LGTM, but have a couple of doubts for which I left comments

src/api/procedures/executeManualInstruction.ts
polymath-eric: This is checking for any involvement in the instruction at all, so I don't think this is the place to check for that is we need it.

If there should be no offchain parties then the check would be true in that case anyway so shouldn't really change things here.

src/api/procedures/modifyInstructionAffirmation.ts
polymath-eric: If it was false then it would be impossible to affirm any instruction without a venue, which can't be right.

src/api/procedures/executeManualInstruction.ts
polymath-eric: I think the suggested logic is the same. signerDid should always be defined so we don't need the separate details check. If ?.handles undefined, then it won't equal signer did.

polymath-eric:
I added a guard to prevent making an instruction with offchain legs but no venue

prashantasdeveloper (approved):

PR #1289, opened 2024-09-04
fix: 🐛 allow reject a proposal by creator with no other votes + modify required signatures

sansan:
### Description

- allows a multiSig propoposal to be rejected by creator if no other votes have been casted
- allows modifying the required signatures for a multiSig

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1288

### DIFF
--- a/src/api/procedures/__tests__/modifyMultiSig.ts
+++ b/src/api/procedures/__tests__/modifyMultiSig.ts
@@ -3,11 +3,15 @@ import BigNumber from 'bignumber.js';
 import { when } from 'jest-when';
 
 import {
+  assertNoDataChange,
+  assertRequiredSignersExceedsSigners,
+  assertValidRequiredSignatures,
   getAuthorization,
+  modifyMultiSig,
   prepareModifyMultiSig,
   prepareStorage,
 } from '~/api/procedures/modifyMultiSig';
-import { Context, ModifyMultiSigStorage, MultiSig, PolymeshError } from '~/internal';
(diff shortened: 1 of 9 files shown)


src/api/procedures/__tests__/modifyMultiSig.ts
```diff
@@ -330,7 +384,135 @@ describe('modifyMultiSig procedure', () => {
...
+        code: ErrorCode.NoDataChange,
+        message:
+          'The given signers are equal to the current signers. At least one signer should be added or removed',
       });
+
+      try {
+        assertNoDataChange(new BigNumber(2), [], [], [oldSigner1, oldSigner2]);
+      } catch (error) {
+        expect(error).toEqual(expectedError);
+      }
```
polymath-eric, line 479:
```suggestion
        expect(() => assertNoDataChange(new BigNumber(2), [], [], [oldSigner1, oldSigner2])).toThrow(expectedError);
```

I think passing a function is cleaner syntax vs try catch.

polymath-eric: With catch syntax would the test fail if the assert function didn't throw?

It looks like the expect would be skipped and the test would still pass, so its not just a cosmetic change.

I think with try catch you need to assign to an error variable and assert outside the catch.

polymath-eric (dismissed):

prashantasdeveloper (approved):

PR #1300, opened 2024-09-11
fix: 🐛 revert mistaken breaking changes in v24

polymath-eric:
### Description

reverts breaking changes around offchain legs in v24 branch. Offchain leg support is now in v25

### Checklist


prashantasdeveloper:
@polymath-eric  You should rename the base branch to something else. When tags are released new branches are created with that naming. So when 24.7.1 will get released, it will try to create v24.7.1 which may cause issues again. Renaming to something like `base-24.7` IDK, anything but `v24.7.1`. And same changes in release config as well

PR #1301, opened 2024-09-11
feat: 🎸 reintroduce joinCreator for dual-version

sansan:
### Description

- reintroduces joinCreator for dual version
- throws an error if joinCreator called in v7

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1311

### Checklist

### DIFF
--- a/src/api/procedures/__tests__/joinCreator.ts
+++ b/src/api/procedures/__tests__/joinCreator.ts
@@ -168,38 +168,13 @@ describe('joinCreator procedure', () => {
       resolver: undefined,
     });
   });
-
-  it('should throw an error if called from v7', async () => {
-    mockContext = dsMockUtils.getContextInstance({ isV6: false });
-
-    const proc = procedureMockUtils.getInstance<Params, void>(mockContext);
-
-    const expectedError = new PolymeshError({
-      code: ErrorCode.ValidationError,
-      message:
-        'This method is deprecated. MultiSig automatically is attached to the creators identity on creation.',
-    });
-
-    await expect(
-      prepareJoinCreator.call(proc, {
-        asPrimary: false,
-        multiSig,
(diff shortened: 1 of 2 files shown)


src/api/entities/Account/MultiSig/index.ts
```diff
@@ -37,12 +39,32 @@ export class MultiSig extends Account {
...
+    if (isV6) {
+      this.joinCreator = createProcedureMethod(
+        {
+          getProcedureAndArgs: joinArgs => [joinCreator, { multiSig: this, ...joinArgs }],
+          optionalArgs: true,
+        },
+        context
+      );
+    } else {
+      this.joinCreator = function () {
```
polymath-eric, line 60: This should have a return type like GitHub says, even if it requires some weird cast

src/api/entities/Account/MultiSig/index.ts
```diff
@@ -224,13 +246,17 @@ export class MultiSig extends Account {
       address,
     } = this;
 
+    console.log('getAdmin');
```
polymath-eric, line 249:
```suggestion
```

src/api/entities/Account/MultiSig/index.ts
```diff
@@ -288,19 +314,23 @@ export class MultiSig extends Account {
       address,
     } = this;
 
+    console.log('getCreator');
```
polymath-eric, line 317:
```suggestion
```

src/api/entities/Account/MultiSig/index.ts
polymath-eric: Why not put this logic inside the procedure though? Then it would be just a single throw, vs having to declare some inline function

polymath-eric (approved):

PR #1302, opened 2024-09-11
feat: 🎸 add set admin method to multi instance

polymath-eric:
### Description

allow for management of a multiSig admin for v7 chains. Signers can create a proposal to set an admin and an admin can relinquish the privilege

### Checklist


polymath-eric:
Needs to be draft for now since this will need to be updated for the approve weight as an option

prashantasdeveloper (approved):

PR #1305, opened 2024-09-12
fix: 🐛 allow joined account to accept rotate to secondary auth

polymath-eric:
### Description

previously the auth `rotatePrimaryKeyToSecondary` was unable to be accepted by a key already joined to the identity, although it is allowed by the chain

### Breaking Changes

None

### JIRA Link

✅ Closes: DA-1310

### Checklist

### DIFF
--- a/src/api/procedures/utils.ts
+++ b/src/api/procedures/utils.ts
@@ -537,16 +537,19 @@ async function assertJoinOrRotateAuthorizationValid(
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
       message: 'Issuing Identity does not have a valid CDD claim',
+      data: { issuer: issuer.did },
     });
   }
 
   assertIsAccount(target);
 
   const targetIdentity = await target.getIdentity();
-  if (targetIdentity) {
+
+  if (targetIdentity && !targetIdentity.isEqual(issuer)) {
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
       message: 'The target Account already has an associated Identity',
+      data: { associatedIdentity: targetIdentity.did },
     });
   }
 }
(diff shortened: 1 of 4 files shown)


src/api/procedures/utils.ts
```diff
@@ -537,16 +537,19 @@ async function assertJoinOrRotateAuthorizationValid(
...
+      data: { issuer: issuer.did },
     });
   }
 
   assertIsAccount(target);
 
   const targetIdentity = await target.getIdentity();
-  if (targetIdentity) {
+
+  if (targetIdentity && !targetIdentity.isEqual(issuer)) {
```
F-OBrien, line 548: @polymath-eric does get identity return an identity for a multisig signing key? Signing keys also cannot be rotated to a primary.

polymath-eric: `getIdentity` will return the MultiSig identity in that case. However, the entire transaction will end up being wrapped in a proposal, so it will be to rotate the MultiSig as primary, not the signing key.

MultiSig signing keys are blocked from submitting transactions directly with some exceptions like approve/reject proposals.

polymath-eric: In general, if a MultiSig signer key is targeted with an auth they would have issues rejecting them via the SDK.

Either we would need to expose an option to let the user specify being able to override the isProposal or not logic, or we extend that logic with inference in the case of auth is targeting.

prashantasdeveloper (approved):

src/api/procedures/utils.ts
F-OBrien: Thanks for clarifying. Accepting an authorization is one of the few instances where the originator of the request pays the fee so that's why I was thinking the sdk might not block it. If the signer key can't submit any a tx to accept an authorization where they (not the MS) is the target then this is already covered.

F-OBrien (approved):

PR #1308, opened 2024-09-13
fix: 🐛 revert `createdAt` being nullable

polymath-eric:
### Description

This means the method will throw if ran on an executed instruction, but we remain API compatible.

The field should be nullable since when its properly checked the method works as expected.

### Checklist


polymath-eric:
Should be closed in favor of: https://github.com/PolymeshAssociation/polymesh-sdk/pull/1309

PR #1311, opened 2024-09-17
feat: 🎸 update chain types to latest

polymath-eric:
### Description

update chain types to 6.3.x. Updates typgen code to use runtime types and update snake case params to camel case

Code from: https://github.com/PolymeshAssociation/polymesh-sdk/tree/typegen

### Checklist


polymath-eric:
I was getting errors leaving the snake case, so brought in those changes as well and tested them out locally.

PR #1348, opened 2024-10-30
feat: 🎸 Allow creation of nft collection with custom nft type

prashantasdeveloper:
### Description

Allows creation of nft collection with custom nft type string. Earlier error was thrown here

### Breaking Changes

NA

### JIRA Link

DA-1335

### Checklist

### DIFF
--- a/src/api/procedures/createNftCollection.ts
+++ b/src/api/procedures/createNftCollection.ts
@@ -45,6 +45,7 @@ import {
   securityIdentifierToAssetIdentifier,
   stringToBytes,
   stringToTicker,
+  u32ToBigNumber,
 } from '~/utils/conversion';
 import { checkTxType, isAllowedCharacters, optionize } from '~/utils/internal';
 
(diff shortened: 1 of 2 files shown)


src/api/procedures/createNftCollection.ts
```diff
@@ -396,23 +409,26 @@ async function prepareStorageForCustomType(
...
-          'createNftCollection procedure was given a custom type string that does not have a corresponding ID. Register the type and try again',
-        data: { nftType },
-      });
+      const rawCustomAssetTypeId = await context.polymeshApi.query.asset.customTypeIdSequence();
+      const nextCustomAssetTypeId = u32ToBigNumber(rawCustomAssetTypeId).plus(1);
+      customTypeData = {
+        rawId: bigNumberToU32(nextCustomAssetTypeId, context),
+        rawValue,
+        isAlreadyCreated: false,
+      };
```
prashantasdeveloper, line 424: Like we do for Asset ID (where we calculate the next Asset ID to be generated and then batch the other txs with this ID), I have calculated the next custom asset type ID to be generated and later used the same while creating Asset (After registering the custom asset ID first)

VictorVicente (approved):

PR #1357, opened 2024-10-30
feat: 🎸 support hex input in internal `asBaseAsset` method

polymath-eric:
### Description

support hex input in a missed internal method

### Checklist

### DIFF
--- a/src/api/client/Assets.ts
+++ b/src/api/client/Assets.ts
@@ -41,6 +41,7 @@ import {
 import {
   asAsset,
   assembleAssetQuery,
+  asUuid,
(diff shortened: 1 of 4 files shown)


src/api/client/Assets.ts
```diff
@@ -345,6 +346,7 @@ export class Assets {
     const { context } = this;
 
     const assetIdValue = await this.getIdFromAssetIdAndTicker(assetId, ticker);
+    console.log({ assetIdValue });
```
F-OBrien, line 349: console.log should be removed

VictorVicente (approved):

PR #1360, opened 2024-11-04
feat: 🎸 Support subsidized batches

prashantasdeveloper:
### Description

Add support for subsidized batches

### Breaking Changes

NA

### JIRA Link

DA-1355

### Checklist

### DIFF
--- a/src/base/PolymeshTransaction.ts
+++ b/src/base/PolymeshTransaction.ts
@@ -3,8 +3,8 @@ import { ISubmittableResult } from '@polkadot/types/types';
 import BigNumber from 'bignumber.js';
 
 import { isMultiSigNoWrapTx } from '~/base/utils';
-import { Context, PolymeshTransactionBase } from '~/internal';
-import { TxTag, TxTags } from '~/types';
+import { Context, PolymeshError, PolymeshTransactionBase } from '~/internal';
+import { ErrorCode, TxTag, TxTags } from '~/types';
 import { PolymeshTx, TransactionConstructionData, TransactionSpec } from '~/types/internal';
 import { transactionToTxTag } from '~/utils/conversion';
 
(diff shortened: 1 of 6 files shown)


src/utils/constants.ts
```diff
@@ -11,6 +11,7 @@ export const MAX_TICKER_LENGTH = 12;
 export const MAX_MODULE_LENGTH = 32;
 export const MAX_MEMO_LENGTH = 32;
 export const MAX_OFF_CHAIN_METADATA_LENGTH = 32;
+export const MAX_BATCH_SIZE_SUPPORTING_SUBSIDY = 5;
```
prashantasdeveloper, line 14: This value needs to be increased upto 7 from the core side in order to support `createNftCollection`

src/base/PolymeshTransactionBase.ts
```diff
@@ -712,7 +712,7 @@ export abstract class PolymeshTransactionBase<
    *
    * @note this depends on the type of transaction itself (e.g. `staking.bond` can't be subsidized, but `asset.createAsset` can)
    */
-  public abstract supportsSubsidy(): boolean;
+  public abstract assertTransactionSupportsSubsidy(): void;
```
prashantasdeveloper, line 715: since i renamed this, do you think @polymath-eric  this should be marked as breaking change? I'm not sure end users have been using this method for explicitly checking if the transactions supports subsidy or not?

polymath-eric: I think `supportsSubsidy` should be left as it is.

Internally assert methods are convenient, but for a user they aren't so nice to use for checking things like this.

If you want an `assertTransactionSupportsSubsidy` make it `protected` and have it call `supportsSubsidy`

polymath-eric: Regardless of if people are using it or not, if a method is in a public API gets changed, then its a breaking change. Better to avoid it now imo.

src/base/PolymeshTransaction.ts
```diff
@@ -133,4 +133,17 @@ export class PolymeshTransaction<
 
     return context.supportsSubsidy({ tag });
   }
+
+  // eslint-disable-next-line require-jsdoc
```
polymath-eric, line 137:
```suggestion
  /** @hidden */
```

polymath-eric (approved):

polymath-eric (approved):

PR #1372, opened 2024-11-20
refactor: 💡 add handling of undefined gql variables

sansan:
### Description

- adds default values to be always used for page start and size
- removes undefined values from variables passed to gql query

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1357

### DIFF
--- a/src/middleware/queries/externalAgents.ts
+++ b/src/middleware/queries/externalAgents.ts
@@ -2,7 +2,7 @@ import { QueryOptions } from '@apollo/client/core';
 import BigNumber from 'bignumber.js';
 import gql from 'graphql-tag';
 
-import { createArgsAndFilters } from '~/middleware/queries/common';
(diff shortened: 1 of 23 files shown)


src/middleware/queries/externalAgents.ts
```diff
@@ -122,6 +122,6 @@ export function tickerExternalAgentActionsQuery(
 
   return {
     query,
-    variables: { ...filters, size: size?.toNumber(), start: start?.toNumber() },
+    variables: { ...filters, ...getSizeAndOffset(size, start) },
```
polymath-eric, line 125: why not always remove undefined here? does it have to do with having all the args produced internally vs user provided?

polymath-eric:
doesn't matter, but why label it as a refactor? This is getting bundled with other changes, but on its own wouldn't trigger a release.

PR #1379, opened 2024-11-27
fix: 🐛 Fix logic to check if instruction exists

prashantasdeveloper:
### Description

When an instruction is queried we check that the instruction ID is less than or equal to the instruction counter from the chain. Description in metadata states `InstructionCounter - Number of instructions in the system (It's one more than the actual number)`. As a result if we query an instruction one more than the latest instruction the chain SDK doesn't throw an error when it should. This fixes the same logic

### Breaking Changes

NA

### JIRA Link

DA-1400

### Checklist

### DIFF
--- a/src/api/entities/Instruction/index.ts
+++ b/src/api/entities/Instruction/index.ts
@@ -1,4 +1,7 @@
-import { PolymeshPrimitivesSettlementInstructionStatus } from '@polkadot/types/lookup';
+import {
+  PolymeshPrimitivesSettlementInstructionStatus,
+  PolymeshPrimitivesSettlementLeg,
+} from '@polkadot/types/lookup';
 import { hexAddPrefix, hexStripPrefix } from '@polkadot/util';
 import BigNumber from 'bignumber.js';
 import P from 'bluebird';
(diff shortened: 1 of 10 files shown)


F-OBrien (approved):

src/api/entities/Instruction/index.ts
```diff
@@ -550,6 +561,110 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
...
+   * @hidden
+   *
+   * Retrieve all legs of this Instruction from chain
+   *
+   * @note supports pagination
+   *
+   * @throws if the instruction is executed/rejected and was pruned from chain (when querying from chain)
+   */
+  public async getLegsFromChain(paginationOpts?: PaginationOptions): Promise<ResultSet<Leg>> {
+    const {
```
prashantasdeveloper, line 574: I also added this method to allow REST API to fetch legs and return those when an instruction is just created. Also, while testing middleware tests with different scenarios, it felt this method would be really useful (if someone doesn't have to wait for middleware to sync). I earlier refactored this to context class, but here the usage felt more suitable. 

FYI I have kept it as hidden (from docs) for now.

PR #1380, opened 2024-11-27
fix: 🐛 Fetch details from chain on instruction execution/affirm

prashantasdeveloper:
### Description

Fetch details from chain on instruction execution/affirmation

### Checklist

### DIFF
--- a/src/api/entities/Instruction/index.ts
+++ b/src/api/entities/Instruction/index.ts
@@ -396,16 +396,13 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
   }
 
   /**
-   * Retrieve information specific to this Instruction
+   * @hidden
    *
-   * @note uses middleware (if available) to retrieve information, otherwise directly queries from the chain
+   * Retrieve information specific to this Instruction from chain
    *
-   * @throws if
-   *  - instruction does not exists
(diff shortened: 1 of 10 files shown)


src/api/entities/Instruction/index.ts
```diff
@@ -396,16 +396,13 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
...
+   * Retrieve information specific to this Instruction from chain
    *
-   * @throws if
-   *  - instruction does not exists
-   *  - instruction is not yet processed by the middleware (when querying from middleware)
-   *  - instruction is executed/rejected and was pruned from chain (when querying from chain)
+   * @throws if instruction is executed/rejected and was pruned from chain
    */
-  public async details(): Promise<InstructionDetails> {
+  public async detailsFromChain(): Promise<InstructionDetails> {
```
polymath-eric, line 405: does this need to be public? I don't think a user should need to think between `details` vs `detailsFromChain`

You should be able to get test coverage by calling details directly

prashantasdeveloper: Main point is to have procedures to use details from chain instead of waiting for middleware to process the instruction to all affirming/executing. Also, there maybe changes where end users want that to, so felt leaving that in with @hidden should do the trick

PR #1390, opened 2024-12-02
test: 💍 add nft creation with name test case

polymath-eric:
### Description

add nft missing coverage case

### Checklist


prashantasdeveloper (approved):

polymath-eric:
I don't get why a merge into alpha would be OK without this case, but the alpha into beta is missing the coverage. The double ?? is bad for sonar complexity measure though, it wants 8 cases


sonar-complexity

PR #1402, opened 2024-12-05
fix: 🐛 handle `tradeDate` and `valueDate` of `null` in instruction details

F-OBrien:
### Checklist

### DIFF
--- a/src/api/entities/Instruction/index.ts
+++ b/src/api/entities/Instruction/index.ts
@@ -474,8 +474,8 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
       return {
         status: middlewareInstructionStatusToInstructionStatus(status),
         createdAt: new Date(createdBlock!.datetime),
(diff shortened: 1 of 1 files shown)


src/api/entities/Instruction/index.ts
```diff
@@ -474,8 +474,8 @@ export class Instruction extends Entity<UniqueIdentifiers, string> {
       return {
         status: middlewareInstructionStatusToInstructionStatus(status),
         createdAt: new Date(createdBlock!.datetime),
-        tradeDate: new Date(tradeDate),
```
polymath-eric, line 477: does this need a test case for branch coverage?

As long as sonar is fine with it its OK to merge

polymath-eric (approved):

polymath-eric (changes requested):
looking into the code coverage. The change looks OK though, but should have a test case for the coverage metric

polymath-eric (approved):

PR #1413, opened 2024-12-13
chore: 🤖 review permissions

sansan:
### Description
- for discussion
- adds comments comparing permissions used on chain vs those that we check for with getAuthorization for procedures

on chain there are multiple ways how we check permissions, the main ones are
 - Identity::check_perms -> cheks if the identity can execute module/extrinsic on chain 
 - External Agents:check_perms -> checks if the external agent has ownership ower asset, can execute the tx (module/extrinsic)
 - Checks for portfolio ownership

Additional checks are done via specific methods in the extrinsic - e.g. if the signer is multiSig admin, if it is multiSig

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1411

### DIFF
--- a/src/api/procedures/addSecondaryAccountsWithAuth.ts
+++ b/src/api/procedures/addSecondaryAccountsWithAuth.ts
@@ -81,6 +81,7 @@ export async function getAuthorization(
 
   return {
     permissions: {
+      // TODO: on chain just checks if is signed by primary key -> so we should check in prepareAddSecondaryKeysWithAuth if signer is the target primary key
       transactions: [TxTags.identity.AddSecondaryKeysWithAuthorization],
       assets: [],
       portfolios: [],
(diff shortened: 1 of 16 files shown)


src/api/procedures/removeSecondaryAccounts.ts
```diff
@@ -55,6 +55,7 @@ export async function prepareRemoveSecondaryAccounts(
 export const removeSecondaryAccounts = (): Procedure<RemoveSecondaryAccountsParams> =>
   new Procedure(prepareRemoveSecondaryAccounts, {
     permissions: {
+      // TODO: might not need this -> chain checks if called by primary key, checks if secondary key exists and can be unlinked, then unlinks
```
polymath-eric, line 58: > chain checks if called by primary key,

I don't get this. Technically we don't need to check any permissions since the chain always checks them, but we do this to avoid having the chain catch obvious errors.

src/api/procedures/addSecondaryAccountsWithAuth.ts
```diff
@@ -81,6 +81,7 @@ export async function getAuthorization(
 
   return {
     permissions: {
+      // TODO: on chain just checks if is signed by primary key -> so we should check in prepareAddSecondaryKeysWithAuth if signer is the target primary key
```
polymath-eric, line 84: I think there is a few methods like this. Would we need to add an extra field to permissions for that?

polymath-eric:
@sansan do we still need this PR open?

PR #1449, opened 2025-03-06
feat: 🎸 Add optional skip checks for procedures and on getting asset

prashantasdeveloper:
### Description

* Adds in optional `skipExistsCheck` to methods of Assets class to get fungible and non-fungible asset
* Adds optional `skipChecks` to `ProcedureOpts` to allow skipping of various checks at procedure level

### JIRA Link

DA-1465

### Checklist

### DIFF
--- a/src/api/client/Assets.ts
+++ b/src/api/client/Assets.ts
@@ -269,17 +269,28 @@ export class Assets {
   /**
    * Retrieve a FungibleAsset
    *
+   * @param args.assetId - Unique Id of the Fungible Asset (for spec version 6.x, this is same as ticker)
    * @param args.ticker - Asset ticker
-   * @param args.assetId - Unique Id of the Asset (for spec version 6.x, this is same as ticker)
+   * @param args.skipExistsCheck - when true, method will not check if the Asset exists
    */
-  public async getFungibleAsset(args: { assetId: string }): Promise<FungibleAsset>;
-  public async getFungibleAsset(args: { ticker: string }): Promise<FungibleAsset>;
+  public async getFungibleAsset(args: {
+    assetId: string;
+    skipExistsCheck?: boolean;
+  }): Promise<FungibleAsset>;
+
+  // eslint-disable-next-line require-jsdoc
+  public async getFungibleAsset(args: {
+    ticker: string;
+    skipExistsCheck?: boolean;
+  }): Promise<FungibleAsset>;
+
   // eslint-disable-next-line require-jsdoc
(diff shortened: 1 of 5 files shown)


src/api/client/Assets.ts
```diff
@@ -269,31 +269,46 @@ export class Assets {
...
    * Retrieve a FungibleAsset
    *
+   * @param args.assetId - Unique Id of the Fungible Asset (for spec version 6.x, this is same as ticker)
    * @param args.ticker - Asset ticker
-   * @param args.assetId - Unique Id of the Asset (for spec version 6.x, this is same as ticker)
+   * @param args.skipExistsCheck - when true, method will not check if the Asset exists
    */
-  public async getFungibleAsset(args: { assetId: string }): Promise<FungibleAsset>;
-  public async getFungibleAsset(args: { ticker: string }): Promise<FungibleAsset>;
+  public async getFungibleAsset(args: {
```
VictorVicente, line 276: Why do we need these extra interfaces? Aren't the covered by the last one?

VictorVicente: Thinking it twice, I guess the last one is the one which we should remove as the other two ensures there is at least one asset reference

src/base/Procedure.ts
```diff
@@ -166,29 +168,51 @@ export class Procedure<Args = void, ReturnValue = void, Storage = Record<string,
...
+    agentPermissionsResult: CheckPermissionsResult<SignerType.Identity>;
+    noIdentity: boolean;
+  }> {
+    const { roles, agentPermissions } = permissions;
+
+    const { roles: skipRolesCheck, agentPermissions: skipAgentPermissionsCheck } = skipChecks || {};
 
     let identity: Identity | null = null;
     let rolesResult: CheckRolesResult;
     let noIdentity = false;
```
VictorVicente, line 209: Is there a way where `noIdentity` can be true?

F-OBrien (approved):

src/api/client/Assets.ts
VictorVicente: Yes, I know the overloaded format. But I mean the third case won't happen where both `ticker` and `assetId` are both not present

prashantasdeveloper: Last one is not a new variant, its the implementation of the above two formats

VictorVicente: Haha that's why I should read the code on the IDE rather than here. Beginning of the implementation was hidden

PR #1460, opened 2025-03-15
feat: 🎸 remove CorporateBallot

sansan:
### Description

adds a method to remove the CorporateBallot

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1484

### DIFF
--- a/src/api/entities/Asset/Fungible/CorporateActions/Ballots.ts
+++ b/src/api/entities/Asset/Fungible/CorporateActions/Ballots.ts
@@ -1,13 +1,12 @@
 import BigNumber from 'bignumber.js';
 
 import { createBallot } from '~/api/procedures/createBallot';
-import { removeBallot } from '~/api/procedures/removeBallot';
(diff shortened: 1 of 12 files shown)


src/api/entities/Asset/Fungible/CorporateActions/Ballots.ts
```diff
@@ -49,47 +51,23 @@ export class Ballots extends Namespace<FungibleAsset> {
    * @throws if the provided Corporate Action does not exist
    */
   public async getOne(args: { id: BigNumber }): Promise<CorporateBallot> {
```
prashantasdeveloper, line 53:
```suggestion
  public async getOne(args: { id: BigNumber }): Promise<CorporateBallotWithDetails> {
```
Similar to dividend distribution, we should create a separate interface here `CorporateBallotWithDetails` 
```
export interface BallotWithDetails {
  ballot: CorporateBallot;
  details: CorporateBallotDetails;
}
```
Ballot details should not be clubbed within entity info

prashantasdeveloper (changes requested):

PR #1466, opened 2025-03-27
chore: 🤖 update release-config for develop

prashantasdeveloper:
### Description

update release-config for develop

### JIRA Link

DA-1500, DA-1487

### Checklist

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -3,12 +3,14 @@ module.exports = {
   branches: [
     'master',
     {
-      name: 'beta',
-      prerelease: true,
+      name: 'develop',
+      channel: 'beta',
+      prerelease: 'beta',
     },
     {
-      name: 'alpha',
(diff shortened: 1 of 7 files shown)


release.config.js
```diff
@@ -3,12 +3,14 @@ module.exports = {
...
+      name: 'develop',
+      channel: 'beta',
+      prerelease: 'beta',
     },
     {
-      name: 'alpha',
-      prerelease: true,
+      name: 'next',
+      channel: 'next',
+      prerelease: 'next',
```
prashantasdeveloper, line 13: I'm not sure we keep the next as prerelease or not?

VictorVicente: I think it should be prerelease. Shouldn't it be a boolean instead?

VictorVicente (approved):

PR #1492, opened 2025-05-19
chore: 🤖 stricter compiler rules

sansan:
### Description

- adds stricret TS rules

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1489

### DIFF
--- a/src/api/client/Assets.ts
+++ b/src/api/client/Assets.ts
@@ -173,9 +173,9 @@ export class Assets {
     const rawAssetIds = await asset.tickerAssetId.multi(rawTickers);
 
     return rawAssetIds.reduce<TickerReservation[]>((result, rawAssetId, index) => {
-      const rawTicker = rawTickers[index];
+      const rawTicker = rawTickers[index]!;
 
-      if (rawAssetId.isNone && rawTicker) {
+      if (rawAssetId.isNone) {
         const ticker = tickerToString(rawTicker);
 
         if (isPrintableAscii(ticker)) {
@@ -380,7 +380,7 @@ export class Assets {
     } = this;
 
     const { entries, lastKey: next } = await requestPaginated(asset.assetNames, {
-      ...(paginationOpts && { paginationOpts }),
+      paginationOpts,
     });
 
     const assetIds: string[] = [];
--- a/src/api/client/Assets.ts
+++ b/src/api/client/Assets.ts
@@ -173,8 +173,10 @@ export class Assets {
     const rawAssetIds = await asset.tickerAssetId.multi(rawTickers);
 
     return rawAssetIds.reduce<TickerReservation[]>((result, rawAssetId, index) => {
-      if (rawAssetId.isNone) {
-        const ticker = tickerToString(rawTickers[index]);
+      const rawTicker = rawTickers[index];
+
+      if (rawAssetId.isNone && rawTicker) {
+        const ticker = tickerToString(rawTicker);
 
         if (isPrintableAscii(ticker)) {
           return [...result, new TickerReservation({ ticker }, context)];
@@ -378,7 +380,7 @@ export class Assets {
     } = this;
 
     const { entries, lastKey: next } = await requestPaginated(asset.assetNames, {
-      paginationOpts,
+      ...(paginationOpts && { paginationOpts }),
     });
 
     const assetIds: string[] = [];
--- a/src/api/client/Polymesh.ts
+++ b/src/api/client/Polymesh.ts
@@ -174,14 +174,14 @@ export class Polymesh {
         rpc,
         runtime: runtime as unknown as DefinitionsCall,
         signedExtensions,
-        metadata,
-        noInitWarn,
-        typesBundle,
+        ...(metadata ? { metadata } : {}),
+        ...(noInitWarn ? { noInitWarn } : {}),
+        ...(typesBundle ? { typesBundle } : {}),
       });
       context = await Context.create({
         polymeshApi,
         middlewareApiV2: createMiddlewareApi(middlewareV2),
-        signingManager,
+        ...(signingManager ? { signingManager } : {}),
       });
     } catch (err) {
       const { message, code } = err;
--- a/src/api/entities/Account/index.ts
+++ b/src/api/entities/Account/index.ts
@@ -224,11 +224,11 @@ export class Account extends Entity<UniqueIdentifiers, string> {
       extrinsicsByArgs(
         context.isSqIdPadded,
         {
-          blockId: blockNumber ? blockNumber.toString() : undefined,
+          ...(blockNumber ? { blockId: blockNumber.toString() } : {}),
           address,
-          moduleId,
-          callId,
-          success: successFilter,
+          ...(moduleId ? { moduleId } : {}),
+          ...(callId ? { callId } : {}),
+          ...(successFilter ? { success: successFilter } : {}),
         },
         size,
         start,
@@ -634,11 +634,24 @@ export class Account extends Entity<UniqueIdentifiers, string> {
 
     const rawBytes = blake2AsU8a(data, 128);
 
+    const rawBytesSix = rawBytes[6];
+    const rawBytesEight = rawBytes[8];
+
+    if (!rawBytesSix || !rawBytesEight) {
+      throw new PolymeshError({
+        code: ErrorCode.UnexpectedError,
+        message: 'Unexpected error while generating asset ID',
+        data: {
+          rawBytes,
+        },
+      });
+    }
+
     // Need to override 6bits to make it a valid v8 UUID.
-    rawBytes[6] = (rawBytes[6] & 0x0f) | 0x80;
+    rawBytes[6] = (rawBytesSix & 0x0f) | 0x80;
 
     // Set the RFC4122 variant (bits 10xx) in the 8th byte
-    rawBytes[8] = (rawBytes[8] & 0x3f) | 0x80;
+    rawBytes[8] = (rawBytesEight & 0x3f) | 0x80;
 
     return hexToUuid(u8aToHex(rawBytes));
   }
--- a/src/api/entities/Account/index.ts
+++ b/src/api/entities/Account/index.ts
@@ -634,18 +634,8 @@ export class Account extends Entity<UniqueIdentifiers, string> {
 
     const rawBytes = blake2AsU8a(data, 128);
 
-    const rawBytesSix = rawBytes[6];
-    const rawBytesEight = rawBytes[8];
-
-    if (!rawBytesSix || !rawBytesEight) {
-      throw new PolymeshError({
-        code: ErrorCode.UnexpectedError,
-        message: 'Unexpected error while generating asset ID',
-        data: {
-          rawBytes,
-        },
-      });
-    }
+    const rawBytesSix = rawBytes[6]!;
+    const rawBytesEight = rawBytes[8]!;
 
     // Need to override 6bits to make it a valid v8 UUID.
     rawBytes[6] = (rawBytesSix & 0x0f) | 0x80;
--- a/src/api/entities/Account/index.ts
+++ b/src/api/entities/Account/index.ts
@@ -224,11 +224,11 @@ export class Account extends Entity<UniqueIdentifiers, string> {
       extrinsicsByArgs(
         context.isSqIdPadded,
         {
-          ...(blockNumber ? { blockId: blockNumber.toString() } : {}),
+          blockId: blockNumber?.toString(),
           address,
-          ...(moduleId ? { moduleId } : {}),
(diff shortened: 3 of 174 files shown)


src/api/client/Assets.ts
```diff
@@ -173,8 +173,10 @@ export class Assets {
     const rawAssetIds = await asset.tickerAssetId.multi(rawTickers);
 
     return rawAssetIds.reduce<TickerReservation[]>((result, rawAssetId, index) => {
-      if (rawAssetId.isNone) {
-        const ticker = tickerToString(rawTickers[index]);
+      const rawTicker = rawTickers[index];
+
+      if (rawAssetId.isNone && rawTicker) {
+        const ticker = tickerToString(rawTicker);
```
prashantasdeveloper: The was no need to add the extra condition here. `rawTickers[index]` will always be defined. A mere type assertion would have done the trick. 
```suggestion
      if (rawAssetId.isNone) {
        const ticker = tickerToString(rawTickers[index] as PolymeshPrimitivesTicker);
```

prashantasdeveloper (commented):
There are other places as well type assertion could have worked (generally when the called method has a type defined). Also, maybe we can figure out some other way to handle optional properties

src/api/client/Polymesh.ts
```diff
@@ -175,14 +175,14 @@ export class Polymesh {
         rpc,
         runtime: runtime as unknown as DefinitionsCall,
         signedExtensions,
-        metadata,
-        noInitWarn,
-        typesBundle,
+        ...(metadata ? { metadata } : {}),
```
polymath-eric, line 177: To me this seems like its making the code worse. 

Its adding more characters to read and conditional paths to test. 

I don't see the benefit for adding in some rule that doesn't play nice with 3rd party libs.

src/api/entities/Account/index.ts
```diff
@@ -224,11 +224,11 @@ export class Account extends Entity<UniqueIdentifiers, string> {
       extrinsicsByArgs(
         context.isSqIdPadded,
         {
-          blockId: blockNumber ? blockNumber.toString() : undefined,
+          ...(blockNumber ? { blockId: blockNumber.toString() } : {}),
```
polymath-eric: Another example of more complex/hard to follow code because of the rule.

src/api/entities/Account/types.ts
```diff
@@ -80,32 +80,32 @@ export interface HistoricPolyxTransaction extends EventIdentifier {
    * Identity from which the POLYX transaction has been initiated/deducted in case of a transfer.
    * @note this can be null in cases where some balance are endowed/transferred from treasury
    */
-  fromIdentity?: Identity;
+  fromIdentity?: Identity | undefined;
```
polymath-eric: Seems weird to read `?` and `| undefined`. I would have thought this would have worked. 

```suggestion
  fromIdentity: Identity | undefined;
```

From my understanding the rule is trying to stop `fromIdentity = undefined` assignment, but that shouldn't really happen, and I don't see the risk from a type perspective, consumers need to handle the undefined case regardless.

src/api/entities/Asset/Base/Metadata/index.ts
```diff
@@ -244,13 +254,13 @@ export class Metadata extends Namespace<BaseAsset> {
...
+        name: bytesToString(nameValue.unwrap()),
         specs: meshMetadataSpecToMetadataSpec(specValues[index]),
-        ...meshMetadataValueToMetadataValue(rawValue, valueDetails[index])!,
+        ...meshMetadataValueToMetadataValue(rawValue, valueDetail),
       });
       index++;
     }
 
-    return data;
+    return data as MetadataWithValue[];
```
polymath-eric: This looks like a bad cast to me. Declaring data with the type should work if TS had trouble inferring.

```ts
const data: MetadataWithValue[] = []
```

Even if it works now, we are giving up type safety on a public interface. It risks breaking our API in a way that would only be uncovered at runtime.

src/api/entities/Asset/Fungible/Checkpoints/Schedules.ts
```diff
@@ -108,13 +108,14 @@ export class Schedules extends Namespace<FungibleAsset> {
...
       const remainingCheckpoints = new BigNumber([...rawSchedule.pending].length);
+
       return {
         schedule,
         details: {
           remainingCheckpoints,
           nextCheckpointDate: points[0],
         },
-      };
+      } as ScheduleWithDetails;
```
polymath-eric: Values being returned on a public interface shouldn't be cast. It risks hiding breaking changes.

src/api/entities/Asset/Fungible/TransferRestrictions/TransferRestrictionBase.ts
```diff
@@ -277,10 +277,10 @@ export abstract class TransferRestrictionBase<
           },
         ]) => identityIdToString(scopeId) // `ScopeId` and `PolymeshPrimitivesIdentityId` are the same type, so this is fine
       );
-      const { value } = transferConditionToTransferRestriction(
-        filteredRequirements[index],
-        context
-      );
+
+      const requirement = filteredRequirements[index]!;
```
polymath-eric: `index` here is from `rawExemptedLists.map`

I don't see why they'd be related here, but seems like it was working before.

If we are going to add stricter rules we should handle the cases they flag. Otherwise we are introducing bad patterns for no real benefit.

src/api/entities/Asset/types.ts
```diff
@@ -170,15 +170,15 @@ export interface BaseHistoricAssetTransaction extends EventIdentifier {
   /**
    * Name of the funding round (if provided while issuing the Asset). This value is present only when the value of `event` is `Issued`
    */
-  fundingRound?: string;
+  fundingRound?: string | undefined;
```
polymath-eric: (nit) The `?` is redundant now.

```suggestion
  fundingRound: string | undefined;
```

src/api/entities/CorporateActionBase/index.ts
```diff
@@ -235,10 +235,16 @@ export abstract class CorporateActionBase extends Entity<UniqueIdentifiers, unkn
...
 
-    return new Checkpoint(
-      { assetId, id: u64ToBigNumber(rawCheckpointIds[createdCheckpointIndex]) },
-      context
-    );
+    const rawCheckpointId = rawCheckpointIds[createdCheckpointIndex];
+
+    if (!rawCheckpointId) {
+      throw new PolymeshError({
+        code: ErrorCode.UnexpectedError,
```
polymath-eric:
```suggestion
        code: ErrorCode.DataUnavailable,
```

src/api/procedures/__tests__/addInstruction.ts
```diff
@@ -923,7 +923,7 @@ describe('addInstruction procedure', () => {
 
     await expect(
       prepareAddInstruction.call(proc, {
-        venueId: args.venueId,
+        venueId: args.venueId as BigNumber,
```
polymath-eric: (nit)
```suggestion
        venueId: args.venueId!,
```

src/api/procedures/issueNft.ts
```diff
@@ -22,11 +23,14 @@ export type Params = IssueNftParams & {
...
   (context: Context) =>
   (receipt: ISubmittableResult): Nft => {
-    const [{ data }] = filterEventRecords(receipt, 'nft', 'NFTPortfolioUpdated');
+    const [record] = filterEventRecords(receipt, 'nft', 'NFTPortfolioUpdated');
 
+    const { data } = record!;
     const { assetId, ids } = meshNftToNftId(data[1]);
 
-    return new Nft({ id: ids[0], assetId }, context);
+    const id = ids[0] as BigNumber;
```
polymath-eric:
```suggestion
    const id = ids[0]!;
```
(nit)

src/api/procedures/modifyComplianceRequirement.ts
```diff
@@ -48,7 +48,9 @@ export async function prepareModifyComplianceRequirement(
     });
   }
 
-  const [{ conditions: existingConditions }] = existingRequirements;
+  const [requirement] = existingRequirements;
+
+  const { conditions: existingConditions } = requirement as Requirement;
```
polymath-eric:
```suggestion
  const { conditions: existingConditions } = requirement!;
```
(nit)

src/api/procedures/payDividends.ts
```diff
@@ -73,7 +73,8 @@ export async function preparePayDividends(
   const alreadyClaimedList: Identity[] = [];
   holderPaidList.forEach((holderPaid, i) => {
     if (boolToBoolean(holderPaid)) {
-      alreadyClaimedList.push(new Identity({ did: signerToString(targets[i]) }, context));
+      const target = targets[i] as string | Identity;
```
polymath-eric:
```suggestion
      const target = targets[i]!
```

src/base/PolymeshTransactionBase.ts
```diff
@@ -370,7 +371,7 @@ export abstract class PolymeshTransactionBase<
     await context.assertHasSigningAddress(signingAddress);
 
     // era is how many blocks the transaction remains valid for, `undefined` for default
-    const era = mortality.immortal ? 0 : mortality.lifetime?.toNumber();
+    const era = mortality.immortal ? 0 : (mortality.lifetime?.toNumber() as number);
```
polymath-eric: The `?.` can be `undefined` so casting to a number looks wrong to me. This is working around another rule of that prevents the normal `!` cast that was used.

src/utils/conversion.ts
```diff
@@ -5110,14 +5116,14 @@ function middlewareExtrinsicPermissionsDataToTransactionPermissions(
   pallets.forEach(({ palletName, dispatchableNames }) => {
     const moduleName = stringLowerFirst(coerceHexToString(palletName));
     if ('except' in dispatchableNames) {
-      const dispatchables = [...dispatchableNames.except];
+      const dispatchables = [...(dispatchableNames.except as string[])];
```
polymath-eric:
```suggestion
      const dispatchables = [...(dispatchableNames.except!)];
```

polymath-eric (commented):
Mostly nits, but left a few minor comments. Main thing I don't like is the repeated pattern of:

```ts
...(someObj ?? {})
```

It makes it look confusing.

A smaller gripe I have is not allowing the `someField?` syntax to be used. We've usually used `null` when we want to assign a falsy value, but if there is some benefit to the rule I can live with it.

src/api/client/Polymesh.ts
sansan: in our code we can work with explicitly stating which params can be undefined.. with other lib's it's either we skip the rule here or stick with conditionally adding the properties..

sansan:
regarding snapshot:

❌ base/Context.d.ts > Context.getIdentityClaimsFromChain: Param 1 mismatch in overload #1
❌ base/Context.d.ts > Context.getIdentityClaimsFromMiddleware: Param 1 mismatch in overload #1
❌ base/Context.d.ts > Context.getSignature: Param 1 mismatch in overload #1
❌ base/Procedure.d.ts > Procedure.prepare: Param 1 mismatch in overload #1
❌ api/entities/Identity/index.d.ts > Identity.getHeldAssets: Param 1 mismatch in overload #1

in these undefined was added as a possible value for optional property

src/api/procedures/__tests__/nftControllerTransfer.ts
```diff
@@ -252,7 +252,8 @@ describe('nftControllerTransfer procedure', () => {
 
     it('should return the default portfolio if destinationPortfolio is not provided', async () => {
       mockContext.getSigningIdentity.mockResolvedValue(
-        entityMockUtils.getIdentityInstance({ did: signerDid })
+        // entityMockUtils.getIdentityInstance({ did: signerDid })
+        { did: signerDid } as unknown as Identity
```
polymath-eric:
```suggestion
        entityMockUtils.getIdentityInstance({ did: signerDid }
```

src/api/entities/Asset/types.ts
```diff
@@ -101,7 +101,7 @@ export interface AssetDetails {
   owner: Identity;
   totalSupply: BigNumber;
   fullAgents: Identity[];
-  ticker?: string;
+  ticker?: string | undefined;
```
polymath-eric, line 104:
```suggestion
  ticker: string | undefined;
```
Other places removed the `?` (e.g. `src/api/entities/Account/types.ts`), so if that's OK then `?` and `| undefined` together seem redundant

sansan: these might cause breaking changes - so API users then would need to explicitly state that ticker: undefined.. so I'd leave as is

polymath-eric (approved):

PR #1511, opened 2025-06-12
fix: 🐛 return a FungibleAsset | NftCollection

sansan:
### Description

Returns either FungibleAsset of NftCollection for assetPermissions.get

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1522

### DIFF
--- a/src/api/entities/Asset/types.ts
+++ b/src/api/entities/Asset/types.ts
@@ -364,7 +364,7 @@ export enum TransferError {
 }
 
 export interface AssetWithGroup {
-  asset: FungibleAsset;
+  asset: Asset;
   group: KnownPermissionGroup | CustomPermissionGroup;
 }
 
--- a/src/api/entities/Identity/AssetPermissions.ts
+++ b/src/api/entities/Identity/AssetPermissions.ts
@@ -44,6 +44,7 @@ import {
   stringToIdentityId,
 } from '~/utils/conversion';
 import {
+  asAsset,
   asBaseAsset,
   calculateNextKey,
   createProcedureMethod,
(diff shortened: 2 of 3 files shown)


src/api/entities/Identity/AssetPermissions.ts
```diff
@@ -116,7 +119,15 @@ export class AssetPermissions extends Namespace<Identity> {
...
-      const asset = new FungibleAsset({ assetId }, context);
+      let asset: FungibleAsset | NftCollection;
+      const typedAsset = await asAsset(assetId, context);
+
+      if (isFungibleAsset(typedAsset)) {
+        asset = new FungibleAsset({ assetId }, context);
+      } else {
+        asset = new NftCollection({ assetId }, context);
+      }
+
```
polymath-eric: The point of the helper function is to make this logic unneeded in other places. The runtime object is going to be one or the other without the if and extra construction

```ts
 const asset = await asAsset(assetId, context);
```

polymath-eric (changes requested):

src/api/entities/Identity/AssetPermissions.ts
sansan: right, looked at the return of the function, and it just showed Asset - mixed that with BaseAsset..

src/api/entities/Asset/types.ts
```diff
@@ -364,7 +364,7 @@ export enum TransferError {
 }
 
 export interface AssetWithGroup {
-  asset: FungibleAsset;
+  asset: FungibleAsset | NftCollection;
```
polymath-eric:
```suggestion
  asset: Asset;
```
There is an alias which makes it look less intimidating in the docs.

polymath-eric (commented):
left a nit about using `Asset` alias for the type. Its the same thing, but optically makes things easier to parse for the user

polymath-eric (approved):

polymath-eric (approved):

PR #1517, opened 2025-06-20
feat: 🎸 Add support for new lock status for instruction

prashantasdeveloper:
### Description

- Update middleware types with latest SQ - 19.3.0-alpha.2
- Make changes in existing methods to return lock status for instructions

### JIRA Link

DA-1523

### Checklist

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -112,7 +112,7 @@ export const ROOT_TYPES = rootTypes;
 /**
  * The Polymesh chain spec version range that is compatible with this version of the SDK
  */
-export const SUPPORTED_SPEC_VERSION_RANGE = '7.0 || 7.1 || 7.2';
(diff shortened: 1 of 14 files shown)


src/utils/constants.ts
```diff
@@ -151,7 +151,7 @@ export const DEFAULT_CDD_ID = '0x00000000000000000000000000000000000000000000000
 /**
  * Minimum version of Middleware V2 GraphQL Service (SubQuery) that is compatible with this version of the SDK
  */
-export const MINIMUM_SQ_VERSION = '18.0.2';
+export const MINIMUM_SQ_VERSION = '19.3.0';
```
VictorVicente, line 154: What would fail if SQ is < 19.3.0?

VictorVicente (approved):

src/utils/constants.ts
prashantasdeveloper: Few enums and a warning will be thrown out

PR #1519, opened 2025-06-24
feat: 🎸 New TransferRestrictions methods

polymath-eric:
### Description

Create new TransferRestrictions API which unifies all types into a unified set of methods. This makes it easier to manage restrictions and simplifies the API.

### Breaking Changes
The methods in the fungible asset namespaces:
- `transferRestrictions.count`
- `transferRestrictions.percentage`
- `transferRestrictions.claimCount`
- `transferRestrictions.claimPercentage`

Are now all in `transferRestrictions`. Restriction types must be passed in as arguments when relevant

### JIRA Link

[DA-1514](https://polymesh.atlassian.net/browse/DA-1514)

### Checklist

[DA-1514]: https://polymesh.atlassian.net/browse/DA-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
```diff
@@ -1,95 +1,216 @@
...
+        exemptKey,
+        identity: new Identity({ did }, context),
+      };
+    });
 
-    return restrictions.map((restriction, index) => ({
-      restriction,
-      value: u128ToBigNumber(values[index]),
-    }));
+    return { exemptions };
```
sansan: do we need to return them as object - rather than just an array -> if we'd return an array we don't need the `ActiveTransferRestrictionExemptions` interface and can just say `Promise<TransferRestrictionExemption[]>`

src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
```diff
@@ -1,95 +1,216 @@
...
-        restriction,
-        context
-      );
-      const stat2ndKey = transferRestrictionToPolymeshPrimitivesStatisticsStat2ndKey(
-        restriction,
-        context
-      );
+    return {
+      stats,
+    };
```
sansan: return just the AssetStat array?

sansan (changes requested):

src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
polymath-eric: The array is hard to extend. I think returning object at the top level is better since a new field can be added in if needed. For the consumer they can use destructure syntax to have similar ergonomics either way

sansan (approved):

PR #1520, opened 2025-06-26
feat: 🎸 allow multiSig signer to remove payer

sansan:
### Description

we have already the required functionality to support subsidizing MultiSig

to remove Payer via MultiSig signer we need not to check for TX permissions as in Procedure the account for which the Permissions are checked gets converted to MultiSig (not the Signer) thus resulting in getting the permissions for MultiSig which might not have (most likely) specified the RemovePayer permissions, chain does not require them and without this check the transaction executes

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1529

### DIFF
--- a/src/api/procedures/__tests__/removeMultiSigPayer.ts
+++ b/src/api/procedures/__tests__/removeMultiSigPayer.ts
@@ -150,7 +150,7 @@ describe('removeMultiSigPayer procedure', () => {
   });
 
   describe('getAuthorization', () => {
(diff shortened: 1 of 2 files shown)


prashantasdeveloper:
@sansan  Should this be from `7.3_upgrade` branch ?

sansan:
> @sansan Should this be from `7.3_upgrade` branch ?

updated

prashantasdeveloper (approved):

PR #1525, opened 2025-07-09
refactor: 💡 modify getValues to check all paths

sansan:
### Description

- allows getting all paths for asset stat values (e.g. how many accredited and non-accredited investors are there)
- updates the enableStat (adjustments to type) to allow setting stat value for investors without a jurisdiction

Note: didn't mark as breaking as the previous version is not yet in use

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1515

### DIFF
--- a/src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
+++ b/src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
@@ -1,3 +1,9 @@
+import {
+  PolymeshPrimitivesStatisticsStat1stKey,
+  PolymeshPrimitivesStatisticsStat2ndKey,
+} from '@polkadot/types/lookup';
+import BigNumber from 'bignumber.js';
+
 import { setTransferRestrictionsExemptions } from '~/api/procedures/setTransferRestrictionExemptions';
 import {
   Context,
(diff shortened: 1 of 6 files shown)


src/api/entities/Asset/Fungible/TransferRestrictions/index.ts
```diff
@@ -57,39 +64,174 @@ export class TransferRestrictions extends Namespace<FungibleAsset> {
...
+          type: stat.type,
+          value: new BigNumber(0),
+        });
+      } else {
+        const statResult = result
+          .filter(resultStat => resultStat.claim)
+          .find(
+            resultStat =>
+              resultStat.type === stat.type &&
+              resultStat.claim!.issuer!.did === stat.issuer!.did &&
```
polymath-eric: I don't like the non-null assertions here.

The filter should use a type guard or this should b`?.` type check imo. The non-null is opting out of type safety.

src/api/entities/Asset/types.ts
```diff
@@ -393,14 +396,66 @@ export interface HeldNfts {
...
+  count: BigNumber;
+};
+
 /**
- * A Transfer Restriction along with its current value
+ * Asset Stat along with its current value
  */
-export interface TransferRestrictionValues {
-  restriction: TransferRestriction;
+export interface TransferStatValues {
```
polymath-eric:
```suggestion
export interface TransferRestrictionStatValues {
```
I think Restriction helps make it clear its not tracking stuff about transfers

polymath-eric (commented):
lgtm, left a nitty comment, and then one about the over-use of non-null assertions

src/utils/conversion.ts
```diff
@@ -6025,64 +6025,39 @@ export function meshBallotDetailsToCorporateBallotDetails(
...
+    statType,
   });
 }
 
 /**
- * @hidden
+ *
+ * @param context
+ * @param args
+ * @returns
```
polymath-eric:
```suggestion
@hidden
```

src/utils/conversion.ts
```diff
@@ -6025,64 +6025,39 @@ export function meshBallotDetailsToCorporateBallotDetails(
...
```
polymath-eric:
```suggestion
 *
@hidden
```

polymath-eric (approved):

PR #1527, opened 2025-07-11
fix: 🐛 allow specify customClaimTypeId for TrustedClaimIssuer

polymath-eric:
### Breaking Changes

BREAKING CHANGE: 🧨 Return of trustedClaimIssuers.get() will be an object for custom claim containing customClaimTypeId

### Checklist

### DIFF
--- a/src/api/entities/DefaultTrustedClaimIssuer.ts
+++ b/src/api/entities/DefaultTrustedClaimIssuer.ts
@@ -1,7 +1,7 @@
 import { Context, FungibleAsset, Identity, PolymeshError } from '~/internal';
 import { trustedClaimIssuerQuery } from '~/middleware/queries/claims';
 import { Query } from '~/middleware/types';
-import { ClaimType, ErrorCode, EventIdentifier } from '~/types';
+import { ErrorCode, EventIdentifier, TrustedFor } from '~/types';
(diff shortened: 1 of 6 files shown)


src/api/entities/DefaultTrustedClaimIssuer.ts
```diff
@@ -78,7 +78,7 @@ export class DefaultTrustedClaimIssuer extends Identity {
   /**
    * Retrieve claim types for which this Claim Issuer is trusted. A null value means that the issuer is trusted for all claim types
    */
-  public async trustedFor(): Promise<ClaimType[] | null> {
+  public async trustedFor(): Promise<TrustedFor[] | null> {
```
sansan, line 81: just a comment - imho - if I were calling a function that returns null for trustedFor - I'd think that the did is trusted for none...

sansan (changes requested):

src/api/entities/DefaultTrustedClaimIssuer.ts
polymath-eric: This is a rebase of b7e6208b81e7674336804b778e782 from next. Its better to leave it since an alpha version already has it as `null`

sansan (approved):

PR #1536, opened 2025-07-16
Enhance asset getStats handling with claims

F-OBrien:
### Description

Improve the handling of asset statistics supporting scoped statistics with claim including custom claim types, so correct stat information is returned.

### Checklist

### DIFF
--- a/src/utils/conversion.ts
+++ b/src/utils/conversion.ts
@@ -278,6 +278,7 @@ import {
   StakingNomination,
   StakingPayee,
   StakingUnlockingEntry,
+  StatClaimIssuer,
   StatClaimType,
   StatType,
   TargetTreatment,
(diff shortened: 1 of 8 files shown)


src/utils/conversion.ts
```diff
@@ -5976,12 +5987,26 @@ export function assetComplianceToTransferRestrictions(
 /**
  * @hidden
  */
-export function assetStatToStat(value: PolymeshPrimitivesStatisticsStatType): AssetStat {
-  const type = value.operationType.isCount ? StatType.Count : StatType.Balance;
+export function assetStatToStat(
+  value: PolymeshPrimitivesStatisticsStatType,
+  context: Context
+): AssetStat {
+  const result: AssetStat = { type: StatType.Count }; // temporary, is set correctly below
```
polymath-eric: I think returning an object directly from the if/else statements is better.

```ts
if(value.claimIssuer.isSome) {
  ...
  return { type, claimIssuer: {...} }
 else {
   ...
   return { type }
 }
 ```

polymath-eric (approved):

PR #1549, opened 2025-08-28
feat: 🎸 Add support for filtering custom claim types

prashantasdeveloper:
### Description

This adds support to filter targeting claims, issued claims etc for custom type claims as well.

### Breaking Changes

BREAKING CHANGE: 🧨 Type of attribute `claimTypes` (present as filtering option in getters in number of methods in Claims namespace ) has been changed from `ClaimType` to `ClaimTypeInput`

### JIRA Link

DA-1537

### Checklist

### DIFF
--- a/src/api/entities/types.ts
+++ b/src/api/entities/types.ts
@@ -313,9 +313,11 @@ export interface ClaimScope {
   assetId?: string | undefined;
 }
 
-export type TrustedFor =
-  | Exclude<ClaimType, ClaimType.Custom>
-  | { type: ClaimType.Custom; customClaimTypeId: BigNumber };
+export type CustomClaimWithTypeId = { type: ClaimType.Custom; customClaimTypeId: BigNumber };
+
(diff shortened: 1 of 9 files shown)


src/api/entities/types.ts
```diff
@@ -313,9 +313,11 @@ export interface ClaimScope {
...
 }
 
-export type TrustedFor =
-  | Exclude<ClaimType, ClaimType.Custom>
-  | { type: ClaimType.Custom; customClaimTypeId: BigNumber };
+export type CustomClaimWithTypeId = { type: ClaimType.Custom; customClaimTypeId: BigNumber };
+
+export type TrustedFor = Exclude<ClaimType, ClaimType.Custom> | CustomClaimWithTypeId;
+
+export type ClaimTypeInput = TrustedFor;
```
prashantasdeveloper, line 320: I have duplicated the type so that its clearer on what is expected as input. `TrustedFor` feels constraint around a single use case.

VictorVicente (approved):

PR #1561, opened 2025-09-29
🚀 feat: Implement Ephemeral Next Major Release Integration System

mpastecki:
This PR introduces a comprehensive system for managing breaking changes through ephemeral
  integration branches, enabling automated preview releases for the next major version of the Polymesh
   SDK.

  📋 Overview

  Implements an automated workflow to manage breaking changes using numbered feature branches
  (bc-<number>-<description>) that are automatically discovered, validated, integrated, and published
  as preview releases.

  ✨ Key Features

  🔄 Ephemeral Integration Branches

  - Creates temporary branches for each integration attempt
  - Ensures deterministic, reproducible results
  - Prevents state accumulation and conflicts
  - Automatically cleaned up after use

  ✅ Zero-Tolerance Conflict Prevention

  - Sequential validation before integration
  - Blocks releases if any conflicts detected
  - Comprehensive conflict reporting via GitHub issues
  - Clear resolution steps for developers

  📦 Automated Preview Releases

  - Publishes to npm with next-major distribution tag
  - Semantic versioning (e.g., v30.0.0-next-major.1)
  - Integrates with existing semantic-release infrastructure
  - Preserves original commit authorship

  📁 Changes

  New Files

  - .github/workflows/generate-next-preview.yml - Main integration workflow (609 lines)
  - .github/workflows/validate-bc-branches.yml - Branch naming validation (133 lines)
  - release.config.next-major.js - Semantic-release configuration for previews
  - NEXT_RELEASE.md - Breaking Change workflow guide

  Modified Files

  - .github/workflows/breaking-changes.yml - Added exemption for bc-* branches
  - .github/workflows/auto-rebase.yml - Added bc-* branch detection and handling
  - package.json - Added scripts for preview management

  🔧 Configuration Updates

  Breaking Change Branch Convention

  bc-<number>-<description>
  Examples:
  - bc-1-remove-deprecated-apis
  - bc-2-async-interface-changes
  - bc-10-new-error-handling

  New npm Scripts

  - preview:generate - Manually trigger the integration workflow
  - semantic-release-next - Run semantic-release with next-major config

  🛡️ Security & Quality

  - ✅ Maintains ed25519-sk signing requirements where possible
  - ✅ Uses existing authentication tokens (no new secrets required)
  - ✅ Preserves linear Git history
  - ✅ Original author attribution preserved via cherry-pick
  - ✅ Comprehensive validation before integration

  🔄 Workflow Coordination

  - Auto-rebase workflow pauses when bc-* branches are active
  - Breaking changes check exempts bc-* branches from develop protection
  - Sequential integration ensures predictable branch ordering
  - GitHub issues provide detailed conflict reporting

  📊 Testing

  The system includes:
  - Pre-validation of all branches before integration
  - Automated test execution on integrated code
  - Build validation before release
  - Stale branch detection and reporting

  🚦 Release Process

  1. Developers create bc-* branches for breaking changes
  2. Workflow discovers and orders branches numerically
  3. Sequential validation detects any conflicts
  4. If clean, creates ephemeral integration branch
  5. Cherry-picks all commits preserving authorship
  6. Runs tests and builds the package
  7. Publishes preview release to npm
  8. Creates GitHub release with detailed notes
  9. Cleans up ephemeral resources

  📈 Benefits

  - Predictable Integration - Numbered branches ensure consistent ordering
  - Early Conflict Detection - Prevents broken preview releases
  - Clean Git History - No merge commits or persistent integration branches
  - Developer Experience - Clear naming convention and conflict resolution
  - Automated Publishing - Weekly scheduled runs or manual triggers

  ---
  Breaking Change: None (this PR adds new functionality without modifying existing behavior)


NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+# Edit files, remove deprecated code, etc.
+
+# Stage changes
+git add .
+```
+
+### Step 3: Commit with Proper Format
+Use conventional commits with breaking change indicators:
+```bash
+git commit -m "feat!: remove legacy Widget class
```
polymath-eric, line 62: Why not

```bash
npx git-cz
```
That has support for saying a change is breaking. I'd prefer to keep all the commits using that tool and config for consistency.

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+## 🌿 Branch Naming Convention
+
+### Format
+```
+bc-<number>-<description>
+```
+
+### Rules
+- **Must start with `bc-`**
+- **Include a number** (determines integration order)
```
polymath-eric, line 18: I am not a fan of numbers being chosen by developers like this. Commits should be apliable in arbitrary order for the most part. 

Pragmatically I don't think this will be an issue, but it feels like its asking for collision or coordination. It feels like its adding in a distributed computing problem of sorts.

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+bc-5-new-error-handling         # Changes error handling approach
+bc-10-update-type-definitions   # Updates TypeScript definitions
+```
+
+### Numbering Guidelines
+- **Start from 1** for the first breaking change
+- **Use sequential numbers** for logical ordering
+- **Numbers determine integration order**: bc-1 → bc-2 → bc-5 → bc-10
+- **Avoid conflicts**: Check for existing branches with the same number
+- **Keep under 50** (higher numbers indicate too many breaking changes)
```
polymath-eric, line 35: I think the guideline needing to be written is a good argument against them. It places cognitive load on the user, which is the developer in this case.
```suggestion
```

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+### Basic Structure
+```
+<type>!: <description>
+
+BREAKING CHANGE: <summary>
+
+Breaking-Change-Category: <category>
+Breaking-Change-Impact: <impact-level>
+Breaking-Change-Migration: <migration-instructions>
+Breaking-Change-Affects: <affected-components>
```
polymath-eric: I think commit citizen can be configured with a template. Automation is preferable to documentation

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+- **Interface**: TypeScript interface changes
+- **Types**: Type definition changes
+- **Behavior**: Changed runtime behavior
+- **Dependencies**: Dependency changes affecting consumers
+
+#### Impact Levels
+- **Low**: Affects edge cases or rarely used features
+- **Medium**: Affects common use cases, straightforward migration
+- **High**: Affects core functionality, requires code changes
+- **Critical**: Major architectural changes, complex migration
```
polymath-eric: I think a breaking change is a breaking change. The impact level is too subjective to be useful. I'd prefer not having them.

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+Breaking-Change-Category: API|Interface|Types|Behavior|Dependencies
+Breaking-Change-Impact: Low|Medium|High|Critical
+Breaking-Change-Migration: <clear migration instructions>
+Breaking-Change-Affects: <what components/users are affected>
+```
+
+### Metadata Guidelines
+
+#### Categories
+- **API**: Public API changes (method signatures, new/removed methods)
```
polymath-eric: Aren't all breaking changes API changes? Like the definition of a breaking change is that it changes the API. Its also synonymous with the types, since those help describe the API. A method signature is a type. Even a TypeScript interface is a type of sorts.

Categorizing well is hard, so its best avoided when possible, but I also feel like API captures all of the sub-types under it.

NEXT_RELEASE.md
```diff
@@ -0,0 +1,370 @@
...
+The assets property is now readonly to prevent accidental mutations.
+Use Portfolio.addAsset() and Portfolio.removeAsset() methods instead.
+```
+
+## 🔄 Integration Process
+
+### Automatic Workflow
+1. **Discovery**: System finds all `bc-*` branches
+2. **Validation**: Tests sequential integration for conflicts
+3. **Integration**: Cherry-picks commits in numerical order
```
polymath-eric, line 136: What about commit timestamp as a proxy for the manual branch number?

.github/workflows/generate-next-preview.yml
```diff
@@ -0,0 +1,609 @@
...
+
+          # Fetch all remote branches to ensure they're available locally
+          git fetch --all --prune
+
+          # Find all bc-<number>-* branches and sort numerically
+          BC_BRANCHES=$(git branch -r | grep -E 'origin/bc-[0-9]+-' | sed 's/origin\///' | while read -r branch; do
+            number=$(echo "$branch" | sed -n 's/bc-\([0-9]\+\)-.*/\1/p')
+            if [[ "$number" =~ ^[0-9]+$ ]]; then
+              echo "$number $branch"
+            fi
```
polymath-eric, line 88:
```suggestion
```

Seems like a debug thing to me.

.github/workflows/generate-next-preview.yml
```diff
@@ -0,0 +1,609 @@
...
+
+      - name: Discover breaking change branches
+        id: discovery
+        run: |
+          echo "🔍 Discovering breaking change branches..."
+
+          # Fetch all remote branches to ensure they're available locally
+          git fetch --all --prune
+
+          # Find all bc-<number>-* branches and sort numerically
```
polymath-eric, line 83: I think grabbing the head of each branch and sorting by timestamp is going to be almost the same result, but with a better developer experience.

.github/workflows/generate-next-preview.yml
```diff
@@ -0,0 +1,609 @@
...
+              git cherry-pick "$commit"
+              INTEGRATED_COMMITS="$INTEGRATED_COMMITS $commit"
+              TOTAL_COMMITS=$((TOTAL_COMMITS + 1))
+              echo "✅ Successfully applied: $commit"
+
+              # Parse breaking change metadata if this is a breaking change commit
+              if [[ "$COMMIT_MSG" == *"!"* ]] && [[ "$COMMIT_BODY" == *"BREAKING CHANGE:"* ]]; then
+                # Extract structured metadata
+                CATEGORY=$(echo "$COMMIT_BODY" | grep "Breaking-Change-Category:" | cut -d: -f2- | xargs || echo "")
+                IMPACT=$(echo "$COMMIT_BODY" | grep "Breaking-Change-Impact:" | cut -d: -f2- | xargs || echo "")
```
polymath-eric: I left a comment on the documentation section, but I'd prefer not having these sections.

polymath-eric (commented):
overall lgtm. I did leave some comments and suggestions

polymath-eric:
I am not sure why SonarCloud marked the untouched files as smells in this PR. I marked them as accepted for now since it seems unrelated.

polymath-eric (commented):
The commits should be squashed into one.

The fix commits won't make sense when they get added into the main branch. The user would not have experienced a fix.

NEXT_RELEASE.md
mpastecki: I've updated the documentation. This expects the format as described in https://www.conventionalcommits.org/en/v1.0.0/ 

If the Commitizen supports that, than it will work.

Otherwise I would need to know how does it mark breaking changes and I can add that to be detected.

NEXT_RELEASE.md
mpastecki: My thinking was that even if merges happened in order without conflicts, it doesn't guarantee we won't have conflicts if they're applied out of order.

While commits from a single bc-* branch would still be applied in order, there's a scenario where one breaking change might depend on a previous breaking change that hasn't been merged to develop yet. Applying the former on top of the latter could generate conflicts that would otherwise be avoided.

Additionally, random ordering could render inconsistent results if the workflow runs multiple times.

Having said that, instead of asking developers to manually provide the order in the branch name, we could use the last commit date for each branch and base the ordering on that.

NEXT_RELEASE.md
mpastecki: This expands on the previous discussion https://github.com/PolymeshAssociation/polymesh-sdk/pull/1561#discussion_r2389144595

If ordering based on the date of the last commit in each branch is ok, we definitely can switch to that.

NEXT_RELEASE.md
mpastecki: It's been simplified a lot. There is no need for anything fancy now, just the standard convention when it comes to breaking changes.

NEXT_RELEASE.md
mpastecki: It's been removed.

NEXT_RELEASE.md
mpastecki: This has been removed.

NEXT_RELEASE.md
mpastecki: I am also thinking about this. There are some risks involved though. Let's continue this discussion for now.

.github/workflows/generate-next-preview.yml
mpastecki: It's been removed

.github/workflows/generate-next-preview.yml
mpastecki: With the current ordering this is actually needed for the sort command following.

.github/workflows/generate-next-preview.yml
mpastecki: As I mentioned before, I was considering this approach too.

Then, we would need to decide if we take the first or last commit date as one to use for sorting.

Plus the whole ordering would go nuts if either happens:
- one of the branch is rebased
- if last commit is taken as the date, then minor commit into a branch (a typo fix e.g.) would also break the ordering.

But if we use the first commit as the date, then it could also lead to some issues. As in branch A is started on Monday and branch B on Tuesday.
The work on B is done by Wednesday, but branch A is done only on Friday and already took what was done in B into consideration.

mpastecki:
The whole @NEXT_RELEASE.md was cleared of AI-generated filler.

Plus the requirement when it comes to breaking change detection were simplified to follow the https://www.conventionalcommits.org/en/v1.0.0/

NEXT_RELEASE.md
polymath-eric: I like the idea of less things for devs to think about, but functionally I don't think its a big difference. I am fine with numbers.

.github/workflows/generate-next-preview.yml
polymath-eric: I'd think the head of each branch, since that would give the order of "updated_at DESC" would give a good chance at having a clean merge, since it should be roughly the last time devs touched it. I see your point about rebase changing the order so if you want to leave it as numbers I am OK with that.

VictorVicente (approved):

polymath-eric (approved):

PR #1562, opened 2025-10-02
feat: 🎸 support preparing a POLYX transfer via a multisig signer

F-OBrien:
### Description

The current `transferPolyx` method always checks the balance of the signing account when preparing a POLYX transfer transaction. This results in an `Insufficient free balance` error if the signing key doesn't not have enough POLYX to match the amount being being sent from the MS which prevents the transaction from being later submitted as a multisig proposal.

This PR adds a optional `preRunValidation` function to the `BaseTransactionSpec` and includes it in the transferPolyx method to perform correct validation of signer or multisig balance depending on whether the transaction is run as a proposal or standard transaction.

### Checklist

### DIFF
--- a/src/base/PolymeshTransactionBase.ts
+++ b/src/base/PolymeshTransactionBase.ts
@@ -65,13 +65,14 @@ export abstract class PolymeshTransactionBase<
   public static toTransactionSpec<R, T>(
     transaction: PolymeshTransactionBase<R, T>
   ): BaseTransactionSpec<R, T> {
-    const { resolver, transformer, paidForBy, multiSig } = transaction;
+    const { resolver, transformer, paidForBy, multiSig, preRunValidation } = transaction;
 
     return {
       resolver,
       transformer,
       paidForBy,
       multiSig,
+      preRunValidation,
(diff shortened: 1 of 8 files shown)


src/base/PolymeshTransactionBase.ts
```diff
@@ -243,6 +253,10 @@ export abstract class PolymeshTransactionBase<
     }
 
     try {
+      if (this.preRunValidation) {
+        await this.preRunValidation(true); // asProposal = true
```
polymath-eric:
```suggestion
        await this.preRunValidation({asProposal: true});
```
Object params are a cleaner way to document this. Plus things can be added in with no refactor.

polymath-eric (commented):
Did you check with it in a batch transaction?

F-OBrien:
> Did you check with it in a batch transaction?

@polymath-eric the PR has now been updated to also support `preRunValidation` in `createTransactionBatch` and use object notation for the inputs to preRunValidation

polymath-eric (approved):

PR #1563, opened 2025-10-12
Added `add` and `remove` methods for document management

F-OBrien:
### Description

Adds `add()` and `remove()` methods to the `Documents` namespace for more efficient asset document management. Previously, adding a single document required removing and re-adding all existing documents via `set()`. Now users can perform granular operations:

- **`add()`** - Add documents to existing list
- **`remove()`** - Remove specific documents by ID  
- **`set()`** - Replace all documents (existing behavior)
- **`get()`** - Now returns documents with their on-chain IDs

**Key improvements:**
- Reduced transaction fees for incremental document changes
- `Documents.get()` now returns `ResultSet<AssetDocumentWithId>` instead of `ResultSet<AssetDocument>`. Since `AssetDocumentWithId` extends `AssetDocument`, this is backward compatible for most use cases, but the type signature includes an additional `id` field. The id is required with the `remove()` method
- Document IDs now exposed in `get()` results for use with `remove()`

### Checklist

### DIFF
--- a/src/api/entities/Asset/Base/Documents/index.ts
+++ b/src/api/entities/Asset/Base/Documents/index.ts
@@ -1,13 +1,21 @@
 import { BaseAsset } from '~/api/entities/Asset/Base/BaseAsset';
-import { Context, Namespace, setAssetDocuments } from '~/internal';
 import {
-  AssetDocument,
+  addAssetDocuments,
+  Context,
+  Namespace,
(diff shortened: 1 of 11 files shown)


polymath-eric (approved):

polymath-eric:
I think the failing API snapshot check is a false negative. Adding in the ID to the return type shouldn't be breaking.

PR #1566, opened 2025-10-29
Add balance stat support to setAssetStats

F-OBrien:
### Description

This PR adds support for Balance and `ScopedBalance` stat types in the `setAssetStats` procedure and optimizes the procedure to avoid redundant chain calls.

### Changes

**New Features:**
- ✅ Support for `StatType.Balance` and `StatType.ScopedBalance` stat types
- ✅ Added `balanceStatInputToStatUpdates()` conversion function for balance stats
- ✅ Added `claimBalanceStatInputToStatUpdates()` for scoped balance stats (Accredited, Affiliate, Jurisdiction)
- ✅ All stat value fields (`count`, `balance`, `value`) are now optional

**Optimizations:**
- ✅ Added `prepareStorage()` to fetch current active stats from chain
- ✅ Only calls `SetActiveAssetStats` when the stat list actually changes
- ✅ Only calls `BatchUpdateAssetStats` when stat values are provided
- ✅ Throws `NoDataChange` error when no updates are needed (prevents empty transactions)

**Infrastructure:**
- ✅ Added `areBtreeSetsEqual()` utility to compare BTreeSets by content (ignoring codec metadata like `createdAtHash`)
- ✅ Fixed `createMockBtreeSet()` to properly set `isEmpty` flag for empty arrays
- ✅ Added `AddBalanceStatInput` type following SDK conventions

**Testing:**
- ✅ Comprehensive test coverage (98.6% statements, 100% branches)
- ✅ Tests for all stat types with and without values
- ✅ Tests for stat comparison logic
- ✅ Tests for error handling (unsupported types, no data change)
- ✅ Full coverage of conversion utilities

### Breaking Changes

**Type changes (backward compatible in practice):**
- `count` field in `AddCountStatParams` is now optional
- `value` field in `ClaimCountStatInput` is now optional
- Changed `AddPercentageStatParams` to `AddBalanceStatParams` and added `balance`
- Changed `AddClaimPercentageStatParams` to `AddClaimBalanceStatParams` and added `ClaimBalanceStatInput`

Existing code that provides these values will continue to work unchanged.

### Checklist

### DIFF
--- a/src/api/entities/types.ts
+++ b/src/api/entities/types.ts
@@ -553,7 +553,13 @@ export interface ClaimPercentageTransferRestriction extends TransferRestrictionB
 }
 
 export interface AddCountStatInput {
-  count: BigNumber;
+  /**
+   * The token holder count value for the stat
+   *
+   * @note If not provided when enabling a stat, it will default to zero.
+   *   If not provided when updating stats, the current value will remain unchanged
+   */
+  count?: BigNumber;
 }
 
 export interface StatClaimIssuer {
@@ -565,17 +571,35 @@ export type ClaimCountStatInput =
   | {
       issuer: Identity;
       claimType: ClaimType.Accredited;
-      value: { accredited: BigNumber; nonAccredited: BigNumber };
+      /**
+       * The count values for token holders with the accredited and non-accredited claim
+       *
+       * @note If not provided when enabling a stat, values will default to zero.
+       *   If not provided when updating stats, the current values will remain unchanged
+       */
+      value?: { accredited: BigNumber; nonAccredited: BigNumber };
     }
   | {
       issuer: Identity;
       claimType: ClaimType.Affiliate;
-      value: { affiliate: BigNumber; nonAffiliate: BigNumber };
+      /**
+       * The count values for token holders with the affiliate and non-affiliate claim
+       *
+       * @note If not provided when enabling a stat, values will default to zero.
+       *   If not provided when updating stats, the current values will remain unchanged
+       */
+      value?: { affiliate: BigNumber; nonAffiliate: BigNumber };
     }
   | {
       issuer: Identity;
       claimType: ClaimType.Jurisdiction;
-      value: { countryCode: CountryCode | undefined; count: BigNumber }[];
+      /**
+       * The count values for token holders per jurisdiction claim
+       *
+       * @note If not provided when enabling a stat, values will default to zero.
+       *   If not provided when updating stats, the current values will remain unchanged
+       */
+      value?: { countryCode: CountryCode | undefined; count: BigNumber }[];
     };
 
 export interface ScheduleWithDetails {
--- a/src/api/procedures/setAssetStats.ts
+++ b/src/api/procedures/setAssetStats.ts
@@ -1,16 +1,35 @@
-import { PolymeshPrimitivesStatisticsStatType } from '@polkadot/types/lookup';
+import {
+  PolymeshPrimitivesStatisticsStatType,
+  PolymeshPrimitivesStatisticsStatUpdate,
+} from '@polkadot/types/lookup';
+import { BTreeSet } from '@polkadot/types-codec';
 
-import { Procedure } from '~/internal';
-import { FungibleAsset, SetTransferRestrictionStatParams, StatType, TxTags } from '~/types';
-import { BatchTransactionSpec, ProcedureAuthorization, TxWithArgs } from '~/types/internal';
+import { Context, PolymeshError, Procedure } from '~/internal';
+import {
+  AddBalanceStatParams,
+  AddClaimBalanceStatParams,
+  AddClaimCountStatParams,
+  AddCountStatParams,
+  ErrorCode,
+  FungibleAsset,
+  SetTransferRestrictionStatParams,
+  StatType,
+  TxTag,
+  TxTags,
+} from '~/types';
+import { BatchTransactionSpec, ProcedureAuthorization } from '~/types/internal';
 import {
   assetToMeshAssetId,
+  balanceStatInputToStatUpdates,
+  claimBalanceStatInputToStatUpdates,
   claimCountStatInputToStatUpdates,
   claimIssuerToMeshClaimIssuer,
   countStatInputToStatUpdates,
   statisticsOpTypeToStatType,
   statisticStatTypesToBtreeStatType,
+  statParamsToMeshStatType,
   statTypeToStatOpType,
+  statUpdatesToBtreeStatUpdate,
 } from '~/utils/conversion';
 import { checkTxType } from '~/utils/internal';
 
@@ -22,8 +41,48 @@ export type SetAssetStatParams = { asset: FungibleAsset } & SetTransferRestricti
 /**
  * @hidden
  */
+export interface SetAssetStatsStorage {
+  currentStats: BTreeSet<PolymeshPrimitivesStatisticsStatType>;
+}
+
+/**
+ * @hidden
+ */
+function statToStatUpdates(
+  stat:
+    | AddCountStatParams
+    | AddBalanceStatParams
+    | AddClaimCountStatParams
+    | AddClaimBalanceStatParams,
+  context: Context
+): BTreeSet<PolymeshPrimitivesStatisticsStatUpdate> {
+  const { type } = stat;
+
+  if (type === StatType.Count && stat.count !== undefined) {
+    return countStatInputToStatUpdates(stat, context);
+  }
+
+  if (type === StatType.ScopedCount && stat.value !== undefined) {
+    return claimCountStatInputToStatUpdates(stat, context);
+  }
+
+  if (type === StatType.Balance && stat.balance !== undefined) {
+    return balanceStatInputToStatUpdates(stat, context);
+  }
+
+  if (type === StatType.ScopedBalance && stat.value !== undefined) {
+    return claimBalanceStatInputToStatUpdates(stat, context);
+  }
+
+  // Return empty BTreeSet when no value is provided
+  return statUpdatesToBtreeStatUpdate([], context);
+}
+
+/**
+ * @hidden
+ */
 export function prepareSetAssetStats(
-  this: Procedure<SetAssetStatParams, void>,
+  this: Procedure<SetAssetStatParams, void, SetAssetStatsStorage>,
   args: SetAssetStatParams
 ): Promise<BatchTransactionSpec<void, unknown[][]>> {
   const {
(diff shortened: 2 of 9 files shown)


src/api/entities/types.ts
```diff
@@ -565,17 +565,17 @@ export type ClaimCountStatInput =
   | {
       issuer: Identity;
       claimType: ClaimType.Accredited;
-      value: { accredited: BigNumber; nonAccredited: BigNumber };
+      value?: { accredited: BigNumber; nonAccredited: BigNumber };
```
polymath-eric, line 580: There should be a comment saying what the default value will assume.

src/api/procedures/setAssetStats.ts
```diff
@@ -1,29 +1,98 @@
...
+ * @hidden
+ */
+export interface SetAssetStatsStorage {
+  currentStats: BTreeSet<PolymeshPrimitivesStatisticsStatType>;
+}
+
+/**
+ * @hidden
+ */
+function getStatUpdates(
```
polymath-eric:
```suggestion
function statToStatUpdates(
```
(nit) The word "get" in the name implies a read, but this function is only changing a representation and not actually introducing new information.

src/api/procedures/setAssetStats.ts
```diff
@@ -1,29 +1,98 @@
...
+  stat:
+    | AddCountStatParams
+    | AddBalanceStatParams
+    | AddClaimCountStatParams
+    | AddClaimBalanceStatParams,
+  context: Context
+): BTreeSet<PolymeshPrimitivesStatisticsStatUpdate> | undefined {
```
polymath-eric:
```suggestion
): BTreeSet<PolymeshPrimitivesStatisticsStatUpdate> {
```

Is there a meaningful difference between an empty set and undefined? Returning an empty BTreeSet instead is cleaner if it can be done.

src/api/procedures/setAssetStats.ts
```diff
@@ -1,29 +1,98 @@
...
+    type !== StatType.Count &&
+    type !== StatType.ScopedCount &&
+    type !== StatType.Balance &&
+    type !== StatType.ScopedBalance
+  ) {
+    throw new PolymeshError({
+      code: ErrorCode.UnexpectedError,
+      message: `Unsupported stat type: ${type}. Please report this to the Polymesh team`,
+    });
+  }
```
polymath-eric: Is this check really needed? Feels like something TS will guard against.

Wouldn't this only be triggered if a dev uses JS or an `any` assertion or the like? If type is coming from the user that's on them, so I don't think we'd be able to do much if this error was ever actually seen.

src/api/procedures/setAssetStats.ts
```diff
@@ -33,16 +102,35 @@ export function prepareSetAssetStats(
...
-  const updateTransactions: TxWithArgs<unknown[]>[] = [];
 
-  stats.forEach(stat => {
+  // Build new stats list and compare with current stats
+  const rawStats = stats.map(stat => {
+    const { type } = stat;
+    const inputStatType =
+      type === StatType.ScopedCount || type === StatType.ScopedBalance
+        ? { type, claimIssuer: { issuer: stat.issuer, claimType: stat.claimType } }
+        : { type };
```
polymath-eric: It feels like this logic should be in the `inputStatTypeToMeshStatType` function to me

src/api/procedures/setAssetStats.ts
```diff
@@ -51,63 +139,106 @@ export function prepareSetAssetStats(
...
-          args: [rawAssetId, newStat, statValue],
+          args: [rawAssetId, statType, rawUpdates],
         })
       );
     }
-  });
+  }
 
-  const rawNewStats = statisticStatTypesToBtreeStatType(newStats, context);
+  if (transactions.length === 0) {
```
polymath-eric, line 143: I think this should checked before the transactions so there is one less level of indentation around the transactions.push. e.g.

```ts
if(rawUpdates.isEmpty) { throw "No Data Change" }

transactions.push(...
```

If the "getStatUpdates" returned an empty set instead of `undefined` it would make it a bit easier to follow as well, unless there is a case where the distinction is important I am not seeing.

src/api/procedures/setAssetStats.ts
```diff
@@ -51,63 +139,106 @@ export function prepareSetAssetStats(
...
+      polymeshApi: {
+        query: { statistics },
+      },
+    },
+    context,
+  } = this;
 
-  return Promise.resolve({ transactions, resolver: undefined });
+  const rawAssetId = assetToMeshAssetId(asset, context);
+  const currentStats = await statistics.activeAssetStats(rawAssetId);
```
polymath-eric, line 170: If this was converted into a standardized representation instead of being left as a raw BTreeSet you could avoid needing a `areBTreeSetsEqual` function.

SDK logic should try to use SDK data structures when it can.

src/utils/internal.ts
```diff
@@ -1240,6 +1241,30 @@ export function hasSameElements<T>(
   return !differenceWith(first, second, comparator).length && first.length === second.length;
 }
 
+/**
+ * @hidden
+ *
+ * Compare two BTreeSets by their contents, ignoring codec metadata like createdAtHash and initialU8aLength
+ * which may not be present in a set created in memory and may be present in a set read from the chain.
+ */
+export function areBtreeSetsEqual<T extends Codec>(set1: BTreeSet<T>, set2: BTreeSet<T>): boolean {
```
polymath-eric: I left a comment in the procedure about avoiding this function by converting from the storage.

Why couldn't this be implemented with:

```ts
return set1.eq(set2)
```

If polkadot compare doesn't work, then we should convert into an internal representation and compare between those values.

src/utils/internal.ts
```diff
@@ -1240,6 +1241,30 @@ export function hasSameElements<T>(
...
+    return false;
+  }
+
+  if (set1.isEmpty && set2.isEmpty) {
+    return true;
+  }
+
+  // Compare contents by converting to strings
+  // BTreeSets are sorted, so we can compare element by element
+  const set1Array = [...set1].map(item => item.toString());
```
polymath-eric: BTreeSets can be large, and this is excessive looping and multiples the garbage created.

For an efficiency point of view arrays shouldn't be double iterated and they shouldn't be copied if it can be helped.

The earlier size check + ordering assumption means you could do the needed conversion inside the `.every` loop.

That said, its better if we can avoid writing this function.

src/utils/internal.ts
```diff
@@ -1240,6 +1241,30 @@ export function hasSameElements<T>(
...
+ * Compare two BTreeSets by their contents, ignoring codec metadata like createdAtHash and initialU8aLength
+ * which may not be present in a set created in memory and may be present in a set read from the chain.
+ */
+export function areBtreeSetsEqual<T extends Codec>(set1: BTreeSet<T>, set2: BTreeSet<T>): boolean {
+  // Quick checks
+  if (set1.size !== set2.size) {
+    return false;
+  }
+
+  if (set1.isEmpty && set2.isEmpty) {
```
polymath-eric: Is this actually needed? I thought an empty set would spread into an empty array which `.every` would always return `true` for.

polymath-eric (commented):
Other than the `areBtreeSetsEqual` function it lgtm

src/utils/internal.ts
F-OBrien: My initial attempt was to use the native `.eq` method to compare the BTreeSets but the BTreeSet read from the chain has additional information compared to one created from the type registry so were not returning as equal. I may be able to strip that information by taking the entries only and recreating the type to be able to use the `.eq` method.

F-OBrien: After digging into the .eq method on the BTreeSet codec type it does
```
function compareSetArray (a: Set<unknown>, b: any[]): boolean {
  // equal number of entries and each entry in the array should match
  return (a.size === b.length) && !b.some((e) => !a.has(e));
}
```
The issue is that compareSet uses `.has(e)` to check if elements exist in the set, which relies on JavaScript's `Set.has()` method. For objects, this uses reference equality, not value equality.

So even though the values are the same, they're different object instances, causing `.has()` to return `false`.

src/api/procedures/setAssetStats.ts
F-OBrien: I'm not clear why converting to a SDK type and then doing a comparison would be a better approach? Let's say I was to convert the `currentStats` with `assetStatToStat`, I'd need to also to convert `rawNewStats` and on top of that I'd need to create a function to do the equality comparison anyway. I think the `areBtreeSetsEqual` function is a simple, generic reusable approach.

I'd have preferred to just be able to use the `.eq` method on the BTreeSet but that requires reference equality and not just value equality.

src/api/procedures/setAssetStats.ts
F-OBrien: I added a new function `statParamsToMeshStatType` to included this logic that calls `inputStatTypeToMeshStatType` as it was also used in `createAsset` with transfer restriction inputs.

src/api/procedures/setAssetStats.ts
F-OBrien: if checking earlier we need to also check there are no values being updated before throwing the error and then we would still need to check `!rawUpdates.isEmpty` independently before trying to push the `batchUpdateAssetStats` transaction so I believe it's cleaner to leave as is.

src/api/procedures/setAssetStats.ts
F-OBrien: done

src/utils/internal.ts
F-OBrien: I've refactored the `areBtreeSetsEqual` function to be more efficient.

src/utils/internal.ts
polymath-eric: Thanks for looking into it. I thought it would eq would do value equality for some reason.

I see there is `hash` defined on BTreeSet, I would expect the built in hash value to match there. Does that not work either?

```ts
set1.hash === set2.hash
```
Impl ref: https://github.com/polkadot-js/api/blob/f45dfc72ec320cab7d69f08010c9921d2a21065f/packages/types-codec/src/extended/BTreeSet.ts#L116

src/api/procedures/setAssetStats.ts
polymath-eric: For me its conceptually clearer to write functions that depend on types that aren't from 3rd party libs, but polkadot is a bit of an exception for the SDK.

I agree having a generic way to compare sets directly is useful so that's fine to.

polymath-eric (approved):

PR #1568, opened 2025-11-11
feat: 🎸 create asProposal methods

sansan:
### Description

Adds an optional `asProposal` parameter (default: `true`) to `getTotalFees()` and `toSignablePayload()` methods, allowing flexibility to handle MultiSig transactions as direct transactions rather than proposals.

- **`PolymeshTransactionBase.getTotalFees(asProposal = true)`**

  - When `asProposal: false`, uses the signing account directly for fee calculation (ignores MultiSig payer)
  - When `asProposal: true` (default), maintains existing behavior (uses MultiSig payer if available)

- **`PolymeshTransactionBase.toSignablePayload(metadata, asProposal = true)`**

  - When `asProposal: false`, returns payload with `multiSig: null` and unwrapped transaction
  - When `asProposal: true` (default), maintains existing behavior (includes MultiSig address and wraps as proposal)

- **New helper methods:**

  - `composeTxForFees(asProposal: boolean)` - Conditionally wraps transaction based on `asProposal` flag
  - `getBaseTransaction()` - Abstract method to get unwrapped transaction (implemented in `PolymeshTransaction` and `PolymeshTransactionBatch`)

- **Updated `getPayingAccount(asProposal = true)`**
  - Only uses MultiSig payer logic when `asProposal: true`
  - Falls back gracefully when MultiSig payer's primary account has no identity

### JIRA Link

https://polymesh.atlassian.net/browse/DA-1558

### DIFF
--- a/src/base/types.ts
+++ b/src/base/types.ts
@@ -212,7 +212,8 @@ export interface TransactionPayload {
   /**
    * The address of the MultiSig if the transaction is a proposal.
    *
-   * Will be set only if the signing account is a MultiSig signer and the transaction is not approving or rejecting an existing proposal
+   * Will be set only if the signing account is a MultiSig signer, the transaction is not approving or rejecting an existing proposal,
+   * @note `asProposal: false` will force this to be null, even if the signing account is a MultiSig signer
    */
   readonly multiSig: string | null;
 }
--- a/src/base/PolymeshTransaction.ts
+++ b/src/base/PolymeshTransaction.ts
@@ -104,11 +104,16 @@ export class PolymeshTransaction<
 
   // eslint-disable-next-line require-jsdoc
   protected composeTx(): SubmittableExtrinsic<'promise', ISubmittableResult> {
+    const baseTx = this.getBaseTransaction();
+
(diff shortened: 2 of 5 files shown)


src/base/types.ts
```diff
@@ -212,7 +212,9 @@ export interface TransactionPayload {
   /**
    * The address of the MultiSig if the transaction is a proposal.
    *
-   * Will be set only if the signing account is a MultiSig signer and the transaction is not approving or rejecting an existing proposal
+   * Will be set only if the signing account is a MultiSig signer, the transaction is not approving or rejecting an existing proposal,
+   * and `asProposal: true` was used when calling `toSignablePayload()`. When `asProposal: false`, this will be `null` even if the
+   * signing account is a MultiSig signer.
```
polymath-eric:
```suggestion
  @note `asProposal: false` will force this to be null, even if the signing account is a MultiSig signer
```

(nit)

polymath-eric (dismissed):

polymath-eric (approved):

PR #1594, opened 2026-04-05
feat: 🎸 Add dual version support with 8.x chain

prashantasdeveloper:
### New changes

- `AccountManagement.approveSubsidy` has been added to allow an subsidizer to approve allowance
  amount for a beneficiary account.
- `AccountManagment.acceptSubsidy` has been added to accept an already approved subsidy by the
  beneficiary.
- `AccountManagment.revokeSubsidy` has been added to revoke an already approved subsidy by the
  subsidizer.
- `Assets.transferFunds` now allows to transfer funds between two asset holders (`Account` or
  `Portfolio`) owned by same identity
- `Identities.selfRegisterDid` allows to register a new DID for a signing Account
- Added new ErroCode - `NotSupported` for throwing errors when a specific deprecated method is
  used with incompatible chain version
- Added new methods in Account class to get balances
  - `getAssetBalance` to retrieve the balances of all fungible assets in an Account
  - `getCollections` to retrieve NFTs held in an Account
- Added a new types
  - `AssetHolderLike` which extends `PortfolioLike` with `AccountLike` possible values. This is
    added for convenience to pass either an Account or Portfolio to methods that accept
    an asset holder.
  - `AssetHolder` which is a concrete representation of an asset holder, either an Account or
    Portfolio.
- Adds new methods to Fungble class to support asset allowance
  - `approveAllowance` to Fungible asset class for v8 chain, allowing an entity to approve a
    specific amount of tokens to be consumed by another Account
  - `getAllowance` to retrieve the amount of allowance for owner spender relation
- Adds `setMandatoryReceiverAffirmation` to Identity class. From chain v8, receiver affirmation
  are set to automatic by default. Use this method to change the affirmation requirement to
  `Required` for the receiver asset holder DID to explicitly affirm the instruction for it
  to be executed
- `Subsidies.getPendingSubsidies` has been added to retrieve all pending subsidies for
  acceptance by the beneficiary account. Returns subsidy with allowance details

### Modified types/methods

Modified changes -

- `canTransfer` method of `FungibleSettlements` and `NonFungibleSettlements`
  now supports `AssetHolderLike` param for `from` and `to` params
  instead of `PortfolioLike`.
- `getOwner` method of `Nft` class now returns `AssetHolder` instead of `Portfolio`
- Optional `fromAccount` and `toAccount` has been added to `BaseHistoricAssetTransaction`
  to support fetching Account info for historic asset transaction using middleware.
  This affects `NftCollection.getTransactionHistory` and
  `Fungible.getTransactionHistory`
- `general` attribute `TransferBreakdown` has been changed from `TransferError[]`
  to `(TransferError | string)[]` to support custom errors reported by
  chain as well
- `getInstructions` method of `Identity` class now returns all the instructions
  where the identities either the custodian of one or more portfolios in the
  legs or owns one or more accounts in the legs
- `InstructionAffirmation` type has been modifed to resolve the affirming party
  not only as identity but also accounts. Parameter `identity: Identity`
  has been changed to `party: Account | Identity`.
  This affects `getAffirmations` method of `Instruction` class
- Adds new argument - `expiresAt?: Date;` as mandatory field for generating
  off chain affirmations in `generateOffChainAffirmationReceipt`
  of `Instruction` class
- `from` and `to` parameters of `FungibleLeg` and `NftLeg` now support
  `Account` as valid value and their type has been changed from
  `DefaultPortfolio | NumberedPortfolio` to `AssetHolder`
- Procedure type changes
  - RoleType.CddProvider has been deprecated in favour of RoleType.DidRegistrar
  - `AcceptPrimaryKeyRotationParams.cddAuth` is no longer needed from v8
  - `IssueTokenParams` now has an optional `account` field which is used to
    specify an account from which tokens should be issued. This field should
    be provided only if `portfolioId` is not provided
  - `Withdraw` and `WithdrawAsMediator` values of `InstructionAffirmationOperation`
    have been deprecated for chain v8
  - `portfolio` param in `RejectInstructionParams` has been changed to `assetHolder`
    of type `AssetHolderLike`
  - `portfolios` param in `WithdrawInstructionParams` has been changed to `holders`
    of type `AssetHolderLike[]`. Note that withdrawing has been discontinued from v8
  - `portfolios` param in `AffirmInstructionParams` has been changed to `holders`
    of type `AssetHolderLike[]`
  - New optional field `destination` of type `AssetHolderLike` has been added to
    `ControllerTransferParams` to support controller transferring tokens to either
    account or portfolio
  - `destinationPortfolio` param of `NftControllerTransferParams` has been marked
    deprecated in favour of new optional field `destination` of type `AssetHolderLike`
    to support controller transferring NFTs to either account or portfolio
  - `fromAccount` has been added to `RedeemTokensParams` to allow redeeming tokens
    from an accountd
  - `fromAccount` has been added to `RedeemNftParams` to allow redeeming Nfts
    from an account

### Deprecated

Few features are no longer supported in chain v8.

Deprecated changes -

- `AccountManagement.subsidizeAccount` has been deprecated in favour of
  `AccountManagement.approveSubsidy`
- Cdd claims are not longer support in chain v8. Deprecated methods -
  - `Claims.getCddClaims`
  - `hasValidCdd` of `Identity` class
- Child identities have been discontinued. Deprecated methods
  - `Identities.getChildIdentity`
  - `Identities.createChild`
  - All methods of `ChildIdentity`
- Withdrawing affirmation has been discontinued and following methods are now deprecated
  - `withdraw`, `withdrawAsMediator` in `Instruction` class

### Checklist


src/api/entities/Account/index.ts
```diff
@@ -655,4 +669,174 @@ export class Account extends Entity<UniqueIdentifiers, string> {
...
+          locked: new BigNumber(0),
+          free: new BigNumber(0),
+          asset: argAsset,
+        };
+
+        filteredBalances.push(assetBalances[argAsset.id] ?? portfolioBalance);
+      }
+
+      return filteredBalances;
+    }
```
F-OBrien: This works and I see it is the same pattern as used by the Portfolio getAssetBalances but would it not be more efficient to only query the entries passed in args when present instead of querying all and then filtering?
e.g. add the following after `const rawAccountId = stringToAccountId(address, context);`
```
    if (args?.assets.length) {
      const queriedAssets = await Promise.all(
        args.assets.map(assetToQuery => asFungibleAsset(assetToQuery, context))
      );
      const multiParams = queriedAssets.map(({ id }) =>
        tuple(rawAccountId, stringToAssetId(id, context))
      );

      const [totalBalances, lockedBalances] = await Promise.all([
        asset.assetBalance.multi(multiParams),
        asset.lockedBalance.multi(multiParams),
      ]);

      return queriedAssets.map((queriedAsset, index) => {
        const total = balanceToBigNumber(totalBalances[index]!);
        const locked = balanceToBigNumber(lockedBalances[index]!);

        return {
          asset: queriedAsset,
          total,
          locked,
          free: total.minus(locked),
        };
      });
    }
```
and delete
```
    if (args?.assets.length) {
      const filteredBalances: PortfolioBalance[] = [];
      for (const asset of args.assets) {
        const argAsset = await asFungibleAsset(asset, context);
        const portfolioBalance = {
          total: new BigNumber(0),
          locked: new BigNumber(0),
          free: new BigNumber(0),
          asset: argAsset,
        };

        filteredBalances.push(assetBalances[argAsset.id] ?? portfolioBalance);
      }

      return filteredBalances;
    }
```

codegen.yaml
```diff
@@ -1,5 +1,5 @@
 overwrite: true
-schema: 'http://localhost:3001'
+schema: 'https://dev.polymesh.tech/playground/testnet'
```
F-OBrien: Was this intentionally committed?

src/utils/conversion.ts
```diff
@@ -858,6 +914,57 @@ export function portfolioIdToMeshPortfolioId(
...
 }
 
+/**
+ * @hidden
+ */
+export async function assetHolderIdToMeshAssetHolder(
+  assetHolderId: AssetHolderId,
+  context: Context
+): Promise<PolymeshPrimitivesAssetAssetHolder> {
+  context;
```
F-OBrien: delete

src/api/entities/Identity/index.ts
```diff
@@ -160,14 +159,13 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
-        },
+        polymeshApi: { query },
       } = context;
 
-      const activeMembers = await cddServiceProviders.activeMembers();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const activeMembers = await (query as any).cddServiceProviders.activeMembers();
       const memberDids = activeMembers.map(identityIdToString);
 
       return memberDids.includes(did);
```
F-OBrien: In v8 the CddProviderRole is replaced with a didRegistrar role and the cddServiceProviders pallet is renamed to didRegistrars. We should support the role natively. If possibly we should deprecate the CddProviderRole for v8 or use it as an alias for didRegistrars in the dual compatible release

src/api/entities/Identity/index.ts
```diff
@@ -296,7 +296,14 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
+    if (!context.isV7) {
+      throw new PolymeshError({
+        code: ErrorCode.General,
+        message: 'CDD claims are discontinued from chain v8',
+      });
+    }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const result = await (call.identityApi as any).isIdentityHasValidCdd(identityId, null);
     return cddStatusToBoolean(result);
   }
```
F-OBrien, line 321: for the dual compatibility release should this just return `true` if the DID exists? Having it throw and error only when called may mask the fact it is broken until called as it won't error at compile time

src/api/entities/Identity/index.ts
```diff
@@ -594,7 +608,10 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
   public async getInstructions(): Promise<GroupedInstructions> {
     const { did, portfolios } = this;
 
-    const ownedPortfolios = await portfolios.getPortfolios();
+    const [ownedPortfolios, secondaryPermissionedAccounts] = await Promise.all([
+      portfolios.getPortfolios(),
+      this.getSecondaryAccounts(),
+    ]);
```
F-OBrien: The `getInstructions()` method collects secondary accounts as account-level holders for the `assembleGroupedInstructions` call, but omits the **primary key** of the Identity. Any instruction where the primary account is a direct participant will be silently missed.

**Suggested fix** — include the primary account alongside secondary accounts:

```typescript
const [ownedPortfolios, secondaryPermissionedAccounts, primaryAccount] =
  await Promise.all([
    portfolios.getPortfolios(),
    this.getSecondaryAccounts(),
    this.getPrimaryAccount(),
  ]);
```

The same gap exists in `getInvolvedInstructions()` — the primary account is also omitted there when building the `ownedPortfolios`/`secondaryAccounts` holder list passed to `assembleGroupedInstructions`.

src/api/entities/Identity/index.ts
```diff
@@ -692,10 +715,12 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
-    const [allPortfolios, { data: custodiedPortfolios }] = await Promise.all([
-      portfolios.getPortfolios(),
-      portfolios.getCustodiedPortfolios(),
-    ]);
+    const [allPortfolios, { data: custodiedPortfolios }, { data: secondaryPermissionedAccounts }] =
+      await Promise.all([
+        portfolios.getPortfolios(),
+        portfolios.getCustodiedPortfolios(),
+        this.getSecondaryAccounts(),
+      ]);
```
F-OBrien: a per previous comment - primary account is missing

src/api/entities/Identity/index.ts
```diff
@@ -323,14 +330,21 @@ export class Identity extends Entity<UniqueIdentifiers, string> {
...
-          query: { cddServiceProviders },
-        },
+        polymeshApi: { query },
       },
+      context,
       did,
     } = this;
 
-    const activeMembers = await cddServiceProviders.activeMembers();
+    if (context.isV7) {
```
F-OBrien: The logic here is backwards. it should be `if (!context.isV7)`

src/api/entities/Identity/index.ts
F-OBrien: v8 queries to CddProvider role will just fall through to
```
    throw new PolymeshError({
      message: `Unrecognized role "${JSON.stringify(role)}"`,
    });
```

src/api/client/AccountManagement.ts
```diff
@@ -187,9 +191,20 @@ export class AccountManagement {
...
   public subsidizeAccount: ProcedureMethod<SubsidizeAccountParams, AuthorizationRequest>;
 
+  /**
+   * Approves a subsidy request
+   *
+   * @note to be used from chain v8 onwards
+   */
+  public approveSubsidy: ProcedureMethod<SubsidizeAccountParams, void>;
+
+  // TODO add support for accept & revoke subsidy
```
F-OBrien: I believe we also need a method for getting pending subsidies for acceptance

F-OBrien: and to remove an already approved/active subsidy be either the payer or payee

src/api/procedures/utils.ts
```diff
@@ -168,6 +176,22 @@ export async function assertPortfolioExists(
...
+  context: Context
+): Promise<void> {
+  if (typeof assetHolderId === 'string') {
+    if (hexHasPrefix(assetHolderId)) {
+      await assertPortfolioExists({ did: assetHolderId }, context);
+    }
+  } else {
+    await assertPortfolioExists(assetHolderId, context);
+  }
+}
```
F-OBrien, line 188: I don't love the name of this assert. If the `assetHolder` is a string (either a DID or an Account ID) then this is a no-op check as `assertPortfolioExists` only checks for existence of numbered portfolios. Default portfolios, DID strings and Account rely on checking if they have a valid DID for existence checks.

A more accurate but long winded name might be `assertNumberedPortfolioExistsForAssetHolder`

and the function can be simplified to
```
if (typeof assetHolderId !== 'string') {
  await assertPortfolioExists(assetHolderId, context);
}
```

src/api/procedures/utils.ts
```diff
@@ -495,11 +519,10 @@ export async function assertMultiSigSignerAuthorizationValid(
...
-  data: AddRelayerPayingKeyAuthorizationData
+export async function assertOldAddRelayerPayingKeyAuthorizationValid(
+  subsidy: SubsidyData,
+  context: Context
 ): Promise<void> {
-  const subsidy = data.value;
-
   const [beneficiaryIdentity, subsidizerIdentity] = await Promise.all([
     subsidy.beneficiary.getIdentity(),
     subsidy.subsidizer.getIdentity(),
```
F-OBrien, line 527: I thought with v8 we no longer required a key to have a DID to be able to be subsidized or subsidize? This requirement may be specific to legacy migrated authorizations? otherwise the entire method for a v8 chain should be no-op

src/api/entities/Asset/Base/Settlements/index.ts
```diff
@@ -110,37 +117,50 @@ class BaseSettlements<T extends BaseAsset> extends Namespace<T> {
...
+      getAssetHolderDid(from, context),
+      getAssetHolderDid(to, context),
+    ]);
+
+    if (!fromDid || !toDid) {
+      throw new PolymeshError({
+        code: ErrorCode.UnmetPrerequisite,
+        message: 'No DID associated with either from or to account holder',
+      });
+    }
```
F-OBrien, line 159: nit: we could give a slightly more granular error so the user know which (from/to/both) is missing the DID

src/api/entities/Asset/NonFungible/Nft.ts
```diff
@@ -258,10 +262,19 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
     }
```
F-OBrien, line 259: typo in error message `exists` instead of `exist`
```
message: 'NFT does not exists. The token may have been redeemed',
```

src/api/procedures/types.ts
```diff
@@ -1243,8 +1257,14 @@ export interface ControllerTransferParams {
    * amount of Asset tokens to transfer
    */
   amount: BigNumber;
+
+  /**
+   * (optional) portfolio (or portfolio ID) or account to which Assets will be transferred to. Defaults to default. If specified it must be one of the callers own portfolios or accounts
+   */
+  destination?: AssetHolderLike;
```
F-OBrien, line 1292: The chain doesn't currently support this for fungible  - I've raise with the Core team so hopefully they add it before v8

src/api/procedures/types.ts
```diff
@@ -1243,8 +1257,14 @@ export interface ControllerTransferParams {
...
 }
 
+// TODO @prashantasdeveloper add support of controller transfer from account
```
F-OBrien: The destination should also optionally be an account

src/api/procedures/controllerTransfer.ts
```diff
@@ -34,16 +37,18 @@ export async function prepareControllerTransfer(
   } = this;
   const { asset, originPortfolio, amount } = args;
```
F-OBrien, line 38: `destination` is part of the args but we don't handle it in any way - chain is missing support for destination for fungible assets. There is a Jira task to add it

src/api/procedures/redeemTokens.ts
```diff
@@ -75,14 +77,20 @@ export function getAuthorization(
...
 ): ProcedureAuthorization {
   const {
-    storage: { fromPortfolio },
+    storage: { fromAssetHolder },
+    context,
   } = this;
 
+  let transaction = TxTags.asset.Redeem;
+  if (context.isV7 && from) {
+    transaction = TxTags.asset.RedeemFromPortfolio;
```
F-OBrien: There is no redeemFromPortfolio extrinsic in either v7.4 or v8. I'm not clear why we are adding this here? `redeem_from_portfolio` was removed in v7.0

src/api/procedures/executeManualInstruction.ts
```diff
@@ -61,7 +68,7 @@ export async function prepareExecuteManualInstruction(
 
   await assertInstructionValidForManualExecution(instructionDetails, context);
 
-  if (!portfolios.length) {
+  if (!allowedAssetHolders.length) {
```
F-OBrien, line 65: I think we have a gap here as Mediators can also execute an instruction, not just sender, receiver, custodians and venue owners

src/api/procedures/addInstruction.ts
```diff
@@ -701,22 +727,28 @@ async function getTxArgsAndErrors(
       const rawLegValues = await Promise.all([
         ...legs.fungibleLegs.map(async leg => await mapFungibleLeg(leg, context)),
         ...legs.nftLegs.map(async leg => await mapNftLeg(leg, context)),
-        ...legs.offChainLegs.map(leg => mapOffChainLeg(leg, context)),
+        ...legs.offChainLegs.map(async leg => await Promise.resolve(mapOffChainLeg(leg, context))),
       ]);
```
F-OBrien: the async await seems redundant here but harmless

src/api/procedures/addInstruction.ts
```diff
@@ -573,19 +592,26 @@ async function validateInstruction(
...
+      const fromId = assetHolderLikeToAssetHolderId(from);
+      const toId = assetHolderLikeToAssetHolderId(to);
+
+      if (typeof fromId === 'string' && typeof toId === 'string') {
+        return fromId === toId;
+      }
+      if (typeof fromId !== typeof toId) {
+        return false;
+      }
+      return (fromId as PortfolioId).did === (toId as PortfolioId).did;
```
F-OBrien: There is a gap here - If the the sender is an Account and the Receiver is a Portfolio then they could still be from the same DID. I think it is more correct to always get and compare the DIDs for both Accounts and Portfolios. 

I've not validated this so maybe check with the Core team if transfers between accounts/Portfolios under the same DID are allowed for both v7.4 and v8. In v8 there will be a new transfer function that is appropriate for self transfers.

src/api/procedures/addInstruction.ts
```diff
@@ -854,26 +891,41 @@ export async function prepareStorage(
 
   const identity = await context.getSigningIdentity();
 
+  const checkCustodiedPortfolio = async (
+    assetHolder: AssetHolder
+  ): Promise<DefaultPortfolio | NumberedPortfolio | null> => {
+    if (assetHolder instanceof Account) {
+      return null;
+    }
```
F-OBrien, line 903: This means that accounts will never be in `assetHoldersToAffirm`. Should we be checking if an Account is under the same DID? May need to check with the Core team as I'm not 100% sure if accounts must be affirmed by the the account key only or any permissioned key under the DID can affirm for it. The chain accepts a `holderSet` in v8 and a Portfolio BTreeSet in v7.4 which accepted an Account as a portfolio kind.

src/api/procedures/addInstruction.ts
```diff
@@ -162,24 +166,32 @@ async function mapFungibleLeg(
...
+  ];
+
+  if (context.isV7) {
+    if (typeof fromId !== 'string') {
+      assertPromises.push(assertValidCdd(fromId.did, context));
+    }
+    if (typeof toId !== 'string') {
+      assertPromises.push(assertValidCdd(toId.did, context));
+    }
+  }
```
F-OBrien, line 209: Should we be checking valid CDD for a DID string and DID linked to an Account for v7? This is a minor comment as the chain will catch missing CDD at instruction execution and not many people will be using Account at this stage.

src/api/procedures/addInstruction.ts
```diff
@@ -196,25 +208,32 @@ async function mapNftLeg(
...
```
F-OBrien: Same as previous comment here

src/api/procedures/addInstruction.ts
```diff
@@ -854,26 +891,41 @@ export async function prepareStorage(
...
+
+    const isCustodied = await assetHolder.isCustodiedBy({ identity });
+    if (isCustodied) {
+      return assetHolder;
+    }
+
+    return null;
+  };
+
   const portfoliosToAffirm = await Promise.all(
```
F-OBrien: The name is correct as we filter out any accounts but we change it to `assetHoldersToAffirm` when we return it and then later use it as `portfoliosToAffirm `again. If accounts don't need explicit inclusion in affirmation we should keep the name as portfolios so it is clear accounts are omitted.If account s should be included we should stick to `assetHoldersToAffirm`

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +639,108 @@ export function getAuthorization(
...
-      const isCustodied = await legPortfolio.isCustodiedBy({ identity: signingDid });
+    if (legAssetHolder instanceof Account) {
+      // TODO assuming that the account address should be the same as signing did
+      const identity = await legAssetHolder.getIdentity();
+      if (identity?.did === signingDid) {
+        res = [...res, legAssetHolder];
+        if (sender) {
+          senderLegCount = senderLegCount.plus(1);
+        }
+      }
```
F-OBrien, line 721: Account asset holders do not behave the same as portfolios for affirmations. Currently only the account signing key can affirm for Account Asset Holders. There is no permissioning ofr primary/secondary keys to affirm on behalf of an account so the same DID check here is not correct. We would need to bbe checking the signing key is the account holder to include that assetHolder

We should resolve the signing account address from context and compare it directly to the Account asset holder address.
Keep DID/custody logic only for portfolio-style holders.

Also probably worth updating the docs for impacted methods so it is clear Account can only affirm for themselves

src/api/procedures/addInstruction.ts
F-OBrien: Correction: this should check the signer is the Account holder as only an account based holder can affirm for their own account

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -14,18 +16,20 @@ import {
...
+  AssetHolderId,
   DefaultPortfolio,
   ErrorCode,
   InstructionAffirmationOperation,
   ModifyInstructionAffirmationParams,
   NumberedPortfolio,
   OffChainAffirmationReceipt,
   PortfolioId,
-  PortfolioLike,
+  PortfolioLike as AssetHolderLike,
```
F-OBrien: Why are we renaming `PortfolioLike` to `AssetHolderLike`. If there isn't a good reason I think we should import `AssetHolderLike` directly instead and that should allow us to remove the other typecasts to `as AssetHolderLike`

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +639,108 @@ export function getAuthorization(
...
-function extractPortfolioParams(params: ModifyInstructionAffirmationParams): PortfolioLike[] {
+function extractAssetHolderParams(params: ModifyInstructionAffirmationParams): AssetHolderLike[] {
   const { operation } = params;
-  let portfolioParams: PortfolioLike[] = [];
+  let assetHolderParams: AssetHolderLike[] = [];
   if (operation === InstructionAffirmationOperation.Reject) {
     const { portfolio } = params;
     if (portfolio) {
-      portfolioParams.push(portfolio);
+      assetHolderParams.push(portfolio as AssetHolderLike);
```
F-OBrien: portfolio is already a type of `AssetHolderLike`

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +639,108 @@ export function getAuthorization(
...
     }
   } else if (
     operation === InstructionAffirmationOperation.Affirm ||
     operation === InstructionAffirmationOperation.Withdraw
   ) {
     const { portfolios } = params;
     if (portfolios) {
-      portfolioParams = [...portfolioParams, ...portfolios];
+      assetHolderParams = [...assetHolderParams, ...(portfolios as AssetHolderLike[])];
```
F-OBrien: portfolio is already a type of `AssetHolderLike`

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -14,18 +16,20 @@ import {
...
+import { Account, Context, Identity, Instruction, PolymeshError, Procedure } from '~/internal';
 import {
   AffirmationStatus,
   AffirmInstructionParams,
+  AssetHolder,
+  AssetHolderId,
   DefaultPortfolio,
   ErrorCode,
   InstructionAffirmationOperation,
   ModifyInstructionAffirmationParams,
```
F-OBrien, line 29: nit: I think the reject as mediators case seems to be duplicated incorrectly in this type. No parameters are required in either v7 or v8 when rejceting as a mediator

```
export type ModifyInstructionAffirmationParams = InstructionIdParams &
  (
    | ({
        operation: InstructionAffirmationOperation.Affirm;
      } & AffirmInstructionParams)
    | ({
        operation: InstructionAffirmationOperation.Withdraw;
      } & WithdrawInstructionParams)
    | ({
        operation:
          | InstructionAffirmationOperation.Reject
          | InstructionAffirmationOperation.RejectAsMediator; // This is a duplicate
      } & RejectInstructionParams)
    | ({
        operation: InstructionAffirmationOperation.AffirmAsMediator;
      } & AffirmAsMediatorParams)
    | {
        operation:
          | InstructionAffirmationOperation.WithdrawAsMediator
          | InstructionAffirmationOperation.RejectAsMediator; // Keep this
      }
  );
```

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -82,27 +87,28 @@ export interface Storage {
...
+    // since no asset holders are involved in these operations, consider them as valid
     return;
   }
 
-  if (portfolioParams.length && portfolioParams.length !== portfolios.length) {
+  if (assetHolderLikeParams.length && assetHolderLikeParams.length !== assetHolders.length) {
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
-      message: 'Some of the portfolios are not a involved in this instruction',
+      message: 'Some of the asset holders are not a involved in this instruction',
```
F-OBrien, line 106: typo in error message

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -82,27 +87,28 @@ export interface Storage {
...
     });
   }
 
-  if (!portfolios.length) {
+  if (!assetHolders.length) {
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
       message: 'The signing Identity is not involved in this Instruction',
```
F-OBrien: With account based signing we should possible consider making this error more generic e.g. `The signer is not involved in this instruction`. We can have a case where the Account is under the same identity as the signer but Account can only affirm their own legs so saying the signing identity is not involved may be misleading.

src/api/entities/Asset/NonFungible/Nft.ts
```diff
@@ -253,15 +254,25 @@ export class Nft extends Entity<NftUniqueIdentifiers, HumanReadable> {
...
 
     const rawAssetId = assetToMeshAssetId(collection, context);
+    const rawNftId = bigNumberToU64(id, context);
+
+    if (owner instanceof Account) {
+      const rawLocked = await nft.nftHolder(
+        stringToAccountId(owner.address, context),
+        rawAssetId,
+        rawNftId
+      );
```
F-OBrien: for v7 chains we still need the old double map query type 
```
      const rawLocked = await nft.nftHolder(
        stringToAccountId(owner.address, context),
        [
          rawAssetId,
          rawNftId,
        ] 
      );
```

src/base/Context.ts
```diff
@@ -493,6 +493,46 @@ export class Context {
     return assembleResult(subsidies);
   }
 
+  /**
+   * @hidden
+   *
+   * Retrive all pending subsidies for approval
```
F-OBrien: typo - Retrieve

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -476,7 +495,8 @@ export async function prepareModifyInstructionAffirmation(
       validateMediatorStatusForWithdrawl(mediatorStatus, signer, id);
 
       return {
-        transaction: settlementTx.withdrawAffirmationAsMediator,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        transaction: (settlementTx as any).withdrawAffirmationAsMediator,
         resolver: instruction,
         args: [rawInstructionId],
       };
```
F-OBrien: this also needs a isV7 check similar to other withdrawals
```
      if (!context.isV7) {
        throw new PolymeshError({
          code: ErrorCode.UnmetPrerequisite,
          message: 'Withdrawal of affirmed instructions has been discontinued from v8 chain',
        });
      }
```

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +644,134 @@ export function getAuthorization(
...
+      return { did: id, number: '0' };
+    }
+    return {
+      did: id.did,
+      number: !id.number || id.number.isZero() ? '0' : id.number.toString(),
+    };
+  };
+
+  const normalizedLegId = getNormalizedId(
+    (legAssetHolderId as AssetHolderId) ?? {
```
F-OBrien: unnecessary typecast

src/api/procedures/nftControllerTransfer.ts
```diff
@@ -40,31 +46,32 @@ export async function prepareNftControllerTransfer(
...
   }
 
-  if (did !== destinationPortfolio.owner.did) {
+  const destinationDid = await getAssetHolderDid(destinationAssetHolder, context);
+
+  if (did !== destinationDid) {
     throw new PolymeshError({
       code: ErrorCode.UnmetPrerequisite,
-      message: "Controller transfer must send to one of the signer's portfolios",
+      message: "Controller transfer must send to one of the signer's portfolios or accouts",
```
F-OBrien: typo - "accounts"

src/api/procedures/createChildIdentity.ts
```diff
@@ -50,6 +53,13 @@ export async function prepareCreateChildIdentity(
     },
   } = this;
 
+  if (!context.isV7) {
+    throw new PolymeshError({
+      code: ErrorCode.NotSupported,
+      message: 'Child Identiteis are no longer supported in v8',
```
F-OBrien: Typo "Identities"

src/api/procedures/types.ts
F-OBrien, line 1: We should deprecate the `CddProvider` RoleType in favor of `DidRegistrar`

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +644,134 @@ export function getAuthorization(
...
     }
   }
-  return portfolioParams;
+  return assetHolderParams;
 }
 
 /**
  * @hidden
  */
 export const isParam = (
```
F-OBrien, line 663: I found this hard to reason. `AssetHolder` is an account or Portfolio and `AssetHolderId` is a string or Portfolio. Normalizing Account or strings to `{did: string, number: '0'}` is misleading as the string is really an address. Could we make the account vs portfolio cases explicit here, or at least rename the normalized fields to something neutral like ownerOrAddress instead of did?

It works as both are normalized the same way I just found it confusing on first review.

src/api/procedures/modifyInstructionAffirmation.ts
```diff
@@ -614,92 +644,134 @@ export function getAuthorization(
...
+): Promise<[AssetHolder[], BigNumber]> => {
   const [fromExists, toExists] = await Promise.all([from.exists(), to.exists()]);
 
-  const [custodiedPortfolios, amount] = result;
+  const [custodiedAssetHolders, count] = result;
 
-  let res = [...custodiedPortfolios];
-  let legAmount = amount;
+  const res = [...custodiedAssetHolders];
+  const senderLegCount = count;
```
F-OBrien: `result` is always set to an empty array and we don't use `res` or `senderLegCount` so we should be able to clean up to remove

src/api/procedures/types.ts
F-OBrien: `Withdraw` and `WithdrawAsMediator` should also be marked as deprecated
```
export enum InstructionAffirmationOperation {
  Affirm = 'Affirm',
  Withdraw = 'Withdraw',
  Reject = 'Reject',
  AffirmAsMediator = 'AffirmAsMediator',
  WithdrawAsMediator = 'WithdrawAsMediator',
  RejectAsMediator = 'RejectAsMediator',
}
```

F-OBrien (commented):
I still need to review the updates to addInstructions and conversions but here are some initial comments.

src/api/procedures/modifyInstructionAffirmation.ts
prashantasdeveloper: I have changed the logic to directly compare asset holders.

src/api/procedures/types.ts
prashantasdeveloper: Marked both as deprecated

src/api/procedures/modifyInstructionAffirmation.ts
prashantasdeveloper: No longer applicable

src/api/procedures/modifyInstructionAffirmation.ts
prashantasdeveloper: I have updated the logic, doc updates are pending.

src/api/procedures/approveAllowance.ts
```diff
@@ -0,0 +1,79 @@
...
+  } = this;
+
+  const { asset, amount, spender } = args;
+
+  if (amount.lt(0)) {
+    throw new PolymeshError({
+      code: ErrorCode.UnmetPrerequisite,
+      message: 'Amount must be greater than 0',
+    });
+  }
```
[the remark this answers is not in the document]
prashantasdeveloper: Changed the message

src/api/procedures/approveAllowance.ts
```diff
@@ -0,0 +1,79 @@
+import { FungibleAsset, PolymeshError, Procedure } from '~/internal';
+import { ApproveAllowanceParams, AssetHolder, ErrorCode, TxTags } from '~/types';
+import { ExtrinsicParams, ProcedureAuthorization, TransactionSpec } from '~/types/internal';
+import { assetToMeshAssetId, bigNumberToBalance, stringToAccountId } from '~/utils/conversion';
+import { asAccount } from '~/utils/internal';
+
+export interface Storage {
+  did: string;
+  destinationAssetHolder: AssetHolder;
+}
```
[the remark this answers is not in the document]
prashantasdeveloper: Removed this

src/api/procedures/setMandatoryReceiverAffirmation.ts
```diff
@@ -0,0 +1,114 @@
...
+      data: { identity: did },
+    });
+  }
+
+  const rawRequirement = affirmationRequirementToMesh(requirement, context);
+
+  return {
+    transaction: tx.settlement.setMandatoryReceiverAffirmation,
+    args: [rawRequirement],
+    resolver: undefined,
```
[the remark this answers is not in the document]
prashantasdeveloper: No change required here. Check is to restrict setting mandatory confirmation for any other DID. by default extrinsic works for calling DID

src/api/procedures/approveAllowance.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @hidden
+ */
+export function getAuthorization(
+  this: Procedure<Params, void>,
+  { asset }: Params
+): ProcedureAuthorization {
+  return {
+    permissions: {
+      assets: [asset],
```
F-OBrien: The transaction is not restricted by asset permissions on chain

src/api/client/Assets.ts
```diff
@@ -507,4 +516,13 @@ export class Assets {
...
   }
+
+  /**
+   * Transfer funds between two asset holders (Account or Portfolio) owned by same identity.
+   *
+   * @note When `from` account is of type account and the caller is the subsidizer of `from` account, there should be allowance available for transfer and for each transfer said amount is decremented.
+   *
+   * @note To transfer between asset holders owned by separate DID use settlement instructions
+   */
+  public transferFunds: ProcedureMethod<TransferFundsParams, void>;
```
F-OBrien, line 527: This should allow sending to other identities also. Low priority can be changed in a future release

src/api/procedures/transferFunds.ts
```diff
@@ -0,0 +1,221 @@
...
+    });
+  }
+
+  if (fromDid !== toDid) {
+    throw new PolymeshError({
+      code: ErrorCode.ValidationError,
+      message:
+        'For transferring funds between different DIDs, use `Settlements.addInstruction` method instead.',
+    });
+  }
```
F-OBrien, line 150: We should allow transfers to other identities. Low priority update

F-OBrien (approved):

PR #1599, opened 2026-05-18
### Implement mandatory receiver affirmation procedures

F-OBrien:
### Checklist

### DIFF
--- a/src/api/entities/Instruction/types.ts
+++ b/src/api/entities/Instruction/types.ts
@@ -100,6 +100,11 @@ export enum AffirmationStatus {
   Rejected = 'Rejected',
 }
 
+export enum ReceiverAffirmationRequirement {
+  Automatic = 'Automatic',
+  Required = 'Required',
+}
+
 export interface InstructionAffirmation {
   party: Identity | Account;
   status: AffirmationStatus;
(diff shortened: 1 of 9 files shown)


src/api/entities/Instruction/types.ts
```diff
@@ -100,6 +100,11 @@ export enum AffirmationStatus {
   Rejected = 'Rejected',
 }
 
+export enum AffirmationRequirement {
```
prashantasdeveloper:
```suggestion
export enum ReceiverAffirmationRequirement {
```

(nit) Should we make it more explicit?

src/api/procedures/setMandatoryReceiverAffirmation.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (context.isV7) {
+    throw new PolymeshError({
+      code: ErrorCode.NotSupported,
+      message: 'setMandatoryReceiverAffirmation is not supported on v7 chains',
+    });
+  }
+
+  const identity = await context.getSigningIdentity();
+
+  if (identity.did !== did) {
```
prashantasdeveloper: we generally handle these in `getAuthorization` and use `signerPermissions` to specify the failed scenario

prashantasdeveloper (changes requested):

PR #1641, opened 2026-08-07
feat: 🎸 allow transferFunds between asset holders on different DIDs

prashantasdeveloper:
## Summary
- Removes the SDK-side block that rejected `assets.transferFunds` whenever `from` and `to` resolved to different identities. The chain (`settlement.transferFunds`) already supports this: it authorizes the source (allowance for `Account`, custody for `Portfolio`), creates a settlement instruction, auto-affirms the sender, and either executes immediately (if the receiver has automatic affirmation — the default) or leaves the instruction pending until the receiver affirms it.
- Fixed two spots that implicitly assumed same-DID:
  - The client-side balance/allowance pre-check now keys off whether the signer *owns* the source `Account` (spender/allowance mode) rather than DID equality, so cross-DID custody transfers (moving funds from a Portfolio owned by a different identity) get checked too.
  - `getAuthorization` now only requires portfolio permission on `toHolder` when the signer's own identity also owns it (mirroring the chain's auto-affirm condition) — previously it always required it, which would make cross-DID transfers fail permission checks since a signer's secondary-key permissions can never cover another identity's portfolio.
- `addInstruction`'s `createAddInstructionResolver` gained an optional `skipError` param, reused by `transferFunds` since the same-DID path legitimately emits no `InstructionCreated` event.

## Breaking change
`Assets.transferFunds` return type changed from `void` to `Instruction | undefined`. It resolves to the pending `Instruction` when a cross-DID transfer is left awaiting the receiver's affirmation, and `undefined` when the transfer settles immediately (same-DID, or the receiver auto-affirmed).

## Test plan
- [x] Unit tests updated/added in `transferFunds.ts` and `addInstruction.ts` test suites
- [x] Full suite passes (135 suites / 1158 tests) and `tsc` is clean
- [x] Verified end-to-end against a live chain (`polymesh-arm64:8705aca53b-develop-debian`) with two distinct identities:
  - Cross-DID Portfolio → Portfolio: returns pending `Instruction`, balance moves only after receiver affirms
  - Cross-DID Account → Account: same behavior
  - Same-DID Portfolio → Portfolio: still resolves `undefined`, settles immediately (no regression)

### DIFF
--- a/src/api/procedures/transferFunds.ts
+++ b/src/api/procedures/transferFunds.ts
@@ -22,7 +22,7 @@ import {
   fungibleMovementToPortfolioFund,
   nftMovementToPortfolioFund,
 } from '~/utils/conversion';
-import { asAssetId, asNftId } from '~/utils/internal';
+import { asAssetId, asNftId, filterEventRecords } from '~/utils/internal';
 import { isFungibleLegBuilder } from '~/utils/typeguards';
 
 export interface Storage {
@@ -191,8 +191,26 @@ export async function prepareTransferFunds(
   return {
     transaction: settlement.transferFunds,
     args: [rawFrom, rawTo, rawFund],
-    resolver: (receipt): Instruction | undefined =>
-      createAddInstructionResolver(context, true)(receipt)[0],
+    resolver: (receipt): Instruction | undefined => {
+      const [instruction] = createAddInstructionResolver(context, true)(receipt);
+      if (!instruction) {
+        return undefined;
+      }
+
+      // `settlement.InstructionCreated` fires whether or not the instruction is also executed
+      // inline within this same transaction (e.g. when the caller is also the receiver and both
+      // sides end up affirmed), so a settled instruction still resolves an `Instruction` above.
+      // `SettlementManuallyExecuted` only fires on that inline-execution path, so its presence
+      // means the instruction is already done rather than still pending.
+      const [executed] = filterEventRecords(
+        receipt,
+        'settlement',
+        'SettlementManuallyExecuted',
+        true
+      );
+
+      return executed ? undefined : instruction;
+    },
   };
 }
 
--- a/src/api/procedures/transferFunds.ts
+++ b/src/api/procedures/transferFunds.ts
@@ -1,10 +1,12 @@
 import { PolymeshPrimitivesPortfolioFund } from '@polkadot/types/lookup';
 
+import { createAddInstructionResolver } from '~/api/procedures/addInstruction';
 import { getAssetHolderDid } from '~/api/procedures/utils';
 import {
   Account,
   Context,
   DefaultPortfolio,
+  Instruction,
   NumberedPortfolio,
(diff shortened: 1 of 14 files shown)


src/api/procedures/transferFunds.ts
```diff
@@ -59,27 +61,28 @@ export async function getFund(
       });
     }
 
-    // when sender and receiver DID are same we need to only check asset balance has for either account or portfolio
-    if (fromDid === signingDid) {
-      const [balance] = await fromHolder.getAssetBalances({ assets: [asset] });
-      if (!balance || balance.free.lt(amount)) {
+    // spender mode only applies to Account sources where the signer isn't the owner; Portfolio
+    // sources are always authorized via custody on-chain, regardless of the owning DID
+    if (fromHolder instanceof Account && fromDid !== signingDid) {
```
F-OBrien: We should check the allowance when the fromHolder.address isn't the signer. For accounts only the account or another account with an allownace can tranfer the assets. The DID relationship is never checked.
```suggestion
    // allowances are granted key to key, so spender mode applies whenever the submitting Account
    // isn't the source Account itself — even when both keys belong to the same DID. Portfolio
    // sources are always authorized via custody on-chain, regardless of the owning DID
    if (fromHolder instanceof Account && fromHolder.address !== signingAccount) {
```
with that change we no longer need the signingDid from storage

src/api/procedures/transferFunds.ts
```diff
@@ -59,27 +61,28 @@ export async function getFund(
...
+      if (!balance || balance.free.lt(amount)) {
         throw new PolymeshError({
           code: ErrorCode.UnmetPrerequisite,
-          message: 'Spender has insufficient allowance to cover the transfer',
+          message: 'Sender has insufficient balance to cover the transfer',
           data: {
-            allowance,
+            balance,
           },
         });
```
F-OBrien: We should be checking the source balance even when taking the allowance/spender path.

src/api/procedures/transferFunds.ts
F-OBrien, line 1: The NFT branch is missing any checks. since allowances can't authorize it:
```
if (fromHolder instanceof Account && fromHolder.address !== signingAccount) {
  throw new PolymeshError({
    code: ErrorCode.UnmetPrerequisite,
    message: 'Only the owning key can transfer NFTs from an Account. Allowances do not apply to NFTs',
  });
}
```
Also also no ownership or lock status checks are done

F-OBrien: A cleaner, cast-free approach would be like the below for both fungible and nft
```
rawFund = await fungibleMovementToPortfolioFund(
  { asset, amount, ...(memo !== undefined && { memo }) },
  context
);
```

F-OBrien (changes requested):

F-OBrien (dismissed):

F-OBrien (approved):

Issue #1643, opened 2026-08-08
Holders cannot vote on a CorporateBallot: castBallotVote requires External Agent permissions


savelkouls:
Voting on a `CorporateBallot` fails for every ordinary token holder, because the
Procedure is authorised as if voting were an External Agent action.

The practical effect is backwards: an issuer, who *is* an Agent of the Asset, can
vote on its own ballot. The holders the ballot was opened for cannot.

```
The signing Identity doesn't have the required permissions to execute this procedure

{ message: 'The Identity is not an Agent for the Asset',
  missingPermissions: ['corporateBallot.vote'] }
```

## Reproduction

No signing key needed — `checkAuthorization` alone shows it. Requires an Asset, a
ballot in its voting window, and a holder who is not an Agent.

```ts
import { Polymesh } from '@polymeshassociation/polymesh-sdk';
import BigNumber from 'bignumber.js';

const client = await Polymesh.connect({ nodeUrl: 'ws://localhost:9944' });
const asset = await client.assets.getFungibleAsset({ assetId: ASSET_ID });
const { ballot } = await asset.corporateActions.ballots.getOne({
  id: new BigNumber(BALLOT_ID),
});

const result = await ballot.vote.checkAuthorization(
  { votes: [[{ power: new BigNumber(1) }, { power: new BigNumber(0) }, { power: new BigNumber(0) }]] },
  { signingAccount: HOLDER_ADDRESS }, // holds tokens at the record date; not an Agent
);

console.log(JSON.stringify(result.agentPermissions));
console.log(JSON.stringify(result.signerPermissions));
```

Observed:

```
{"missingPermissions":["corporateBallot.vote"],"result":false,"message":"The Identity is not an Agent for the Asset"}
{"result":true}
```

`roles` also returns `{ result: true }`. The Agent check is the only failure.

## Cause

`castBallotVote` declares a single `permissions` block naming the Asset
(`src/api/procedures/castBallotVote.ts`):

```ts
permissions: {
  transactions: [TxTags.corporateBallot.Vote],
  assets: [asset],
  portfolios: [],
}
```

`permissions` applies to both secondary Accounts and External Agent Identities
(`ProcedureAuthorization` in `src/types/internal.ts`), and
`Procedure.checkRolesAndAgentPermissions` runs the Agent check whenever both
`assets` and `transactions` are non-empty:

```ts
if (assets?.length && transactions?.length) {
[...]

PR #1644, opened 2026-08-08
fix: 🐛 authorise ballot voting as a holder action (issue #1643)

savelkouls:
### Description

**Casting a vote on a corporate ballot fails for every ordinary token holder.**
The Procedure is authorised as if voting were an External Agent action, so it is
unreachable for the callers it exists for.

`castBallotVote` declares a single `permissions` block naming the Asset:

```ts
permissions: {
  transactions: [TxTags.corporateBallot.Vote],
  assets: [asset],
  portfolios: [],
}
```

`permissions` applies to **both** secondary Accounts and External Agent
Identities, so scoping the transaction permission to the Asset also asserts that
the signer is an Agent of that Asset — `Procedure.checkRolesAndAgentPermissions`
runs the Agent check whenever both `assets` and `transactions` are non-empty.

A holder is not an Agent of the Asset they hold, so every vote fails with:

```
The signing Identity doesn't have the required permissions to execute this procedure

{ message: 'The Identity is not an Agent for the Asset',
  missingPermissions: ['corporateBallot.vote'] }
```

Note the practical effect: an issuer, who *is* an Agent, can vote on its own
ballot; the holders the ballot was opened for cannot.

**The chain does not require this.** The same vote, signed by the same holder
key, succeeds with `skipChecks: { agentPermissions: true }` and the tally is
recorded correctly. The requirement was only ever asserted client-side.

#### The change

fixing issue #1643

Declare the two separately: keep the Asset in `signerPermissions`, and set
`agentPermissions: true`.

```diff
-  return {
-    permissions: {
-      transactions: [TxTags.corporateBallot.Vote],
-      assets: [asset],
-      portfolios: [],
-    },
-  };
+  return {
+    signerPermissions: {
+      transactions: [TxTags.corporateBallot.Vote],
+      assets: [asset],
+      portfolios: [],
+    },
+    agentPermissions: true,
+  };
```

`ProcedureAuthorization` already documents both fields as taking precedence over
`permissions` for their respective signer types, and a boolean `agentPermissions`
short-circuits the Agent check.

#### Why not `assets: []`

`claimDividends` — the other holder-side corporate action — avoids the same trap
by passing an empty `assets` array, which works because the Agent check only runs
when `assets` is non-empty.

That shape was not copied here deliberately. `signerPermissions` is passed whole
to `Account.checkPermissions`, and `getMissingAssetPermissions` skips the Asset
comparison entirely when the required list is empty — so `assets: []` would also
drop the Asset scoping of a secondary Account's permissions, letting a secondary
key permissioned for one Asset vote on another. Splitting the declaration removes
the incorrect Agent assertion while keeping that scoping intact.

Out of scope, but worth flagging for maintainers: `claimDividends` carries that
same latent gap today.

#### Verification

- Observed and fixed against Polymesh **Development** chain, runtime `8000020`,
  SDK **v30.2.0**.
- `src/api/procedures/__tests__/castBallotVote.ts` — 11 passing, including two
  added cases: that no Agent permissions are required, and that the secondary
  Account's Asset scoping is retained. Reverting the fix fails three of them.
- Full suite green: **210 suites, 2533 tests**.
- ESLint clean on both changed files. `tsc` reports the same four pre-existing
  `rootDir`/`scripts` errors before and after.
- `prepareCastBallotVote` is untouched — the extrinsic and its arguments are
  unchanged, so there is no on-chain behaviour change.

### Breaking Changes

**None.** No public API signature changes, and nothing submitted to the chain
changes. The change only relaxes a client-side assertion, so no caller who
previously succeeded can now fail.

One observable behaviour change, which is the fix itself: for a holder who is not
an Agent, `ballot.vote.checkAuthorization(...)` now returns
`agentPermissions: { result: true }` instead of `{ result: false, message: 'The
Identity is not an Agent for the Asset' }`. Anything asserting that failure — a
test, or UI text telling holders they must be an Agent — will change accordingly.

### JIRA Link

N/A — external contribution, no JIRA access. Happy to open a GitHub issue first
if you would prefer this tracked before review.

### Checklist

  Not required — `README.md` does not document ballot voting or Procedure
  permissions, and no public API surface changed.

### DIFF
--- a/src/api/procedures/__tests__/castBallotVote.ts
+++ b/src/api/procedures/__tests__/castBallotVote.ts
@@ -246,11 +246,46 @@ describe('castBallotVote procedure', () => {
       } as Params;
 
       expect(boundFunc(args)).toEqual({
-        permissions: {
+        signerPermissions: {
           transactions: [TxTags.corporateBallot.Vote],
           assets: [asset],
           portfolios: [],
         },
+        agentPermissions: true,
+      });
+    });
+
+    it('should not require External Agent permissions, since voting is a holder action', () => {
+      // A holder is not an Agent of the Asset they hold. Declaring `permissions`
+      // with a non-empty `assets` applies to Agents as well, which made this
+      // Procedure unreachable for its intended callers.
+      const boundFunc = getAuthorization.bind(proc);
+      const args = {
+        asset,
+        ballot,
+        votes: [[{ fallback: new BigNumber(0), power: new BigNumber(1) }]],
+      } as Params;
+
+      const result = boundFunc(args);
+
+      expect(result.agentPermissions).toBe(true);
+      expect(result.permissions).toBeUndefined();
+    });
+
+    it('should still scope a secondary Account permission to the Asset', () => {
+      // The Agent check is what has to go, not the Asset scoping: a secondary
+      // key permissioned for one Asset should not be able to vote on another.
+      const boundFunc = getAuthorization.bind(proc);
+      const args = {
+        asset,
+        ballot,
+        votes: [[{ fallback: new BigNumber(0), power: new BigNumber(1) }]],
+      } as Params;
+
+      expect(boundFunc(args).signerPermissions).toEqual({
+        transactions: [TxTags.corporateBallot.Vote],
+        assets: [asset],
+        portfolios: [],
       });
     });
   });
--- a/src/api/procedures/castBallotVote.ts
+++ b/src/api/procedures/castBallotVote.ts
@@ -195,12 +195,30 @@ export function getAuthorization(
   this: Procedure<Params, void>,
   { asset }: Params
 ): ProcedureAuthorization {
+  /*
+   * Voting is a holder action, so no External Agent permissions are required.
+   *
+   * `permissions` applies to both secondary Accounts and External Agent
+   * Identities, so naming the Asset there in order to scope the transaction
+   * permission also asserts that the signer is an Agent of that Asset — see
+   * `Procedure.checkRolesAndAgentPermissions`, which runs the Agent check
+   * whenever both `assets` and `transactions` are non-empty. A holder is not an
+   * Agent of the Asset they hold, so that made this Procedure unreachable for
+   * its intended callers, failing with "The Identity is not an Agent for the
+   * Asset".
+   *
+   * These are declared separately rather than by emptying `assets` (as
+   * `claimDividends` does) so that the Asset scoping of a secondary Account's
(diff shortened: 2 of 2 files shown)


src/api/procedures/castBallotVote.ts
```diff
@@ -195,12 +195,30 @@ export function getAuthorization(
...
   return {
-    permissions: {
+    signerPermissions: {
       transactions: [TxTags.corporateBallot.Vote],
       assets: [asset],
       portfolios: [],
     },
+    agentPermissions: true,
   };
 }
```
F-OBrien, line 223: The correct fix here is to remove `asset` from the `assets` permission as voting is not an agent operation.  
```suggestion
export function getAuthorization(this: Procedure<Params, void>): ProcedureAuthorization {
  return {
    permissions: {
      transactions: [TxTags.corporateBallot.Vote],
      assets: [],
      portfolios: [],
    },
  };
}
```

src/api/procedures/__tests__/castBallotVote.ts
```diff
@@ -246,11 +246,46 @@ describe('castBallotVote procedure', () => {
...
+        votes: [[{ fallback: new BigNumber(0), power: new BigNumber(1) }]],
+      } as Params;
+
+      expect(boundFunc(args).signerPermissions).toEqual({
+        transactions: [TxTags.corporateBallot.Vote],
+        assets: [asset],
+        portfolios: [],
       });
     });
   });
```
F-OBrien, line 291: With the removal of the `asset` the test can be simplified per below:
```suggestion
  describe('getAuthorization', () => {
    it('should return the appropriate roles and permissions', () => {
      const boundFunc = getAuthorization.bind(proc);

      expect(boundFunc()).toEqual({
        permissions: {
          transactions: [TxTags.corporateBallot.Vote],
          assets: [],
          portfolios: [],
        },
      });
    });
  });
```

F-OBrien (changes requested):
@savelkouls Thank you for your bug report and PR. I've reviewed your proposed fix and suggested a amendments.

Your fix assumed that the asset permission is for a secondary key to be able to call the the function for that specific asset. However asset permissions are exclusively agent actions. As voting is not an agent operation the correct fix here is to remove the `asset` check which in turn prevents the SDK from doing an agent permission check.

ref permission docs for clarification on the asset permission https://developers.polymesh.network/identity/advanced/secondary-keys/#secondary-key-permissions 

Please also target the `develop` branch with your PR.

F-OBrien:
closed. work complete in PR #1649

PR #1658, opened 2026-08-22
fix: 🐛 read withdrawal slashing spans from the stash, not the controller

F-OBrien:
### Description

`sdk.staking.withdraw` sent the wrong `numSlashingSpans`.

`staking.slashingSpans` is keyed by the **stash**, but `prepareStorage` queried it
with the acting Account — which for `withdrawUnbonded` is the **controller**. Where
`setStakingController` has bonded a stash to a separate controller, the lookup misses
and the SDK sends `0`. The chain rejects the call with `IncorrectSlashingSpans` when a
slashed stash withdraws the last of its bond.

Fix: read the ledger first to learn the stash, then key the query by
`controllerEntry.stash.address`. This costs a serialised round trip in place of the
previous `Promise.all`, since the stash is only known from the ledger. Where the acting
Account is not a controller, the query is skipped and `optSpans` is `null` — that path
already throws before the count is used.

The existing suite only covered the self-controller case, which is why this passed.
Test storage now gives the stash a distinct address, and the new
`should fetch the slashing spans of the stash, not of the signing controller` case
fails against the old keying.

### Breaking Changes

None. `Storage` is `@hidden`, so widening `optSpans` to
`Option<PalletStakingSlashingSlashingSpans> | null` is not part of the public API.

### Checklist

### DIFF
--- a/src/api/procedures/withdrawUnbondedPolyx.ts
+++ b/src/api/procedures/withdrawUnbondedPolyx.ts
@@ -13,7 +13,8 @@ import { bigNumberToU32, stringToAccountId } from '~/utils/conversion';
 export interface Storage {
   actingAccount: Account;
   controllerEntry: StakingLedger | null;
-  optSpans: Option<PalletStakingSlashingSlashingSpans>;
+  /** `null` if there is no `controllerEntry`, since the query is keyed by its stash */
+  optSpans: Option<PalletStakingSlashingSlashingSpans> | null;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/api/procedures/withdrawUnbondedPolyx.ts
```diff
@@ -84,17 +85,18 @@ export async function prepareStorage(this: Procedure<void, void, Storage>): Prom
   } = this;
 
   const actingAccount = await context.getActingAccount();
-  const rawActingAddress = stringToAccountId(actingAccount.address, context);
 
-  const [controllerEntry, spans] = await Promise.all([
-    actingAccount.staking.getLedger(),
-    slashingSpans(rawActingAddress),
-  ]);
+  const controllerEntry = await actingAccount.staking.getLedger();
```
prashantasdeveloper, line 89: both can be merged using Promise.all

prashantasdeveloper (approved):
small nit comment left, other than that LGTM. Approving

src/api/procedures/withdrawUnbondedPolyx.ts
F-OBrien: I don't think they can be - as each await is dependent on the result of the previous one. leaving as is.

PR #1662, opened 2026-08-26
Ordering, filtering and paging for indexer-backed reads

F-OBrien:
### Description

Ordering, filtering and paging fixes for indexer-backed reads. Five commits, each self-contained and green on its own.

**Ten paginated queries had no order that was both defined and unique.** Three had no `orderBy` at all — behind `Claims.getAllCustomClaimTypes`, `DividendDistribution.getPaymentHistory` and `MultiSig.getHistoricalProposals` — so the database was free to return rows in any order. Seven more ordered by a block-level field only, which leaves rows sharing a block undefined relative to each other. `first`/`offset` paging over an undefined order can show one row twice and never show another.

They now order by something unique. Where the indexer's `id` is block-derived it is used directly: it is `<block number>/<event index>` with both parts zero-padded, so ordering it as a string orders by block and then by position within the block. Where an id is *not* block-derived it is only a tiebreaker — a custom claim type's id is the chain's numeric id, and a MultiSig proposal's embeds an unpadded proposal number, so proposals order by `proposalId` instead. Sort direction is unchanged wherever one was defined.

**Falsy filter values were dropped.** The filter builder skipped any value failing a truthiness check, so `Account.getTransactionHistory({ success: false })` sent no `success` filter and returned *all* transactions rather than only the failed ones. `undefined` and `null` are now skipped; `0`, `''` and `false` are honoured.

**`orderBy` on every paginated query**, exposed on nine public reads: asset and NFT transaction history, distribution payments, MultiSig proposals, operation history, investments, custom claim types, historical authorizations, historical instructions and indexed events. Defaults are unchanged, so omitting it behaves exactly as before. Note a supplied ordering *replaces* the default outright, tiebreaker included.

**Filters are built from the attributes supplied.** Six queries wrote their `filter:` block by hand, so the set of filter fields was fixed and adding one meant editing the GraphQL. They now share the existing helper, which also gained comparisons beyond `equalTo`.

**`Identity.getAssetHoldings()`** returns each fungible Asset with the amount held, and takes `heldNow` to exclude holdings transferred away in full — the indexer keeps a holding record once it exists, so those otherwise come back at zero. The exclusion is applied **in the query**; filtering client-side would leave `count` and `next` describing the unfiltered set, so paging would give short pages against a cursor that never ends. `getHeldNfts` gains the same option. `getHeldAssets` is deprecated in favour of it.

### Breaking Changes

Each replaces silently wrong output with correct output. All listed in `changelogs/v31.0.0-next.md` under **Behaviour changes**.

- Ten paginated queries changed the relative order of rows that were previously tied. Sort direction is unchanged wherever one was defined.
- `Identity.getHeldAssets()` is deprecated, for removal in the next major. Otherwise unchanged.
- `multiSigProposalsQuery` takes a filters object rather than a bare address — internal only, one caller updated.

### Worth passing to the indexer team

- Ids are padded inconsistently. Event-derived ids are zero-padded so lexical order equals numeric, but `CustomClaimType`, `Instruction` and `MultiSigProposal` are not — ordering those by `id` puts proposal 10 before 9.
- `createdBlock` is unindexed on 8 of the 10 entities we order by. Moving to `id` sidesteps it (primary key), but `CREATED_BLOCK_ID_*` orderings want an index if they are to stay usable.
- `schema.graphql` and the served API disagree on scalars: every SubQuery `BigInt!` surfaces as `BigFloat`, so `schema.graphql` is not safe to write queries against — `src/middleware/types.ts` is.

### Follow-up

Left out deliberately, since each adds pagination surface and touches the shared `MiddlewarePaginationOptions` — better designed together than piecemeal:

- Paging the reads that still have none. `Portfolio.getTransactionHistory` and `Portfolios.getTransactionHistory` fetch every movement ever recorded.
- `PaginationOptions` on `Account`/`Portfolio.getAssetBalances` and `getCollections`.
- Adopting `MiddlewarePaginationOptions` across the indexer reads, which is also what unblocks `orderBy` on `Instruction.getAffirmations`, `getMediators` and `getLegs`.

### Checklist

### DIFF
--- a/changelogs/v31.0.0-next.md
+++ b/changelogs/v31.0.0-next.md
@@ -226,6 +226,32 @@ const { data } = await account.getPolyxTransactions({
 
 It defaults to `IdDesc`, newest first — see the behaviour change below. The indexer's `id` is `<block number>/<event index>` with both parts zero-padded, so it is unique and sorts by block and then by position within the block. `CreatedBlockIdDesc` would have left transactions sharing a block in no defined order, which makes pages repeat or skip entries.
 
+### Choose the order of indexer-backed results
+
+Every paginated indexer query now accepts an `orderBy`, so a caller is no longer stuck with the SDK's default. The reads that expose it:
+
+| Method                                                                       | Ordering type                       |
+| ---------------------------------------------------------------------------- | ----------------------------------- |
+| `FungibleAsset.getTransactionHistory`, `NftCollection.getTransactionHistory` | `AssetTransactionsOrderBy`          |
+| `DividendDistribution.getPaymentHistory`                                     | `DistributionPaymentsOrderBy`       |
+| `MultiSig.getHistoricalProposals`                                            | `MultiSigProposalsOrderBy`          |
+| `AssetPermissions.getOperationHistory`                                       | `TickerExternalAgentActionsOrderBy` |
+| `Offering.getInvestments`                                                    | `InvestmentsOrderBy`                |
+| `Claims.getAllCustomClaimTypes`                                              | `CustomClaimTypesOrderBy`           |
+| `Authorizations.getHistoricalAuthorizations`                                 | `AuthorizationsOrderBy`             |
+| `Identity.getHistoricalInstructions`                                         | `InstructionsOrderBy`               |
+| `Network.getEventsByIndexedArgs`                                             | `EventsOrderBy`                     |
+
+```ts
+const { data } = await asset.getTransactionHistory({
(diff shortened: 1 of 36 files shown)


prashantasdeveloper (dismissed):
Should we make order by as list of order by enums? Ideally that would allow users to resolve conflicting cases as well based on use case. Other than that LGTM

prashantasdeveloper (dismissed):

prashantasdeveloper (approved):

